PR #2, opened 2015-03-06
Fix pep8 errors reported for oauthenticator.py

yuvipanda:


minrk:
please don't make cleanup PRs.

yuvipanda:
Uhm, ok. Thanks. Perhaps you should put it in your readme or something so
people do not waste their time?

Carreau:
Sorry about that, we usually follow IPython one, and assume that cleanup like that are done by scripts. 

https://github.com/jupyter/oauthenticator/pull/3 add a contributing.md, and also the reason why we don't like cleanup PR is that it screw up git blame most of the time.

yuvipanda:
@Carreau Thanks, that makes more sense. I've found git blame to be not a problem with the `-w -M` flags to it. Either way, thanks for making it clearer!

minrk:
>  Perhaps you should put it in your readme or something so people do not waste their time?

It is a reasonable conclusion that if a cleanup PR would make any changes it is because the cleanup is applying a policy that the project doesn't follow. Don't assume a policy is strictly followed by a project unless the project explicitly states that there is such a policy. Style PRs are rarely a good idea, and never a good idea without asking first.

Cleanup PRs have a history of introducing bugs as well (this has happened multiple times on IPython and matplotlib), even when the intention is to make no code changes at all. They simply offer zero benefit and have nonzero cost.

yuvipanda:
I forked this to build https://github.com/yuvipanda/mwoauthenticator and my instinctive first reaction is to fix the PEP8 errors that my editor shows up. I just pushed this one patch upstream in case it was useful. #3 is useful. Sorry if you think I've wasted your time, and I'll try to not do that in the future.

minrk:
Not a problem, sorry for being curt. I just really dislike cleanup PRs and programmatically enforced styles, I think they are a bad idea for everyone.

PR #4, opened 2015-06-23
Use standalone traitlets

EdwardJKim:
jupyter/jupyterhub#261 removes the dependency on IPython.utils.traitlets. This PR uses the standalone traitlets directly.


minrk:
Thanks!

rgbkrk:
:+1:

PR #6, opened 2015-07-16
Authenticate Bitbucket's users basing on team membership

Xarthisius:
This PR introduces new config option `c.BitbucketOAuthenticator.team_whitelist` that allows to specify a set of *teams* (using bb's nomenclature) that are allowed to access Jupyter hub.

### DIFF
--- a/oauthenticator.py
+++ b/oauthenticator.py
@@ -158,6 +158,7 @@ class BitbucketOAuthenticator(Authenticator):
                         config=True)
     client_secret = Unicode(os.environ.get('BITBUCKET_CLIENT_SECRET', ''),
                             config=True)
+    team = Unicode('', config=True)
 
     def login_url(self, base_url):
         return url_path_join(base_url, 'oauth_login')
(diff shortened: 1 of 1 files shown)


oauthenticator.py
```diff
@@ -216,10 +220,38 @@ def authenticate(self, handler):
...
             username = None
         raise gen.Return(username)
 
+    @gen.coroutine
+    def check_whitelist(self, username, headers):
+        if self.team_whitelist:
+            valid_user = yield self._check_group_whitelist(username, headers)
+        else:
+            valid_user = yield super().check_whitelist(username)
+        raise gen.Return(valid_user)
```
minrk: We can use `return valid_user` here and elsewhere, since we don't need to support legacy Python.

Xarthisius:
I've tried using `return` at the very beginning, but somehow I couldn't get it to work. Now it works.

minrk:
Thanks.

PR #7, opened 2015-07-25
Update README.md

jzmiller1:
GithubOAuthenticator example doesn't work after the Bitbucket branch was merged in.  This commit fixes the example.


minrk:
Sorry about that, I should have noticed the change in the BB PR. I'll fix it so that the previous way continues to work without any changes.

PR #10, opened 2015-08-04
Adding support for GitHub Enterprise

cmclaughlin:
I added support for a new GITHUB_HOST environment variable so we can use our GitHub Enterprise instance for jupyterhub auth.

### DIFF
--- a/oauthenticator.py
+++ b/oauthenticator.py
@@ -21,10 +21,16 @@ from jupyterhub.utils import url_path_join
 
 from traitlets import Unicode, Set
 
+# Support github.com and github enterprise installations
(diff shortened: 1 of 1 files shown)


oauthenticator.py
```diff
@@ -21,10 +21,16 @@
 
 from traitlets import Unicode, Set
 
+# Support github.com and github enterprise installations
+GITHUB_HOST = os.environ.get('GITHUB_HOST', 'github.com')
```
minrk: This should use `get('GITHUB_HOST') or 'github.com'` to handle the `GITHUB_HOST=''` case, which should behave the same as being unset.

cmclaughlin: Done.

PR #22, opened 2015-12-02
Add MediaWiki based OAuth authenticator

### LINKED ISSUE #21: Add a MediawikiOAuthenticator
Hello! I've been using https://github.com/yuvipanda/mwoauthenticator at Wikimedia's test installation, and would like to get it merged 'upstream' (here). Is this repo accepting new such authenticators, or should it continue to live in a separate repo? If this is acceptable here I'll do the work in integrating it and send a PR. Wanted to ask before I expend effort :)

Thanks!

yuvipanda:
Fixes #21

### DIFF
--- a/oauthenticator/__init__.py
+++ b/oauthenticator/__init__.py
@@ -1,3 +1,4 @@
 from .oauth2 import *
 from .github import *
 from .bitbucket import *
+from .mediawiki import *
--- a/oauthenticator/__init__.py
+++ b/oauthenticator/__init__.py
@@ -1,4 +1,3 @@
 from .oauth2 import *
 from .github import *
 from .bitbucket import *
-from .mediawiki import *
--- a/oauthenticator/mediawiki.py
+++ b/oauthenticator/mediawiki.py
@@ -1,0 +1,123 @@
+"""
+Custom Authenticator to use GitHub OAuth with JupyterHub
+
+Most of the code c/o Yuvi Panda (@yuvipanda)
+"""
+
+import os
+import json
+from concurrent.futures import ThreadPoolExecutor
+
+from tornado import gen, web
+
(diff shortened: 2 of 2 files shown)


yuvipanda:
Will need followup to modify the README possibly.

oauthenticator/__init__.py
```diff
@@ -1,3 +1,4 @@
 from .oauth2 import *
 from .github import *
 from .bitbucket import *
+from .mediawiki import *
```
minrk: since mediawiki has added dependencies, lets exclude it from `__init__` here. I only put the ones that existed before the multi-module split here, and don't think we should add more.

yuvipanda: Alright! In that case, I wonder if I should call the class mediawiki.OAuthenticator, since mediawiki.MWOAuthenticator seems redundant.

yuvipanda: Another option would be for it to fail gracefully if it can not load mwoauth dependency (by just printing an appropriate message when the class is attempted to be used, rather than at import time)

minrk: I would just remove it, and leave the name as MWOauthenticator. It shouldn't be imported if it isn't used. I wouldn't even put GitHub and BitBucket here if they weren't already present in the same namespace when it was just `oauthenticator.py`.

yuvipanda: My underlying concern is that there'll be inconsistency between oauthenticator.GitHubOAuthenticator vs oauthenticator.mediawiki.MWOauthenticator.

minrk: It should generally be `oauthenticator.github.GitHubOAuthenticator`. It is already this way for CILogon.

yuvipanda: Fair enough. Updated.

PR #24, opened 2015-12-05
[Mediawiki] Add an option to pass OAuth secrets to env

yuvipanda:
For https://phabricator.wikimedia.org/T120331

### DIFF
--- a/oauthenticator/mediawiki.py
+++ b/oauthenticator/mediawiki.py
@@ -21,7 +21,7 @@ from mwoauth.tokens import RequestToken
 
 from traitlets import Any, Integer, Unicode
 
-from oauthenticator import OAuthenticator
+from oauthenticator import OAuthenticator, OAuthCallbackHandler
 
 
 # Name of cookie used to pass auth token between the oauth
@@ -42,7 +42,6 @@ def dejsonify(js):
     return RequestToken(key.encode('utf-8'), secret.encode('utf-8'))
 
 class MWLoginHandler(BaseHandler):
-
     @gen.coroutine
     def get(self):
         consumer_token = ConsumerToken(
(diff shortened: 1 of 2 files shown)


yuvipanda:
Depends on https://github.com/jupyter/jupyterhub/pull/346

oauthenticator/mediawiki.py
```diff
@@ -116,6 +120,23 @@ class MWOAuthenticator(OAuthenticator):
...
+    )
+
+    def pre_spawn_start(self, user, spawner):
+        if not self.pass_secrets:
+            return
+
+        spawner.env.update({
+            'CLIENT_SECRET': self.client_secret,
+            'CLIENT_ID': self.client_id,
+            'ACCESS_KEY': user.state['ACCESS_KEY'],
```
minrk: Spawners shouldn't ever access user.state. These should always be loaded by `self.load_state` and retrieved by `self.get_state`.

willingc:
@yuvipanda Hi Yuvi, Doing a bit of spring cleaning on repos. 🌸 🌷 I'm wondering if this needs additional action or should be closed. Thoughts? Thanks, Carol

yuvipanda:
ooo nice! I'm waiting for some time to be able to test
https://github.com/jupyterhub/jupyterhub/pull/508 and then rebase this to
make use of that. I'm swamped and travelling until end of month however...

On Tue, May 10, 2016 at 8:26 PM, Carol Willing <notifications@github.com>
wrote:

> @yuvipanda <https://github.com/yuvipanda> Hi Yuvi, Doing a bit of spring
> cleaning on repos. 🌸 🌷 I'm wondering if this needs additional action or
> should be closed. Thoughts? Thanks, Carol
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/jupyterhub/oauthenticator/pull/24#issuecomment-218183854>
>



-- 
Yuvi Panda T
http://yuvi.in/blog

willingc:
@yuvipanda Cool. Then I will leave this open. Happy travels.

gwincr11:
@yuvipanda I would be very interested in seeing how you accomplished this, I have a very similar need.

yuvipanda:
fwiw, this is still actually a problem. You can't keep auth secrets by patching spawner to use spawner's state, since that gets cleared when user's servers get stopped. So if your server gets stopped, but you aren't logged out - and start server again, there's no way to get this state back...

minrk:
Authenticators now have their own state in `.auth_state`. They can access this in pre_spawn_start if they need to. State isn't forcefully cleared, `Spawner.clear_state()` is called. It's entirely up to the Spawner what data gets stored in `.get_state()`, even after state is cleared on stop. DockerSpawner will still return the container ID for stopped but not destroyed containers, for instance.

oauthenticator/mediawiki.py
```diff
@@ -116,6 +120,23 @@ class MWOAuthenticator(OAuthenticator):
...
+        help='Pass OAuth consumer and access secrets to the spawner'
+    )
+
+    def pre_spawn_start(self, user, spawner):
+        if not self.pass_secrets:
+            return
+
+        spawner.env.update({
+            'CLIENT_SECRET': self.client_secret,
+            'CLIENT_ID': self.client_id,
```
minrk: It's probably not right to pass the Hub's own OAuth id and secret to all users, is it? This would mean that all users have the ability to access MW *as the Hub*.

willingc:
@yuvipanda Thoughts on this PR... keep open? close?

yuvipanda:
@willingc I'm going to pick this up again today, since the auth_state patch has been merged :)

yuvipanda:
Opening a new PR!

PR #26, opened 2015-12-24
mediawiki: Be even more specific about where cookies are valid

yuvipanda:
- Only on the callback URL!
- Can not be read from JS!

Securitayh! Do not leak the secure cookie as much as possible

### DIFF
--- a/oauthenticator/mediawiki.py
+++ b/oauthenticator/mediawiki.py
@@ -24,7 +24,7 @@ from oauthenticator import OAuthenticator
 
 # Name of cookie used to pass auth token between the oauth
 # login and authentication phase
-AUTH_REQUEST_COOKIE_NAME = 'mw_oauth_request_token_v1'
+AUTH_REQUEST_COOKIE_NAME = 'mw_oauth_request_token_v2'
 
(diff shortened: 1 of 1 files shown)


yuvipanda:
Travis seems misconfigured? Uses rake?!

yuvipanda:
Hmm, this is actually failing to read the cookies in the callback, not fully sure why. I see that the cookie header is being sent...

yuvipanda:
Ah, had just missed a few more commits. Works fine now :)

rgbkrk:
Looks like the case of a not existing .travis.yml yet Travis was enabled on the repo.

rgbkrk:
Added a simple `.travis.yml` in #27 that installs the package and makes sure that the import works. If you rebase, you'll get the .travis.yml. Otherwise, it's not a big deal - this wasn't being tested before.

yuvipanda:
@rgbkrk ok, I've rebased now :)

rgbkrk:
Ok, cool.

rgbkrk:
On another note, I think we should welcome you aboard in helping maintain this plugin so you're not blocked on mediawiki things. I'm going to add you as a collaborator.

PR #29, opened 2016-01-06
Allow github usernames to map to specified system usernames

### LINKED ISSUE #28: Is it possible to have a github username map to a different system username
My github name is `rcompton`. The username on the computer I want to use is `rycpt`. Can I configure this authenticator to have `rcompton` map to `/home/rycpt` ?

rcompton:
closes #28

This allows one to specify
```
c.GitHubOAuthenticator.username_map = {<github username>: <system username>}
```
 in `jupyterhub_config.py`

### DIFF
--- a/oauthenticator/github.py
+++ b/oauthenticator/github.py
@@ -54,8 +54,8 @@ class GitHubOAuthenticator(OAuthenticator):
     client_secret_env = 'GITHUB_CLIENT_SECRET'
     login_handler = GitHubLoginHandler
     
-    github_to_nix_usernames_map = Dict(config=True, 
-        help="""Optional dict to remap github usernames to nix usernames.
+    username_map = Dict(config=True, default_value={},
+                        help="""Optional dict to remap github usernames to nix usernames.
         
         User github usernames for keys and existing nix usernames as values.
         cf https://github.com/jupyter/oauthenticator/issues/28 
@@ -108,10 +108,8 @@ class GitHubOAuthenticator(OAuthenticator):
         
         github_username = resp_json["login"]
         #remap gihub username to system username
-        if self.github_to_nix_usernames_map and github_username in self.github_to_nix_usernames_map.keys():
-            nix_username = self.github_to_nix_usernames_map[github_username]
-        else:
-            nix_username = github_username
+        nix_username = self.username_map.get(github_username, github_username)
+
         #check system username against whitelist
         if self.whitelist and nix_username not in self.whitelist:
             nix_username = None
--- a/oauthenticator/github.py
+++ b/oauthenticator/github.py
@@ -16,7 +16,7 @@ from tornado.httpclient import HTTPRequest, AsyncHTTPClient
 
 from jupyterhub.auth import LocalAuthenticator
 
-from traitlets import Unicode
+from traitlets import Unicode, Dict
 
 from .oauth2 import OAuthLoginHandler, OAuthenticator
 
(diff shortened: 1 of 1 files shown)


oauthenticator/github.py
```diff
@@ -99,10 +106,16 @@ def authenticate(self, handler):
         resp = yield http_client.fetch(req)
         resp_json = json.loads(resp.body.decode('utf8', 'replace'))
         
-        username = resp_json["login"]
-        if self.whitelist and username not in self.whitelist:
-            username = None
-        return username
+        github_username = resp_json["login"]
+        #remap gihub username to system username
+        if self.github_to_nix_usernames_map and github_username in self.github_to_nix_usernames_map.keys():
```
yuvipanda: You can set the default to be {} and just use:

```
if github_username in self.github_to_nix_usernames_map:
```

oauthenticator/github.py
```diff
@@ -54,6 +54,13 @@ def _github_client_secret_changed(self, name, old, new):
     client_secret_env = 'GITHUB_CLIENT_SECRET'
     login_handler = GitHubLoginHandler
     
+    github_to_nix_usernames_map = Dict(config=True, 
+        help="""Optional dict to remap github usernames to nix usernames.
```
yuvipanda: Use dedent to make sure that the extra leading whitespace is handled properly

oauthenticator/github.py
```diff
@@ -54,6 +54,13 @@ def _github_client_secret_changed(self, name, old, new):
     client_secret_env = 'GITHUB_CLIENT_SECRET'
     login_handler = GitHubLoginHandler
     
+    github_to_nix_usernames_map = Dict(config=True,
```
yuvipanda: I think this should just be called github_user_name_map, since these are not necessarily Local Unix Users - can be anything depending on the particular spawner being used.

yuvipanda:
Thank you for your patch! I've left some comments about possible style changes, and I can test it later as well :) Do you think you can also squash the commits into one with a more descriptive commit message?

oauthenticator/github.py
yuvipanda: This entire if / else can also be replaced with:

```
self.github_to_nix_usernames_map.get(github_username, github_username)
```

oauthenticator/github.py
minrk: There's no need to use dedent on help strings. Any code that renders this help string should be responsible for dedent.

oauthenticator/github.py
minrk: probably just `username_map`

rcompton:
Made the style changes and squashed down the commits.

I use it like:

```
c.JupyterHub.authenticator_class = 'oauthenticator.LocalGitHubOAuthenticator'
c.GitHubOAuthenticator.oauth_callback_url = os.environ['OAUTH_CALLBACK_URL']
c.GitHubOAuthenticator.username_map = {<github username>: <system username>}
```

oauthenticator/github.py
```diff
@@ -99,10 +106,14 @@ def authenticate(self, handler):
...
-        username = resp_json["login"]
-        if self.whitelist and username not in self.whitelist:
-            username = None
-        return username
+        github_username = resp_json["login"]
+        #remap gihub username to system username
+        nix_username = self.username_map.get(github_username, github_username)
+
+        #check system username against whitelist
+        if self.whitelist and nix_username not in self.whitelist:
```
minrk: There's ambiguity here. I would expect the whitelist to be applied *before* transforming the name to the system user.

oauthenticator/github.py
```diff
@@ -54,6 +54,13 @@ def _github_client_secret_changed(self, name, old, new):
     client_secret_env = 'GITHUB_CLIENT_SECRET'
     login_handler = GitHubLoginHandler
     
+    username_map = Dict(config=True, default_value={},
```
minrk: You don't have to specify `default_value={}`, that is already the default.

minrk:
Thanks, this is nice. I think we may fold this into an upstream proposal to add a generic mechanism for username normalization, but that doesn't mean we can't proceed with this PR as-is. What do you folks think about whether the whitelist should apply to the incoming or outgoing names? My initial reaction is that it should apply to the incoming names, but if we implement username_map as part of normalization, I would expect it to take effect afterward.

PR #32, opened 2016-05-08
use Authenticator.get_authenticated_user

minrk:
This is a wrapper around Authenticator.authenticate that applies:

- normalization (username_map, case-insensitive)
- validation
- whitelist checking

so we don't need to have copies of those steps here

### DIFF
--- a/oauthenticator/bitbucket.py
+++ b/oauthenticator/bitbucket.py
@@ -40,7 +40,7 @@ class BitbucketOAuthenticator(OAuthenticator):
     )
 
     @gen.coroutine
-    def authenticate(self, handler):
+    def authenticate(self, handler, data=None):
         code = handler.get_argument("code", False)
         if not code:
             raise web.HTTPError(400, "oauth callback made without a token")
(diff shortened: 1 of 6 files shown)


minrk:
@Analect this should solve your case-sensitivity problem. If you could try logging in to https://demohub.jupyer.org, this PR is deployed there for testing.

@ellisonbg I believe you were seeing case-sensitivity issues in usernames as well, which should be fixed by this.

Analect:
@minrk 
Thought https://demohub.jupyter.org was down ... but now see it was a typo at link above.
Anyway, I rebuilt with my own set-up using `-e git+https://github.com/jupyterhub/oauthenticator@b04f7946bcae066052ed02a6bdeb24d4f8632919#egg=oauthenticator` and it works. I think it might need a new release to get it to install the new fix with just `oauthenticator==0.3.*` ... or at least the latter didn't work for me.
Thanks again for helping to diagnose the problem .. and of course for fixing it too!

minrk:
Thanks @Analect! I'll push a new release shortly.

PR #35, opened 2016-06-08
Gitlab plugin added.

dobos:
Just a simple oauth plugin for gitlab based on the code written for github. Tested with local gitlab installation.


minrk:
Thanks!

PR #40, opened 2016-07-16
Gitlab configuration and host override details

vilhelmen:
Just notes from brute forcing the Gitlab config for our CE instance. Mostly just a copy/paste from the GitHub notes, but it may help others.


minrk:
Thanks!

PR #42, opened 2016-07-18
adding reference to enviornment variable for ubuntu closes #41

### LINKED ISSUE #41: Add README reference to how to add environment variables for Ubuntu
It would be helpful to have the simple instructions for adding environment variables `/etc/environment` in the README documentation

datapolitan:
Added reference to location of environment variables to make the documentation clearer for non-experts (like myself), trying to install JupyterHub

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ auth).
 
 You will additionally need to specify the OAuth callback URL, the client ID, and
 the client secret (you should have gotten these when you created your OAuth app
-on GitHub). For example, if these values are in the environment variables
+on GitHub). For example, if these values are in the environment variables ([`/etc/environment` in Ubuntu](https://help.ubuntu.com/community/EnvironmentVariables#System-wide_environment_variables))
 `$OAUTH_CALLBACK_URL`, `$GITHUB_CLIENT_ID` and `$GITHUB_CLIENT_SECRET`, you
 should add the following to your `jupyterhub_config.py`:
 ```


README.md
```diff
@@ -45,7 +45,7 @@ auth).
 
 You will additionally need to specify the OAuth callback URL, the client ID, and
 the client secret (you should have gotten these when you created your OAuth app
-on GitHub). For example, if these values are in the environment variables
+on GitHub). For example, if these values are in the environment variables ([`/etc/environment` in Ubuntu](https://help.ubuntu.com/community/EnvironmentVariables#System-wide_environment_variables))
```
minrk: Credentials like these shouldn't actually go in system-wide environment variables, because that would expose information to other users on the system. The Ubuntu page is good, but can we remove the reference to system-wide settings / `/etc/environment`?

datapolitan: I guess this speaks to the confusion I have over whether this should be run in `sudo

minrk: Whether it should be run with sudo depends on what permissions it needs. If it needs to switch users for the LocalProcessSpawner, generally the Hub does need to be root. However, OAuth deployments often use docker, where the Hub *may* not need to be root, in which case root permissions are not required. The Authenticator generally does not dictate whether the Hub needs root (though some do), it's typically the Spawner that determines whether root is needed or not.

In every case, though, setting credentials system-wide for all processes would not be the way to go.

Often, environment variables are set in a startup script, e.g.:

```bash
#!/bin/bash
# call this launch-jupyterhub.sh
export GITHUB_CLIENT_ID=...
exec jupyterhub $@
```

or in a `supervisor.conf` service configuration, or a docker script.

There are many different ways to specify environment variables, depending on the tool used to launch JupyterHub, and these are generally not sensitive to whether `sudo` is used or not.

willingc:
I'm going to close this PR and the related issue. Thanks @datapolitan for the question and interest in JupyterHub and oauthenticator.

PR #54, opened 2016-11-22
Adding auth0 OAuthenticator implementation

namdets:
Auth0 is a OAuth2 authentication aggregator service that supports mixing LDAP, SAML, OAuth2, etc. backing authentication providers.

We find it very useful.

You can test it by creating a free account here : https://auth0.com/opensource

### DIFF
--- a/oauthenticator/auth0.py
+++ b/oauthenticator/auth0.py
@@ -11,18 +11,14 @@ The following environment variables may be used for configuration:
     OAUTH_CALLBACK_URI - Your callback handler URI
 
 Additionally, if you are concerned about your secrets being exposed by
-an env dump(I know I am!) you can create a function that returns them in 
-a dict and attach them to the config object for Auth0OAuthenticator.
+an env dump(I know I am!) you can set the client_secret, client_id and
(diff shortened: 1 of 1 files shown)


oauthenticator/auth0.py
```diff
@@ -0,0 +1,149 @@
...
+
+    oauth_variable_config = Any(
+        help="""Any callable that returns a dictionary of oauth variables by the following names:
+        OAUTH_CLIENT_ID
+        OAUTH_CLIENT_SECRET
+        OAUTH_CALLBACK_URI
+        """
+    ).tag(config=True)
+
+    def oauth_environment_variable_config(self):
```
minrk: Can we skip this, and leave the getting of CLIENT_ID, etc. from the env the same as all the other implementations?

namdets:
Wow, Thanks for that!

So much cleaner!

namdets:
Hold on, there may now be a bug with the environment variable based config. Update incoming.

namdets:
OK, false alarm. I had a bug in my environment config, not the code.

I'm happy if you are.

minrk:
Thanks!

PR #56, opened 2016-11-23
fix bitbucket team whitelisting

jbweston:
Now we perform team whitelist checks inside `authenticate`, as
the check requires the OAuth access token. Previously this token
was passed implicitly to the `check_whitelist` method with use
of an instance variable. This would break in the case of concurrent
authentications.


jbweston:
Note that I haven't been able to test this code as there is currently no testing framework for this project.

minrk:
Thanks, @jbweston! I'll work on mocking so we can test these things better.

PR #58, opened 2016-11-24
Group/Organization whitelisting

### LINKED ISSUE #57: Add team whitelisting for GitHubOAuthenticator and  GitLabOAuthenticator
BitBucketOAuthenticator has team whitelisting, so it would seem reasonable to also add this for Gitlab and Github.

I am already working towards submitting a pull request, but am creating this issue here to serve as a centre for discussion and to avoid spamming the jupyterhub gitter channel.

jbweston:
This adds Gitlab group whitelisting and Github organization whitelisting to the corresponding authenticators. Fixes #57.

As this contains nontrivial changes, I am marking this as work in progress (WIP) until @minrk has set up a mocking framework to test this.

If I should submit this as 2 (or 3) separate pull requests, feel free to close this and I will split everything up.

### DIFF
--- a/oauthenticator/github.py
+++ b/oauthenticator/github.py
@@ -131,26 +131,30 @@ class GitHubOAuthenticator(OAuthenticator):
         # Check if user is a member of any whitelisted organizations.
         # This check is performed here, as it requires `access_token`.
         if self.github_organization_whitelist:
-            user_in_org = yield self._check_organization_whitelist(username, access_token)
-            return username if user_in_org else None
+            for org in self.github_organization_whitelist:
+                user_in_org = yield self._check_organization_whitelist(org, username, access_token)
+                if user_in_org:
+                    return username
+            else:  # User not found in member list for any organisation
+                return None
         else:  # no organization whitelisting
             return username
 
 
     @gen.coroutine
-    def _check_organization_whitelist(self, username, access_token):
+    def _check_organization_whitelist(self, org, username, access_token):
         http_client = AsyncHTTPClient()
         headers = _api_headers(access_token)
-        # Get all the user's organizations
-        next_page = "https://%s/orgs" % GITHUB_API
+        # Get all the members for organization 'org'
+        next_page = "https://%s/orgs/%s/members" % (GITHUB_API, org)
         while next_page:
             req = HTTPRequest(next_page, method="GET", headers=headers)
             resp = yield http_client.fetch(req)
             resp_json = json.loads(resp.body.decode('utf8', 'replace'))
             next_page = _get_next_page(resp)
-            user_orgs = set(entry["login"] for entry in resp_json)
+            org_members = set(entry["login"] for entry in resp_json)
             # check if any of the organizations seen thus far are in whitelist
-            if len(self.github_organization_whitelist & user_orgs) > 0:
+            if username in org_members:
                 return True
         return False
 
--- a/oauthenticator/github.py
+++ b/oauthenticator/github.py
@@ -25,9 +25,9 @@ from .oauth2 import OAuthLoginHandler, OAuthenticator
 # Support github.com and github enterprise installations
 GITHUB_HOST = os.environ.get('GITHUB_HOST') or 'github.com'
 if GITHUB_HOST == 'github.com':
-    GITHUB_API = 'api.github.com/user'
+    GITHUB_API = 'api.github.com'
 else:
-    GITHUB_API = '%s/api/v3/user' % GITHUB_HOST
+    GITHUB_API = '%s/api/v3' % GITHUB_HOST
 
 
 def _api_headers(access_token):
@@ -118,7 +118,7 @@ class GitHubOAuthenticator(OAuthenticator):
         access_token = resp_json['access_token']
         
         # Determine who the logged in user is
-        req = HTTPRequest("https://%s" % GITHUB_API,
+        req = HTTPRequest("https://%s/user" % GITHUB_API,
                           method="GET",
                           headers=_api_headers(access_token)
                           )
--- a/oauthenticator/github.py
+++ b/oauthenticator/github.py
@@ -12,6 +12,7 @@ import re
 from tornado.auth import OAuth2Mixin
 from tornado import gen, web
 
+import requests
 from tornado.httputil import url_concat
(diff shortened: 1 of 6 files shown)


jbweston:
Apologies for the rebase -- I only noticed that there were conflicts after pushing

minrk:
@jbweston sorry for leaving you hanging so long. I've started with some testing utilities in #66, which should give you a place to test what you are doing here. It's not the easiest to use, I realize, but it might be enough.

manics:
`GitHubOAuthenticator` with `github_organization_whitelist` set fails for me due to insufficient permissions on the Github token:

```
[E 2017-03-30 14:42:41.766 JupyterHub web:1548] Uncaught exception GET /jupyter/hub/oauth_callback?code=XXXXXXXXXXXXXXXXXXXX (::ffff:192.168.2.9)
    HTTPServerRequest(protocol='http', host='XXXXXXXXXX', method='GET', uri='/jupyter/hub/oauth_callback?code=XXXXXXXXXXXXXXXXXXXX', version='HTTP/1.1', ...
      File "/opt/conda/lib/python3.5/site-packages/tornado/web.py", line 1469, in _execute
        result = yield result
      File "/opt/conda/lib/python3.5/site-packages/oauthenticator/oauth2.py", line 58, in get
        username = yield self.authenticator.get_authenticated_user(self, None)
      File "/opt/conda/lib/python3.5/site-packages/jupyterhub/auth.py", line 171, in get_authenticated_user
        username = yield self.authenticate(handler, data)
      File "/opt/conda/lib/python3.5/site-packages/oauthenticator/github.py", line 134, in authenticate
        user_in_org = yield self._check_organization_whitelist(username, access_token)
      File "/opt/conda/lib/python3.5/site-packages/oauthenticator/github.py", line 149, in _check_organization_whitelist
    tornado.httpclient.HTTPError: HTTP 403: Forbidden
```

If I add some additional logging to `GitHubOAuthenticator._check_organization_whitelist` I can see the GitHub API request:
`[D 2017-03-30 14:46:11.719 JupyterHub github:148] url:https://api.github.com/user/orgs headers:[('User-Agent', 'JupyterHub'), ('Authorization', 'token XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX'), ('Accept', 'application/json')]
`

Checking this with curl fails:
```
$ curl -H 'Authorization: token XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX' https://api.github.com/user/orgs
{
  "message": "You need at least read:org scope or user scope to list your organizations.",
  "documentation_url": "https://developer.github.com/v3/orgs/#list-your-organizations"
}
```

The default scope for the GitHub oauth token [doesn't allow listing of a user's organisations](https://developer.github.com/v3/orgs/#list-your-organizations).

It's also worth noting that [users can change the scope](https://developer.github.com/v3/oauth/#requested-scopes-vs-granted-scopes):

> It is important to handle error cases where a user chooses to grant you less access than you originally requested. For example, applications can warn or otherwise communicate with their users that they will see reduced functionality or be unable to perform some actions.

jbweston:
@manics thanks for the feedback. I haven't been busy with other stuff the past few weeks, but will sit down to look at this an evening next week.

oauthenticator/github.py
```diff
@@ -88,18 +119,40 @@ def authenticate(self, handler, data=None):
...
+
+
+    @gen.coroutine
+    def _check_organization_whitelist(self, username, access_token):
+        http_client = AsyncHTTPClient()
+        headers = _api_headers(access_token)
+        # Get all the user's organizations
+        next_page = "https://%s/orgs" % GITHUB_API
+        user_orgs = set()
+        while next_page:
```
minrk: This can check the orgs on each page and return immediately if a match is found, rather than walking through all pages every time:

```python
while next_page:
    ...
    user_orgs = set(entry["login"] for entry in resp_json)
    if self.github_organization_whitelist & user_orgs:
        return True
return False
```

jbweston:
Just rebasing on `master`

oauthenticator/github.py
```diff
@@ -27,6 +28,30 @@
...
+
+
+def _get_next_page(response):
+    # Github uses Link headers for pagination.
+    # See https://developer.github.com/v3/#pagination
+    link_header = response.headers.get('Link')
+    if not link_header:
+        return
+    # parse the Link header; is this regex general enough?
+    link_regex = r',? *< *(.*?) *> *; *rel="(.*?)"'
```
minrk: To avoid maintaining our own implementation, we could vendor requests' [own implementation of this](https://github.com/kennethreitz/requests/blob/v2.13.0/requests/utils.py#L670):

```python
for link in parse_header_links(link_header):
    if link.get('rel') == 'next':
        return link['url']
```

jbweston: I didn't know requests was a dependency, but now I look it seems that it is a dependency of jupyterhub, which *is* a dependency

minrk: I said `vendor`, by which I meant copy/paste the function (with source link), since it's small. I don't think we need to rely on it as a dependency

jbweston: Isn't it preferable to just import it from `requests`, given that it is a dependency of `jupyterhub`? This is what I have done for now. If vendoring is the preferable solution in this case, then I will do this

oauthenticator/github.py
jbweston: Good point! I have changed this in the most recent version

jbweston:
@minrk Sorry for the inordinate amount of time that it took me to get around to this. I believe that I am finished (modulo vendoring the link header parsing, depending on your opinion on my comment).

jbweston:
The build appears to be broken, however the failure occurs in the `mediawiki` tests. I have not touched and of the concerned files in this branch, and all the tests pass locally...

oauthenticator/github.py
minrk: I'm okay either way, though I have a *slight* preference for copying it in, since it's so small. However, if it's imported here let's make sure to call it an explicit dependency of this package, so we aren't assuming the upstream dependency in JupyterHub going forward.

minrk:
I fixed the mediawiki tests in #82, merging here. Thanks!

PR #60, opened 2016-12-09
Adding oauth2 generic RFC authentication

dmvieira:
Following https://tools.ietf.org/html/rfc6749


oauthenticator/oauth2.py
```diff
@@ -1,104 +1,135 @@
...
-    authenticate (method takes one arg - the request handler handling the oauth callback)
-    """
-    
-    login_service = 'override in subclass'
-    oauth_callback_url = Unicode(
-        os.getenv('OAUTH_CALLBACK_URL', ''),
+
+class OAuth2EnvMixin(OAuth2Mixin):
+    _OAUTH_ACCESS_TOKEN_URL = os.environ.get('OAUTH2_TOKEN_URL', '')
+    _OAUTH_AUTHORIZE_URL = os.environ.get('OAUTH2_AUTHORIZE_URL', '')
```
minrk: Env vars are the only thing really available at this level, so this is fine.

oauthenticator/oauth2.py
```diff
@@ -1,104 +1,135 @@
 """
-Base classes for Custom Authenticator to use GitHub OAuth with JupyterHub
```
minrk: To avoid confusion / compatibility, can you put the new class in `oauthenticator/generic.py`, rather than changing what `oauthenticator.oauth2` provides and means?

oauthenticator/oauth2.py
```diff
@@ -1,104 +1,135 @@
...
+    ).tag(config=True)
+
+    userdata_method = Unicode(
+        os.environ.get('OAUTH2_USERDATA_METHOD', 'GET'),
+        config=True,
+        help="Userdata method to get user data login information"
+    )
+
+    token_url = Unicode(
+        os.environ.get('OAUTH2_USERDATA_METHOD', 'GET'),
```
minrk: copy/paste USERDATA_METHOD

oauthenticator/__init__.py
```diff
@@ -1,6 +1,7 @@
 # include github, bitbucket, google here for backward-compatibility
 # don't add new oauthenticators here.
 from .oauth2 import *
+from .generic import *
```
minrk: See the comment above for 'don't add new things here'. Authenticators should be imported directly from their submodule, so that we don't import all the unused Authenticators as this package grows.

oauthenticator/generic.py
```diff
@@ -0,0 +1,135 @@
...
+
+class OAuth2LoginHandler(OAuthLoginHandler, OAuth2EnvMixin):
+    pass
+
+
+class OAuth2OAuthenticator(OAuthenticator):
+
+    login_service = "OAuth2"
+
+    client_id_env = 'OAUTH2_CLIENT_ID'
```
minrk: I think this can leave the defaults from the parent class, of `OAUTH_...`

dmvieira:
I think it's now ready ;) Really thx

dmvieira:
I'm using here and it's working ;)

minrk:
Thanks!

PR #62, opened 2017-01-21
fix typo in doc about environment variable

shusson:


yuvipanda:
Thanks for the patch!

PR #64, opened 2017-02-03
Added option that allows self-signed ssl cert (for GitLab)

zh3w4ng:
I use self-signed certs for my intranet server. This fix is necessary to avoid custom authentication (gitlab in my case). Please review. 🙇

### DIFF
--- a/oauthenticator/gitlab.py
+++ b/oauthenticator/gitlab.py
@@ -39,6 +39,7 @@ class GitLabOAuthenticator(OAuthenticator):
 
     client_id_env = 'GITLAB_CLIENT_ID'
     client_secret_env = 'GITLAB_CLIENT_SECRET'
+    validate_server_cert_env = 'VALIDATE_SERVER_CERT'
     login_handler = GitLabLoginHandler
 
     @gen.coroutine
@@ -62,6 +63,9 @@ class GitLabOAuthenticator(OAuthenticator):
             redirect_uri=self.oauth_callback_url
         )
 
+
+        validate_server_cert = self.validate_server_cert
+
         url = url_concat("%s/oauth/token" % GITLAB_HOST,
                          params)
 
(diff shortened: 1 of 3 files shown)


oauthenticator/gitlab.py
```diff
@@ -39,6 +39,7 @@ class GitLabOAuthenticator(OAuthenticator):
 
     client_id_env = 'GITLAB_CLIENT_ID'
     client_secret_env = 'GITLAB_CLIENT_SECRET'
+    validate_server_cert_env = 'VALIDATE_SERVER_CERT'
```
minrk: You can leave this line out, since it's inherited from the parent class

oauthenticator/oauth2.py
```diff
@@ -94,6 +94,11 @@ def _client_id_default(self):
     def _client_secret_default(self):
         return os.getenv(self.client_secret_env, '')
 
+    validate_server_cert_env = 'VALIDATE_SERVER_CERT'
```
minrk: Maybe call this `OAUTH_TLS_VERIFY` to follow the `DOCKER_TSL_VERIFY` example.

oauthenticator/oauth2.py
```diff
@@ -94,6 +94,11 @@ def _client_id_default(self):
...
+    validate_server_cert = Bool(config=True)
+    def _validate_server_cert_default(self):
+        return os.getenv(self.validate_server_cert_env, True)
```
minrk: Since this is a bool, we'll want to treat '0' as False, other set values as True, and the default ('' or unset) as True:

```python
env_value = os.getenv(self.validate_server_cert_env, '')
if env_value == '0':
    return False
else:
    return True
```

zh3w4ng:
Updated and tested. 😺  👂

minrk:
Thanks!

PR #66, opened 2017-02-15
wip: testing framework

minrk:
Adds basic mocking utilities for exercising OAuthenticators.

- [x] utilities for mocking OAuth providers
- [x] auth0
- [x] bitbucket
- [ ] cilogon
- [x] generic
- [x] github
- [x] gitlab
- [x] google
- [ ] mediawiki
- [x] openshift

Since this mocks out the OAuth provider, the functionality of the OAuthenticator talking to the provider correctly is not tested except by assuming that the mocked responses are accurate.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,2 @@
+[report]
+omit = *test*
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,4 @@
 [flake8]
 select = F
-ignore = E,W,C
+ignore = E,W,C,F841
 exclude = __init__.py
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,4 @@
 [flake8]
 select = F
-ignore = E,W,C,F841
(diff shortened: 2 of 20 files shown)


minrk:
@ryanlovett I moved a few things around in the Google OAuthenticator here, to make it reuse more code and be more consistent with the others. Can you verify that I haven't broken it when you get the chance?

minrk:
I've got everything working here for the normal cases. Two authenticators remain untested:

- cilogon
- mediawiki

and this is largely because they have very little in common with the others, and probably warrant their own PRs. I'm happy with this one as-is.

minrk:
@ryanlovett I managed to setup Google OAuth myself to test and verified that it does work.

ryanlovett:
Sorry for the delay! I'm glad you were able to verify regardless. :)

minrk:
No problem. I assumed it would be harder to setup a Google OAuth instance to test than it was.

PR #67, opened 2017-02-16
test mediawiki

minrk:
- move more MediaWiki logic onto Authenticator, so common handler logic can be reused
- make executor reusable so it isn't recreated on every request

- [x] tests authenticate()
- [x] test LoginHandler.get()

### DIFF
--- a/oauthenticator/mediawiki.py
+++ b/oauthenticator/mediawiki.py
@@ -97,7 +97,7 @@ class MWOAuthenticator(OAuthenticator):
 
         consumer_token = ConsumerToken(
             self.client_id,
-            self.client_secret
+            self.client_secret,
         )
 
         handshaker = Handshaker(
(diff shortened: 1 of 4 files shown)


oauthenticator/mediawiki.py
```diff
@@ -73,61 +67,53 @@ def get(self):
...
-            user = self.find_user(username)
-            if user is None:
-                user = orm.User(name=username, id=identity['sub'])
-                self.db.add(user)
-                self.db.commit()
-            self.set_login_cookie(user)
-            self.redirect(url_path_join(self.hub.server.base_url, 'home'))
+            # this shouldn't be necessary anymore,
+            # but keep for backward-compatibility
+            return identity['username'].replace(' ', '_')
```
yuvipanda: I don't entirely remember, but MW is... very weird with usernames. The user Yuvipanda is the same as the user yuvipanda (first character of usernames is case insensitive) but different from YuviPanda (rest of characters in usernames are case sensitive!). Also user named 'Some User' is equivalent to a user named 'Some_User' and 'some_User'... 

There's theoretically a canonical form, but I've had it come back with spaces instead of _s sometimes, and it's confusing. What this really needs is for me to look at the username canonicalization code in the MW php and port it here. Not a requirement for here tho :D

yuvipanda:
<3 this is great! Thanks!

PR #72, opened 2017-03-16
Adding Globus Auth as a possible ID provider for jupyterhub

briedel:


oauthenticator/globus.py
```diff
@@ -0,0 +1,123 @@
...
+        id_token = tokens.decode_id_token(client)
+        username = id_token.get('preferred_username')
+        if not username.endswith('@globusid.org'):
+            raise HTTPError(403, ("You are not signed in "
+                                  "to your {} account.".format("Globus ID")))
+        # Need to return a username without the "email" ending
+        username = username.split('@')[0]
+        # Checking the user exists... Paranoia!
+        try:
+            pwd.getpwnam(username)
```
minrk: OAuthenticators don't generally require local users to exist. That's what the Local... variations are for. You can see examples of these in the other implementations here.

briedel: I will remove it

globus-requirements.txt
```diff
@@ -0,0 +1,6 @@
+# extra requirements for globus
+-r ./requirements.txt
+requests>=2.0.0,<3.0.0
+six==1.10.0
```
briedel: I took it straight from the globus_sdk requirements. I will ask the dev

globus-requirements.txt
```diff
@@ -0,0 +1,6 @@
...
+python-jose
+pycrypto
```
briedel: `python-jose` and `pycrypto` are hidden dependencies. The sdk relies on them for some of the exchanges

oauthenticator/globus.py
```diff
@@ -0,0 +1,123 @@
...
+    login_service = "Globus"
+    login_handler = GlobusLoginHandler
+
+    def get_globus_app_id(self):
+        """
+        Getting the custom configuration value globus_app_id.
+        globus_app_id is the UUID of the Globus app to authenticate
+        against
+        """
+        return self.config.GlobusOAuthenticator.globus_app_id
```
briedel: Looking at others, could I just rename it as `client_id` and `client_secret`? The mediawiki and  bitbucket example use those

briedel:
I tried putting in the tests, but I am little confused what to pass to the `setup_oauth_mock` function.

globus-requirements.txt
sirosen: Weighing in as one of the Globus SDK maintainers.
I can't remember why we pinned the exact version of `six`. We can probably be more permissive. I'll open an issue to dig into it when I have spare cycles.

globus-requirements.txt
sirosen: You should be okay to only depend on `python-jose`. `pycrypto` is just a dependency of `python-jose`.

I want to get around to making this doable via an extra, so you could do `globus_sdk[jwt]`, for JWT support, but the time to tinker with setuptools hasn't materialized.

minrk: OK, then I think this requirements.txt can be updated to have just those:

```
-r ./requirements.txt
python-jose
globus_sdk
```

Any other dependencies globus_sdk may have should be specified by that package.

oauthenticator/globus.py
```diff
@@ -0,0 +1,101 @@
...
+        self.log.info('globus redirect: %r', redirect_uri)
+        self.authorize_redirect(client)
+
+
+class GlobusOAuthenticator(OAuthenticator):
+
+    login_service = "Globus"
+    login_handler = GlobusLoginHandler
+
+    def globus_portal_client(self):
```
minrk: SHould this be a property instead of a method? Is there a reason to create a new Client on every request?

oauthenticator/tests/test_globus.py
```diff
@@ -0,0 +1,35 @@
...
+def user_model(username):
+    """Return a user model"""
+    return {
+        'login': username,
+    }
+
+@fixture
+def globus_client(client):
+    setup_oauth_mock(client,
+        host=['github.com', 'api.github.com'],
```
minrk: these should be the host(s) that the globus client will access

oauthenticator/tests/test_globus.py
```diff
@@ -0,0 +1,35 @@
...
+        access_token_path='/login/oauth/access_token',
+        user_path='/user',
```
minrk: These should be the paths that the globus client accesses. e.g. if globus accesses `globus.org/oauth/access_token` (I don't know the actual domains and paths, you will need to find them), this would be:

```python
host=['globus.org'],
access_token_path='/oauth/access_token/',
```

oauthenticator/tests/test_globus.py
briedel: @sirosen Do you know which host to use here?

oauthenticator/tests/test_globus.py
briedel: Globus SDK generates those internally. I can construct them, but they might break down the road. Should I just invoke the Globus sdk here?

oauthenticator/globus.py
briedel: I based this on what was done in the GitHub and Google version
```
    login_service = "GitHub"
...
    login_handler = GitHubLoginHandler
```

https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/google.py#L41-L42

oauthenticator/tests/test_globus.py
sirosen: The SDK helper for this will use `auth.globus.org` internally, so that's what you'll want if you're building this out like other oauthenticators for Jupyterhub.

However, if you're using [`get_authorize_url`](http://globus-sdk-python.readthedocs.io/en/stable/clients/auth/#globus_sdk.AuthClient.oauth2_get_authorize_url), it doesn't really matter what you set here, since the value will never get used.

I'm not sure what the norms of Jupyterhub around this type of library usage are. Should it be `host=None`?

oauthenticator/tests/test_globus.py
sirosen: I'm more in favor of using the SDK lib than not using it ( :open_mouth: ), as a general rule, but it's not obvious that it will actually make your life easier here.

You can use the lib as a reference and go straight to the [API docs](https://docs.globus.org/api/auth/reference/) -- as far as Globus is concerned, integrating directly against this API is totally fine.

We only built a lib because we consider OAuth2 to be unnecessary knowledge for most people developing against our APIs. Clearly, building this integration requires decent familiarity with OAuth2, so it's not obvious that doing things directly is in any way inferior.

minrk:
Merged as #83

PR #75, opened 2017-04-18
Integrate OkpyAuthenticator to OAuthenticator

stevenwuyinze:
- Updated README.md to include information on OkpyAuthenticator
- Imported okpy.py in init file.
- Created an okpy.py.

### DIFF
--- a/oauthenticator/okpy.py
+++ b/oauthenticator/okpy.py
@@ -2,7 +2,6 @@
 Custom Authenticator to use okpy OAuth with JupyterHub
 """
 import json
-import os
 import base64
 
 from tornado.auth import OAuth2Mixin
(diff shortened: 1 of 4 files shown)


oauthenticator/okpy.py
```diff
@@ -0,0 +1,109 @@
...
+        code = handler.get_argument("code", False)
+        if not code:
+            raise web.HTTPError(400, "Authentication Cancelled.")
+        http_client = AsyncHTTPClient()
+        auth_request = handler.get_auth_request(code)
+        response = yield http_client.fetch(auth_request)
+        if not response:
+            self.clear_all_cookies()
+            raise HTTPError(500, 'Authentication Failed: Token Not Acquired')
+        state = json.loads(response.body.decode('utf8', 'replace'))
```
yuvipanda: Where else can we store them? This is the only place we can store them that survives hub restarts and user container restarts. If the db is blown away, that's ok - it gets repopulated on login. I don't think there's any way other than this, and there's other people wanting to do the exact same thing in https://github.com/jupyterhub/jupyterhub/issues/1063

oauthenticator/okpy.py
```diff
@@ -0,0 +1,109 @@
...
+            user = self.user_from_username(username)
+            user.auth_state = state
+            self.db.add(user)
+            self.db.commit()
+            self.set_login_cookie(user)
+            self.redirect(url_path_join(self.hub.server.base_url, 'home'))
+        else:
+            raise web.HTTPError(403, "Authentication Failed.")
+
+    def get_user_info_request(self, access_token):
```
yuvipanda: Ideally we'll have github.com/jupyterhub/jupyterhub#1063 sorted and that'll simplify this a lot :D

oauthenticator/okpy.py
minrk: This is the downside of auth_state, and why I don't actually like it very much and think maybe we should remove it. The Hub has a policy of never storing auth info in cleartext in the db, because if the db is compromised everything is a mess. If Authenticators start doing this, that is violated, so I don't want any Authenticators in this repo to set a bad example by doing so.

What's more important is the ultimate goal of passing a token (or certs, etc.) from the upstream auth service to the Spawner on start (not on login). Storing this token in auth_state isn't a particularly good idea because the lifecycle of the data is wrong. Really, `start` should be the action where the token is requested, not login. Keeping that data in memory gives it a better lifetime, but one problem is left unsolved: when to retrieve a new token for a user with a valid login cookie. This is still true when the token is stored in auth_state, though, because there is no check for whether the token is valid. So if the token is revoked, the behavior is the same as the only-in-memory case where the token is lost (but worse, because it's present and thinks it's valid).

What we really need is a mechanism to request a token on start - such as allow `pre_spawn_start()` to be a more complicated action that may involve OAuth redirects. I'm not sure that's feasible, though. What should be feasible and solve this problem is if the Authenticator could intercede and double-check "Is the auth for this user still valid" and trigger re-login, that would also completely remove the need for auth_state in any case I'm aware of.

stevenwuyinze:
Hi All. Thanks for the comments and conversations going on here. I've refactored the AuthHandler, and left out the token/state saving step.

PR #76, opened 2017-04-18
Add Okta Authenticator

frankhsu:
This adds Okta Authenticator using OpenID Connect
http://developer.okta.com/docs/api/resources/oidc.html


frankhsu:
nevermind believe the generic works well. must've missed that one.

PR #81, opened 2017-05-16
Updating Generic acording to OpenID Connect specifications

rafael-ladislau:
According to OpenID specs, the token request should be made using an HTTP POST and form serialization ([http://openid.net/specs/openid-connect-basic-1_0.html#TokenRequest](http://openid.net/specs/openid-connect-basic-1_0.html#TokenRequest)). The Generic Authenticator was sending the parameters as a query string. With this change, the Generic Authenticator is working with Keycloak.


yuvipanda:
woohoo KeyCloak!

PR #82, opened 2017-05-19
fix tests

minrk:
- pre-emptive change to mocks for JupyterHub 0.8
- pin mwoauth to < 0.3 because there appears to be a bug where it assumes responses will be JSON that aren't


yuvipanda:
can you describe the mwoauth bug in a little more details so I can report it upstream?

minrk:
Sorry, forgot to link. I've been working on it and think I have a fix: https://github.com/mediawiki-utilities/python-mwoauth/issues/27

yuvipanda:
we should unpin mwoauth now that your patch was merged and uploaded to pypi :)

PR #83, opened 2017-05-24
Add Support for Globus OAuth

NickolausDS:
The following adds support for Globus Auth, allowing users to login with their GlobusID. Alternatively, this authenticator supports other organizations supported by Globus Auth. For instance if someone wanted to setup their Jupyterhub server and restrict logins to the University of Chicago, they can configure their app to do so. The full list of institutions supported can be found [here](https://www.globus.org/app/transfer). 

In addition to Authentication, Globus also provides a transfer service for transferring data across nodes. This Authenticator is setup to procure `access_tokens` for users when they login, allowing them to easily transfer data from their python notebook with little hassle. 

I followed the discussion [here](https://github.com/jupyterhub/jupyterhub/issues/1063) on storing tokens in auth_state, and in some outside discussion decided `auth_state` was the best place to do it. If security is a concern, admins can easily turn off transfer `access_tokens` in their `jupyterhub_config.py` file. By default, the Authenticator excludes user Auth `access_tokens`.

### DIFF
--- a/globus-requirements.txt
+++ b/globus-requirements.txt
@@ -1,4 +1,3 @@
 # extra requirements for globus
 -r ./requirements.txt
-python-jose
-globus_sdk
+globus_sdk[jwt]>=1.0.0,<2.0.0
--- a/globus-requirements.txt
+++ b/globus-requirements.txt
@@ -4,3 +4,4 @@ requests>=2.0.0,<3.0.0
 six==1.10.0
 python-jose
 pycrypto
+globus_sdk
--- a/globus-requirements.txt
+++ b/globus-requirements.txt
@@ -1,0 +1,6 @@
+# extra requirements for globus
+-r ./requirements.txt
+requests>=2.0.0,<3.0.0
+six==1.10.0
+python-jose
+pycrypto
--- a/globus-requirements.txt
+++ b/globus-requirements.txt
@@ -1,7 +1,4 @@
 # extra requirements for globus
 -r ./requirements.txt
-requests>=2.0.0,<3.0.0
-six==1.10.0
 python-jose
-pycrypto
 globus_sdk
--- a/oauthenticator/globus.py
+++ b/oauthenticator/globus.py
@@ -11,9 +11,9 @@ from .oauth2 import OAuthLoginHandler, OAuthenticator
 try:
     import globus_sdk
 except:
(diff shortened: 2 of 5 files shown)


globus-requirements.txt
```diff
@@ -0,0 +1,3 @@
+# extra requirements for globus
+-r ./requirements.txt
+globus_sdk[jwt]>=1.0.0,<2.0.0
```
NickolausDS: I looked into doing authentication without the globus_sdk here, but most people will want to initiate file transfers using the globus_sdk from within their python notebooks, so they would end up installing it anyway.

oauthenticator/globus.py
```diff
@@ -0,0 +1,223 @@
...
+    login_handler = GlobusLoginHandler
+    callback_handler = GlobusOAuthCallbackHandler
+
+    identity_provider = Unicode(help="""Restrict which institution a user
+    can use to login (GlobusID, University of Hogwarts, etc.). This should
+    be set in the app at developers.globus.org, but this acts as an additional
+    check to prevent unnecessary account creation.""").tag(config=True)
+
+    def _identity_provider_default(self):
+        return os.getenv('IDENTITY_PROVIDER', 'globusid.org')
```
NickolausDS: If an admin wanted to change which Identity is used by their users logging in, they can configure it in their [app](developers.globus.org). However, this serves as a nice additional check that can easily be seen in the `jupyterhub_config.py` file.

oauthenticator/globus.py
```diff
@@ -0,0 +1,223 @@
...
+        self.globus_data = {}
+
+    @gen.coroutine
+    def authenticate(self, handler, data=None):
+        """
+        Authenticate with globus.org. Usernames (and therefore Jupyterhub
+        accounts) will correspond to a Globus User ID, so foouser@globusid.org
+        will have the 'foouser' account in Jupyterhub.
+        """
+        code = handler.get_argument("code")
```
NickolausDS: I went ahead and removed the error check that was removed last week, in addition to the `no_code_test` in the test code, so this should be up to date.

minrk:
Thanks!

NickolausDS:
And thanks for adding the line to `test-requirements.txt` to fix Travis, I appreciate it!

PR #85, opened 2017-05-29
Vendor link headers from 'requests' module

jbweston:
This was suggested by @minrk previously, but I did not implement it before #58 was merged.

A separate commit moves out the function for getting the 'next' link header to a common location


minrk:
Thanks!

PR #87, opened 2017-06-23
update README

### LINKED ISSUE #53: Change description to be more accurate
Currently it mentions only GitHub - but the repo has a lot more than just that now :)

minrk:
- mention all of the current providers
- support `OAUTH_CLIENT_` everywhere instead of needing separate env docs for every provider
- consolidate docs in general section that applies to all providers

cc @willingc

closes #53

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,19 @@
 
 OAuth + JupyterHub Authenticator = OAuthenticator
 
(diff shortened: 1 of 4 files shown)


willingc:
@minrk Sorry I missed this PR. It's a nice improvement 👍

PR #88, opened 2017-07-11
handle get_authenticated_user returning a dict

minrk:
which it will do in 0.8

The recent change to get_authenticated_user causes
an inscrutable `sqlalchemy.exc.InterfaceError: <unprintable InterfaceError object>` error due to passing a dict for a string arg.

cc @athornton


willingc:
Thanks @minrk. I'm going to merge. I'm thinking back to Django stuff and these errors tended to happen when there was a relationship to foreign key missing. And yes, inscrutable describes this well.

PR #95, opened 2017-07-30
Add support for returning auth_state from OAuthenticators

yuvipanda:
woohoo!

I've been waiting for this for so long :)


minrk:
Nice! We'll be able to simplify this much further for 0.8, since we added the `.login_user` method, which replaces all of the set-cookie, save state, redirect, etc.

PR #96, opened 2017-07-31
update link to Jupyter contributor guide

minrk:
we only had the IPython placeholder from before the Jupyter contributor docs

cc @willingc

I think we're in the same situation on several JupyterHub repos. I can update the rest if this seems reasonable.


willingc:
Thanks  @minrk. I'll go ahead and make the changes on the other repos. If a repo doesn't have a CONTRIBUTING.md, I'll add one since GH, Libraries.io, etc. now look for its presence.

minrk:
Perfect, thanks!

PR #97, opened 2017-08-01
[mediawiki] Store extra info in auth_state

yuvipanda:


yuvipanda:
@minrk questions!

1. How should the things in auth_state be named? Up to each authenticator? Do we clear it when we switch authenticators?
2. Should authenticators indiscriminately put stuff in auth_state? Or have traitlets on them that provide more fine grained control?

minrk:
> How should the things in auth_state be named? Up to each authenticator?

100% up to the Authenticator. As it is now, nothing will ever access this info other than the Authenticator. I was thinking that it would be up to `Authenticator.pre_spawn_start` to pass any relevant info to the Spawner, and I was planning to sketch out a version of this here as a test case. Alternately, we could bless a particular key, say `auth_state.env`, for the simple case of environment variables to pass to the spawning environment, and support retrieving that env by default in all Spawners (it would require an async `get_env`, which might be tricky to do in a backward-compatible way).

> Should authenticators indiscriminately put stuff in auth_state?

I added [`Authenticator.enable_auth_state`](https://github.com/jupyterhub/jupyterhub/pull/1178/files#diff-8263a63fc781091dc1195f33f5ea7fc4R41) config for this, so Authenticators shouldn't return auth_state if `enable_auth_state` is disabled. If they do and encryption is unavailable, login will always fail due to encryption being requested but unavailable. I realize now that I could move that handling to Hub-side, so that the Hub drops auth_state if this flag is False and Authenticators don't need to handle it. What do you think?

oauthenticator/mediawiki.py
```diff
@@ -121,7 +121,16 @@ def authenticate(self, handler, data):
         if identity and 'username' in identity:
             # this shouldn't be necessary anymore,
             # but keep for backward-compatibility
-            return identity['username'].replace(' ', '_')
```
minrk: I think it has to still return a string on 0.7 or it will error.

minrk:
> Or have traitlets on them that provide more fine grained control?

I think this is going to be up to the given Authenticator whether such control makes sense. I suspect most will be covered by the single `enable_auth_state` flag.

yuvipanda:
I've changed this out a little - now auth_state contains the entire JSON dict that is the user's identity as returned by mediawiki rather than picking and choosing. This puts control in the hands of the admins, which is the right place for this to be IMO.

@minrk did you implement your change re: enable_auth_state being dealt with by hub itself?

minrk:
> This puts control in the hands of the admins, which is the right place for this to be IMO.

Agreed. Maybe we should do the same in GitHub to set an example.

> did you implement your change re: enable_auth_state being dealt with by hub itself?

Yes, so it's fine to return the whole thing without checking, as long it doesn't require extra work to retrieve.

yuvipanda:
@minrk yeah, I agree we should do the same for GitHub too.

PR #99, opened 2017-08-01
Add additional features to GitHub OAuth2

athornton:
As documented in README.md, there are a few feature flags you can set (as environment variables) which govern the scope of the token that GitHub OAuth asks for.  Based on that, it creates an auth_state entry in the user dictionary which may have the GitHub token, GitHub ID, GitHub organization membership (which maps to the orgs' IDs), GitHub "real name", and GitHub email address. 

I am using all of these in my container for user provisioning.  It allows setting up a .git-credentials that is preconfigured to be able to push to GitHub, for instance, and capturing UID/GID information allows you to provision users and groups with globally consistent uid/gids, which makes shared filesystems a whole lot easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,7 +234,7 @@ c.GlobusOAuthenticator.identity_provider = 'globusid.org'
 
 If you want to use a _Linked Identity_ such as `malcolm@universityofindependence.edu`,
 go to your [App Developer page](http://developers.globus.org) and set
-*Required Identity Provider* for your app to _<Your University>_, and set the
+*Required Identity Provider* for your app to `<Your University>`, and set the
 following in the config:
 
 ```python
--- a/README.md
+++ b/README.md
@@ -102,37 +102,46 @@ You can use your own Github Enterprise instance by setting the `GITHUB_HOST` env
 
 ### GitHub-specific features
 
-There are four environment variables to use to configure additional
-GitHub features: `GITHUB_USE_ORGANIZATIONS`, `GITHUB_USE_PUSH_TOKEN`,
-`GITHUB_USE_PRIVATE_PUSH_TOKEN`, and `GITHUB_USE_EMAIL`.
+Additional features are turned on by overriding the authenticator's
+`scope` list with particular GitHub scopes:
 
-Each of these turns on a feature if it is set and requests additional
-scope on the GitHub token requested.
+https://developer.github.com/apps/building-integrations/setting-up-and-registering-oauth-apps/about-scopes-for-oauth-apps/
 
-`GITHUB_USE_ORGANIZATIONS` enables the use of GitHub organizations to
-allow provisioning of backend gids, which requires `read:org` scope to
-iterate through the user's organizations and map their names to their id
-numbers.
+Some example scopes you might want:
 
-`GITHUB_USE_PUSH_TOKEN` requests `public_repo` access in order to push
-code into public repositories--we use magic on the backend to cache
-the GitHub token and set up .git-credentials with it.
+`read:org` grants access to the users' organizations.  This is handy if
+you want to use GitHub organizations in your backend environment as Unix
+groups for collaboration purposes.  Having globally consistent UIDs
+(from the GitHub ID) and GIDs (from the organization IDs) makes access
+permissions on shared storage much easier.
 
-`GITHUB_USE_PRIVATE_PUSH_TOKEN` does the same but with `repo` access, so
-it can access both public and private repositories. 
+`public_repo` allows read and write of public repositories; if you want
+to pass the token back to your Lab or Notebook to automatically
+provision git pushes to GitHub magically working, you will want this.
 
-`GITHUB_USE_EMAIL` looks at the GitHub email field; this is used to set
-up the user email address for GitHub in conjunction with the push token.
-It uses "user:email" scope but it's less than useful since private email
-addresses are still not visible.
+`repo` does the same for private repositories too.
 
-These are all stored in the authenticator's `auth_state` structure, so
-you'll need to enable `auth_state` and install the Python `cryptography`
-package to be able to use these.
+The additional fields exposed by expanded scope are all stored in the
+authenticator's `auth_state` structure, so you'll need to enable
+`auth_state` and install the Python `cryptography` package to be able to
+use these.
 
-You will also need to subclass your spawner to be able to pull these
-fields out of `auth_state` and use them to provision your Notebook or
-Lab user.
+We currently use the following fields: 
+
+* `uid` is an integer set to the GitHub account ID.
+* `name` is the full name GitHub knows the user by.
+* `email` is the publicly visible email address (if any) for the user.
+* `auth_token` is the token used to authenticate to GitHub.
+* `organization_map` is a dict mapping the users' organization
+   memberships to the organization IDs, intended to be used to construct
+   group ID mappings for the user.
+
+If you are going to use this expanded user information, you will
+need to subclass your spawner to be able to pull these fields out
+of `auth_state` and use them to provision your Notebook or Lab
+user.
+
+
 
 ## GitLab Setup
 
--- a/README.md
+++ b/README.md
@@ -233,7 +233,7 @@ c.GlobusOAuthenticator.identity_provider = 'globusid.org'
 ```
 
 If you want to use a _Linked Identity_ such as `malcolm@universityofindependence.edu`,
-go to your [App Developer page](http://www.developers.globus.org) and set
+go to your [App Developer page](http://developers.globus.org) and set
 *Required Identity Provider* for your app to _<Your University>_, and set the
 following in the config:
 
--- a/README.md
+++ b/README.md
@@ -4,25 +4,27 @@ OAuth + JupyterHub Authenticator = OAuthenticator
 
 OAuthenticator currently supports the following authentication services:
 
-- Auth0
-- Bitbucket
-- CILogon
-- GitHub
-- Google
-- MediaWiki
-- Okpy
-- OpenShift
+- [Auth0](oauthenticator/auth0.py)
+- [Bitbucket](oauthenticator/bitbucket.py)
+- [CILogon](oauthenticator/cilogon.py)
+- [GitHub](#github-setup)
+- [GitLab](#gitlab-setup)
+- [Globus](#globus-setup)
+- [Google](#google-setup)
+- [MediaWiki](oauthenticator/mediawiki.py)
+- [Okpy](#okpyauthenticator)
+- [OpenShift](#openshift-setup)
 
-and a generic implementation, which you can use with any provider.
+A [generic implementation](oauthenticator/generic.py), which you can use with
+any provider, is also available.
 
 ## Examples
 
 For an example docker image using OAuthenticator, see the [example](example)
 directory.
 
-There is [another
-example](https://github.com/jupyterhub/dockerspawner/tree/master/examples/oauth)
-for using GitHub OAuth to spawn each user's server in a separate docker
+[Another example](https://github.com/jupyterhub/dockerspawner/tree/master/examples/oauth)
+is using GitHub OAuth to spawn each user's server in a separate docker
 container.
 
 ## Installation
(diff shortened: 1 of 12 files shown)


oauthenticator/github.py
```diff
@@ -43,23 +47,80 @@ class GitHubMixin(OAuth2Mixin):
...
+    structure, so you'll need to enable `auth_state` and install the
+    Python `cryptography` package to be able to use these.
+
+    You will also need to subclass your spawner to be able to pull
+    these fields out of `auth_state` and use them to provision your
+    Notebook or Lab user.
+
+    """
+
+    use_organizations = False
```
yuvipanda: These should all probably be traitlets rather than environment variables directly.

oauthenticator/github.py
```diff
@@ -43,23 +47,80 @@ class GitHubMixin(OAuth2Mixin):
...
+        use_organizations = True
+    if os.environ.get('GITHUB_USE_PUSH_TOKEN'):
+        use_push_token = True
+    if os.environ.get('GITHUB_USE_PRIVATE_PUSH_TOKEN'):
+        use_private_push_token = True
+    if os.environ.get('GITHUB_USE_EMAIL'):
+        use_email = True
+    scope = []
+    if use_organizations:
+        scope.append("read:org")
```
yuvipanda: Maybe have a traitlet just be a list of scopes?

oauthenticator/github.py
```diff
@@ -109,20 +172,50 @@ def authenticate(self, handler, data=None):
...
+        auth_state = {}
+        auth_state["username"] = username
+        # We may want to do user provisioning in the server container.
+        #  This next bit is about that.
+        #  1) make the username look Unixy
+        #  2) use the GitHub ID as the uid
+        #  3) set list of orgs/gids
+        #  4) set up name/email for .gitconfig
+        safe_chars = set(string.ascii_lowercase + string.digits)
+        safe_username = ''.join(
```
yuvipanda: Don't we already do this in JupyterHub? (I'm not sure).

yuvipanda:
See #97 for similar discussion

yuvipanda:
It'll also be very helpful if you can describe your ideal end user workflow!

oauthenticator/github.py
minrk: Yes, there's `Authenticator.normalize_username`, which does this already on the name returned from this.

oauthenticator/github.py
```diff
@@ -109,20 +172,50 @@ def authenticate(self, handler, data=None):
...
+            [s if s in safe_chars else '-' for s in username.lower()])
```
minrk: It's probably best to use the existing `Authenticator.normalize_username` here or something like `escapism`, but I'm not sure what `canonicalname` would be for and I'd rather remove it. As it is now, this allows collisions on usernames which have an unsafe character in the same position.

oauthenticator/github.py
minrk: Yes, I think oauth scopes being a trait on the base OAuthenticator, and each provider can define their own defaults.

oauthenticator/github.py
minrk: 👍 I think traitlets are the right way to go for this config. I'd like to limit environment variable config to the credentials.

athornton:
Will circle back to this today or tomorrow.  Changes seem unobjectionable.  Escapism is used in KubeSpawner already, I think.  

In my setup I need canonicalname to create a back-end user that seems reasonable for the back-end container, but it does indeed have a collision risk, and I haven't thought about how to mitigate that.  So maybe I should drop it from oauthenticator and just pass the canonicalified version back to my spawner, thereby assuming that risk myself in my jupyterhub_config (at least until I figure out a better solution).

athornton:
The tokens are long-lived, but we end up getting them with each spawn, since you get a token as part of the OAuth2 dance.  Presumably GitHub just assumes that the space of tokens is large enough that collisions are unlikely.

I guess I am not sure about whether we should make a single list option for token scope or whether we should have a set of boolean capabilities with traitlets that select the scope, or both.  GitHub token scopes are (I think) non-obvious, but on the other hand you can make a good case that if you don't understand scopes you shouldn't be implicitly setting them in the authenticator.

minrk:
> So maybe I should drop it from oauthenticator and just pass the canonicalified version back to my spawner

I think that's appropriate. It seems a little too specific to be in the base Authenticator.

> I guess I am not sure about whether we should make a single list option for token scope or whether we should have a set of boolean capabilities with traitlets that select the scope, or both.

I'd make it a single list:

    scopes = List(Unicode(), config=True)

athornton:
`scope` is already defined in oauth2.py, and it's just defined as an empty list, not a traitlet.  Trying to make it a traitlet makes the spawner unhappy about incorrect default values.  My inclination is just to leave it alone and document some things you might want to set it to.

athornton:
My ideal end-user workflow is that the user auths to JupyterHub with GitHub.  Kubernetes spawns a user container, which runs a provisioning stage (as root) to add the user (with normalized username, GitHub ID as UID, and GitHub Organizations the user is a member of as GIDs), and then becomes that user to run the containerized Lab.

README.md
```diff
@@ -100,6 +100,49 @@ You can also use `LocalGitHubOAuthenticator` to map GitHub accounts onto local u
 
 You can use your own Github Enterprise instance by setting the `GITHUB_HOST` environment variable.
 
+### GitHub-specific features
```
willingc: Hi @athornton,
 
Let's make this a separate markdown file `github_scope.md`, and add one sentence here linking to the new document.

Keeping in mind that we have a range of user experience levels, I think it would be best to keep the `README.md` focused on the basics. Additional documents can provide more detailed/advanced use cases. If needed over time, we could create Sphinx documentation.

README.md
```diff
@@ -100,6 +100,49 @@ You can also use `LocalGitHubOAuthenticator` to map GitHub accounts onto local u
...
+
```
willingc: Perhaps something along these lines:


[GitHub scopes](https://developer.github.com/apps/building-integrations/setting-up-and-registering-oauth-apps/about-scopes-for-oauth-apps/) may be used to extend the GitHub OAuthenticator. By overriding the `scope` list in the authenticator, additional features can be enabled for specific deployment needs.

README.md
```diff
@@ -100,6 +100,49 @@ You can also use `LocalGitHubOAuthenticator` to map GitHub accounts onto local u
...
+Additional features are turned on by overriding the authenticator's
+`scope` list with particular GitHub scopes:
+
+https://developer.github.com/apps/building-integrations/setting-up-and-registering-oauth-apps/about-scopes-for-oauth-apps/
+
+Some example scopes you might want:
```
willingc: Create a new section.

## Example GitHub scopes

The following GitHub scopes may be suitable for certain use cases:

README.md
```diff
@@ -100,6 +100,49 @@ You can also use `LocalGitHubOAuthenticator` to map GitHub accounts onto local u
...
+
+`read:org` grants access to the users' organizations.  This is handy if
+you want to use GitHub organizations in your backend environment as Unix
+groups for collaboration purposes.  Having globally consistent UIDs
+(from the GitHub ID) and GIDs (from the organization IDs) makes access
+permissions on shared storage much easier.
+
+`public_repo` allows read and write of public repositories; if you want
```
willingc: allows "trusted users" read and write privileges for public repositories. If you want to automatically provision `git` pushes to GitHub, you can accomplish this by passing a token to your Lab or classic Notebook instance.

README.md
```diff
@@ -100,6 +100,49 @@ You can also use `LocalGitHubOAuthenticator` to map GitHub accounts onto local u
...
+
+* `uid` is an integer set to the GitHub account ID.
+* `name` is the full name GitHub knows the user by.
+* `email` is the publicly visible email address (if any) for the user.
+* `auth_token` is the token used to authenticate to GitHub.
+* `organization_map` is a dict mapping the users' organization
+   memberships to the organization IDs, intended to be used to construct
+   group ID mappings for the user.
+
+If you are going to use this expanded user information, you will
```
willingc: To use this expanded user information, you will need to modify and subclass your current spawner to read these fields from `auth_state`...

oauthenticator/_version.py
```diff
@@ -7,7 +7,7 @@
     0,
     7,
     0,
-    'dev', # comment-out this line for a release
+    'dev',  # comment-out this line for a release
```
willingc: Very pythonic :smile:

athornton: That would be FlyCheck plus autopep8....

oauthenticator/github.py
```diff
@@ -109,20 +155,41 @@ def authenticate(self, handler, data=None):
...
-                if user_in_org:
-                    return username
-            else:  # User not found in member list for any organisation
-                return None
-        else:  # no organization whitelisting
-            return username
-
+                if not user_in_org:
+                    # User not found in member list for any organisation
+                    return None
```
minrk: This has broken the logic of group whitelist checking. If the user is not in the first organization in the whitelist, they will not be allowed to login anymore. Let's try to leave the whitelist-checking logic unchanged.

oauthenticator/github.py
```diff
@@ -141,10 +208,23 @@ def _check_organization_whitelist(self, org, username, access_token):
...
+        headers = _api_headers(access_token)
+        next_page = "https://%s/user/orgs" % (GITHUB_API)
+        orgmap = {}
+        while next_page:
+            req = HTTPRequest(next_page, method="GET", headers=headers)
+            resp = yield http_client.fetch(req)
+            resp_json = json.loads(resp.body.decode('utf8', 'replace'))
+            next_page = next_page_from_links(resp)
+            for entry in resp_json:
+                orgmap[entry["login"]] = entry["id"]
```
athornton: Because I'm using the ID to construct group ID mappings on the back end.  As with the GH ID, this keeps my group IDs the same, which then lets me create globally-consistent user groups from github org membership.

oauthenticator/github.py
athornton: Whoops.  That was my bad.  The if needs an outdent.

athornton: We need to change the logic a little, obviously, since we need to fall through to set up auth_state, so....

```
        if self.github_organization_whitelist:
            for org in self.github_organization_whitelist:
                user_in_org = yield self._check_organization_whitelist(org, use\
rname, access_token)
                if user_in_org:
                    break
            else:  # User not found in member list for any organisation
                return None
```

Not a huge fan of `for/else` myself but this is what it's for.

athornton:
I don't think github_scopes.py needs to exist.  `scopes` is defined in the base oauth2.py file as a simple list (not even a traitlet), so I think just documenting "here are how you get useful effects from GitHub scopes, if you want" is probably adequate?  A less-sophisticated user just won't set it, and then they will get authentication but not the additional already-set-up-to-push magic.

athornton:
It will probably be Monday before I submit a further-corrected PR; we're interviewing a job candidate today and I think I'm probably tied up the rest of the day.

athornton:
OK, new and improved version is up.

oauthenticator/github.py
athornton: 1) (unless you go with my original proposal of feature switches, but that does get slightly gross with respect to choosing them/adding environment variables) is definitely true.  (So, for that matter, does getting the GitHub email unless the user has a public primary email.)

I hadn't really thought about using the info at spawn time rather than in the authenticator class.  I guess that might work.  I don't know that I am likely to get this buttoned down before JupyterCon, though.  Today and tomorrow are full of job interviews, and then I need to stand up our JupyterLab tutorial environment for our All-Hands meeting next week...and then JupyterCon is upon us.

athornton:
I've incorporated @minrk 's suggestions (about moving my specific use-case stuff off into a subclass).  What I've got is still a WIP, but I think it's pretty close.  We should definitely make `scope` a configurable in the parent LoginHandler class though.

athornton:
Sorry about all the whitespace changes.  Autopep8+Flycheck.  I mean, on the bright side, pep8-compliant-ish!

minrk:
@athornton thanks! I've just pushed two small updates:

1. match GitHub API keys, so there isn't as much to keep track of (login -> login, id -> id, etc.)
2. update tests for the new data

It would be nice to avoid including auto-style-enforcement changes in the future since they do make code review more difficult, but it's not usually a big deal.

PR #108, opened 2017-08-19
Add VSTSOAuthenticator (Visual Studio Team Services)

puneetjindal:


oauthenticator/vsteamservices.py
```diff
@@ -0,0 +1,110 @@
...
+
+VSTS_HOST = os.environ.get('VSTS_HOST') or 'app.vssps.visualstudio.com'
+
+
+class VstsEnvMixin(OAuth2Mixin):
+    _OAUTH_AUTHORIZE_URL = "https://%s/oauth2/authorize" % VSTS_HOST
+    _OAUTH_ACCESS_TOKEN_URL = "https://%s/oauth2/token" % VSTS_HOST
+
+
+class VstsLoginHandler(OAuthLoginHandler, VstsEnvMixin):
```
minrk: Minor style nit: Would you mind using `VSTS` instead of `Vsts` for the acronym part?

oauthenticator/vsteamservices.py
```diff
@@ -0,0 +1,110 @@
...
+    scope = ["vso.profile"]
+
+    def get(self):
+        redirect_uri = self.authenticator.get_callback_url(self)
+        self.log.info('VSTS redirect: %r', redirect_uri)
+        state = self.get_state()
+        self.set_state_cookie(state)
+        self.authorize_redirect(
```
minrk: Can you add a comment about why the default `get` method doesn't work here?

consideRatio:
@puneetjindal thank you for making this PR!

For this authenticator specifically, I'm not sure if it is possible to use the generic oauthenticator class, but if it is either an typical OpenID Connect or OAuth2 authentication, it should work fine. I'll close this PR at this point.

Thank you for contributing to the Jupyter ecosystem @puneetjindal! :heart:

PR #109, opened 2017-08-24
fix gitlab request headers

lxneng:
Authentication token as a header should be `"Authorization": "Bearer {access_token}"`


willingc:
Thanks for the correction @lxneng. :cake:

[ref](https://docs.gitlab.com/ce/api/oauth2.html#access-gitlab-api-with-access-token)

cc/@minrk

PR #116, opened 2017-08-30
Make is_admin=False by default for GitLab

### LINKED ISSUE #115: GitLab auth fails for non admin users
Recent changes in GitLab's API have removed the "is_admin" property from most API calls: https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/10846

Now it is only shown when the request comes from an admin user:
https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/12211

This causes an error 500 for non-admin users, when "is_admin" is not found in the response.

balkian:
Newer versions of the GitLab API do not add the "is_admin" field unless the call
made by an admin user:

https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/12211

This change adds a default when this field is missing.
It closes #115

### DIFF
--- a/oauthenticator/gitlab.py
+++ b/oauthenticator/gitlab.py
@@ -107,7 +107,7 @@ class GitLabOAuthenticator(OAuthenticator):
 
         username = resp_json["username"]
         user_id = resp_json["id"]
-        is_admin = resp_json["is_admin"]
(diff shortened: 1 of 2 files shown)


willingc:
Thanks @balkian. :cookie:

balkian:
That was quick! Thank you all for your great work.

PR #117, opened 2017-08-31
store the whole GitHub user model in auth_state

minrk:
good unopinionated default, which we can follow easily in all OAuthenticators

cc @athornton this affects where the fields you are using will be found.
Instead of auth_state.email, it will be auth_state.github_user.email, etc.


willingc:
Thanks 👍

PR #118, opened 2017-08-31
Make OAuthenticator.scope configurable

### LINKED ISSUE #111: Make oauth scopes configurable
OAuth scopes should be a configurable trait on OAuthenticator classes, so that deployments can request specific permissions when passing auth state to spawners.

It should start by adding the scopes as a configurable:

```python
class OAuthenticator():
    ....
    scopes = List(Unicode(), config=True)
```

Then the scopes will need to be passed to the LoginHandler.

minrk:
this is becoming needed now that we are preserving tokens via auth_state

closes #111

### DIFF
--- a/oauthenticator/cilogon.py
+++ b/oauthenticator/cilogon.py
@@ -51,28 +51,15 @@ class CILogonMixin(OAuth2Mixin):
 
 
 class CILogonLoginHandler(OAuthLoginHandler, CILogonMixin):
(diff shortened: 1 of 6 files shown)


stefansedich:
YES! I just went through some pain trying to get this to work assuming that setting GitHubOAuthenticator.scope already worked as it wasn't clear to me in the documentation that it did not do anything.

I ended up subclassing GitHubLoginHandler for now and setting the scope I need to get it to work.

PR #121, opened 2017-09-21
Replace cilogon with OAuth2 implementation

athornton:
Update CILogon from OAuth1.0a to OAuth2.


minrk:
Hooray! Thanks, @athornton!

davidedelvento:
This is great. Any ETA for when it's released and posted to pypi so I can simply install it with pip and not worry about checking out the trunk of the repo?

minrk:
I want to get basic auth_state and configurable scopes implemented across all providers, then I think it will be time for another release.

PR #127, opened 2017-10-02
changelog for 0.7

minrk:
Mainly:

- includes auth_state for all authenticators
- `.scope` is configurable
- CILogon uses OAuth 2
- other fixes

cc @willingc

Can release with this and #125, #126


minrk:
Released 0.7.0

PR #129, opened 2017-10-04
typo in name field

### LINKED ISSUE #128: Google auth on hosted domain breaking in 0.7.0 if no name in claim
Hi,

`0.7.0` broke Google authentication for me. Works fine in `0.6.1`. The issue appears to be that my Google account (which is a G Suite account) does not return the `name` claim in the ID Token without additional scope. So when the `auth_state` makes its way to `jupyterhub/auth.py`, it chokes on `get_authenticated_user`, which checks to see if `name` is present (line 226). 

One possible workaround is to ask for the `profile` scope, which *should* include `name`. Trouble with that is, [Google says](https://developers.google.com/identity/protocols/OpenIDConnect#obtainuserinfo) it doesn't _have to_ include `name`. 

I'm not entirely sure what the best way around this is--I guess that depends on why `jupyterhub/auth.py` needs to have `name`. I feel like it should be using `username` (derived from `email`, which I think should always be present in a hosted_domain).

Here's a (slightly obscured) traceback:
```Traceback (most recent call last):

      File "/opt/conda/lib/python3.5/site-packages/tornado/web.py", line 1511, in _execute

        result = yield result

      File "/opt/conda/lib/python3.5/site-packages/oauthenticator/oauth2.py", line 182, in get

        user = yield self.login_user()

      File "/opt/conda/lib/python3.5/site-packages/jupyterhub/handlers/base.py", line 327, in login_user

        authenticated = yield self.authenticate(data)

      File "/opt/conda/lib/python3.5/site-packages/jupyterhub/auth.py", line 226, in get_authenticated_user

        raise ValueError("user missing a name: %r" % authenticated)

    ValueError: user missing a name: {'username': 'charles.forelle', 'auth_state': {'google_user': {'given_name': '', 'name': '', 'email': 'charles.forelle@XXXXXXXXX', 'family_name': '', 'id': '105xxxxxxxxxxxxx', 'hd': 'XXXXXXX', 'picture': 'https://lh3.googleusercontent.com/XXXXX/photo.jpg', 'verified_email': True}, 'access_token': 'XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX'}}

minrk:
it should have been 'name' not 'username'

github and mediawiki were doing it right already

closes #128

will release 0.7.1 immediately

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,13 @@ command line for details.
 
 ## [Unreleased]
 
-## [0.7] - 2017-10-02
+## 0.7
+
+### [0.7.1] - 2017-10-04
+
+- fix regression in 0.7.0 preventing authentication via providers other than GitHub, MediaWiki
+
+### [0.7.0] - 2017-10-02
 
 0.7.0 adds significant new functionality to all authenticators.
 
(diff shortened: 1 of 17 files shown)


cforelle:
Thanks! `0.7.1` is working for me.

PR #131, opened 2017-10-06
Globus OAuth: Bug fix, code cleanup, and added security features

NickolausDS:
The bug fix included in these changes adds the `state` parameter to the OAuth flow. Now that the base authenticator checks for `state`, the authenticator also now properly sets it. 

Since some new changes to `auth_state` for storing confidential user information, old code that stored this information manually has been removed.

New security features include setting a custom logout URL for a specific identity provider, and automatic token revocation whenever a user does a logout. Both of these options are configurable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,7 +212,7 @@ and set your `OAUTH_` environment variables.
 Visit https://developers.globus.org/ to set up your app. Ensure _Native App_ is
 unchecked and make sure the callback URL looks like:
 
-    https://[your-host]/hub/oauth_callback
+    https://[your-host]/oauth_callback
 
 Set scopes for authorization and transfer. The defaults include:
 
@@ -224,13 +224,22 @@ Set the above settings in your `jupyterhub_config`:
 # Tell JupyterHub to create system accounts
 from oauthenticator.globus import LocalGlobusOAuthenticator
 c.JupyterHub.authenticator_class = LocalGlobusOAuthenticator
+c.LocalGlobusOAuthenticator.enable_auth_state = True
+c.LocalGlobusOAuthenticator.oauth_callback_url = 'https://[your-host]/oauth_callback'
+c.LocalGlobusOAuthenticator.client_id = '[your app client id]'
+c.LocalGlobusOAuthenticator.client_secret = '[your app client secret]'
 ```
 
-and set the `OAUTH_` environment variables or config values.
+Alternatively you can set env variables for the following: `OAUTH_CALLBACK_URL`, `OAUTH_CLIENT_ID`,
+and `OAUTH_CLIENT_SECRET`. Setting `JUPYTERHUB_CRYPT_KEY` is required, and can be generated
+with OpenSSL: `openssl rand -hex 32`
+
+You are all set by this point! Be sure to check below for tweaking settings
+related to User Identity, Transfer, and additional security.
 
 ### User Identity
 
-By default, all users are restricted to their *Globus IDs* (malcolm@globusid.org)
+By default, all users are restricted to their *Globus IDs* (example@globusid.org)
 with the default Jupyterhub config:
 
 ```python
@@ -265,6 +274,11 @@ c.LocalGlobusOAuthenticator.exclude = ['auth.globus.org']
 # If the JupyterHub server is an endpoint, for convenience the endpoint id can be
 # set here. It will show up in the notebook kernel for all users as 'GLOBUS_LOCAL_ENDPOINT'.
 c.LocalGlobusOAuthenticator.globus_local_endpoint = '<Your Local JupyterHub UUID>'
+# Set a custom logout URL for your identity provider
+c.LocalGlobusOAuthenticator.logout_redirect_url = 'https://auth.globus.org/v2/web/logout'
+# For added security, revoke all service tokens when users logout. (Note: users must start
+# a new server to get fresh tokens, logging out does not shut it down by default)
+c.LocalGlobusOAuthenticator.revoke_tokens_on_logout = False
 ```
 
 If you only want to authenticate users with their Globus IDs but don't want to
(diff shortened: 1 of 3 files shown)


oauthenticator/globus.py
```diff
@@ -72,22 +43,58 @@ class GlobusLoginHandler(OAuthLoginHandler, GlobusMixin):
...
         redirect_uri = self.authenticator.get_callback_url(self)
         client = self.authenticator.globus_portal_client()
+        state = self.get_state()
+        self.set_state_cookie(state)
         client.oauth2_start_flow(
             redirect_uri,
             requested_scopes=' '.join(self.authenticator.scope),
-            refresh_tokens=self.authenticator.allow_refresh_tokens
+            refresh_tokens=self.authenticator.allow_refresh_tokens,
+            state=state
```
NickolausDS: `state` wasn't previously set in the OAuth flow, which began causing errors since the base authenticator started checking for it. Now that it's properly set, the OAuth flow should proceed normally.

oauthenticator/globus.py
```diff
@@ -193,10 +204,27 @@ def authenticate(self, handler, data=None):
         return {
             'name': username,
             'auth_state': {
-                'globus_data': globus_data,
+                'client_id': self.client_id,
+                'tokens': {
+                    tok: v for tok, v in tokens.by_resource_server.items()
+                    if tok not in self.exclude_tokens
+                },
```
NickolausDS: Moving the `auth_state` data down into the return dict was mostly a stylistic choice. Its only purpose is being returned in `auth_state` so removing it from the identity provider lines makes the code slightly more readable.

oauthenticator/globus.py
NickolausDS: Yes. `oauth2_start_flow` does provide the base auth URL which would automatically update if the Globus Auth URL were ever to change, but that is unlikely any time soon and updating the URL would be trivial. I think the reduction in code is very much worth it. I'll post an update shortly.

oauthenticator/globus.py
```diff
@@ -72,22 +43,58 @@ class GlobusLoginHandler(OAuthLoginHandler, GlobusMixin):
...
+    def get(self):
+        user = self.get_current_user()
+        if user:
+            if self.authenticator.revoke_tokens_on_logout:
+                self.clear_tokens(user)
+            self.clear_login_cookie()
+        if self.authenticator.logout_redirect_url:
+            self.redirect(self.authenticator.logout_redirect_url)
+        else:
+            super().get()
```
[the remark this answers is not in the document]
NickolausDS: The problem is the `self.redirect()` in the base [Jupyterhub Logout Handler](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/handlers/login.py#L25). We can't do a redirect after a redirect, so this code explicitly checks to make sure that doesn't happen. An alternative could be adding an issue to Jupyterhub to make the logout URL configurable:

    class LogoutHandler(BaseHandler):
        self.redirect_url = self.settings['login_url']
        def get(self):
            ...
            self.redirect(self.redirect_url, permanent=False)

Also, since the addition of `auth_state` to the authenticator, I needed to add some duplicate code to grab the user again to get at `auth_state` to revoke tokens. It's possible other services such as MediaWiki may want to do something to `auth_state` on logout, where the following addition would be convenient: 

    class LogoutHandler(BaseHandler):
        def get(self):
            user = self.get_current_user()
            if user:
                post_user_logout(user)
            ...
        def post_user_logout(self, user):
            """Override this to revoke user tokens in auth_state"""
            pass

I can add those as separate Jupyterhub issues if you think they merit the discussion. Unfortunately for the current code above, I don't think `super().get()` can be moved due to the second redirect.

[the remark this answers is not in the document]
minrk: > I can add those as separate Jupyterhub issues if you think they merit the discussion. 

Please do, I think both make sense as upstream issues to be handled.

oauthenticator/globus.py
```diff
@@ -210,6 +215,16 @@ def get_callback_url(self, handler=None):
                             )
         return self.oauth_callback_url
 
+    def logout_url(self, base_url):
+        return url_path_join(base_url, 'logout')
+
+    def get_handlers(self, app):
+        return [
```
minrk: Perhaps

```python
return super().get_handlers(app) + [(r'/logout', self.logout_handler)]
```

to only add handlers, rather than duplicate

NickolausDS: Will do, thanks!

oauthenticator/globus.py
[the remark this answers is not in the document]
NickolausDS: Great! I'll add them.

NickolausDS:
@minrk Were there any other suggested changes you would like to add?

README.md
```diff
@@ -212,7 +212,7 @@ and set your `OAUTH_` environment variables.
 Visit https://developers.globus.org/ to set up your app. Ensure _Native App_ is
 unchecked and make sure the callback URL looks like:
 
-    https://[your-host]/hub/oauth_callback
+    https://[your-host]/oauth_callback
```
minrk: This should stay as `/hub/oauth_callback`

README.md
```diff
@@ -224,13 +224,22 @@ Set the above settings in your `jupyterhub_config`:
 # Tell JupyterHub to create system accounts
 from oauthenticator.globus import LocalGlobusOAuthenticator
 c.JupyterHub.authenticator_class = LocalGlobusOAuthenticator
+c.LocalGlobusOAuthenticator.enable_auth_state = True
+c.LocalGlobusOAuthenticator.oauth_callback_url = 'https://[your-host]/oauth_callback'
```
minrk: same here, add `/hub/`

NickolausDS:
@minrk Readme fix has been added

minrk:
Thanks!

NickolausDS:
@minrk Thanks for the detailed review!

Issue #133, opened 2017-10-09
Problems with CILogon parameters


davidedelvento:
This is partially a follow up on issue #114 (only partially, so I'm opening a new issue instead of continuing there), and involves usernames (as #114) but also skins and identity providers.

Let's start from usernames. As the comment says in this code snippet https://github.com/jupyterhub/oauthenticator/blob/2b462c547ed8832f7e832f094134762e7cab6cc0/oauthenticator/cilogon.py#L149-L152 this is not ideal. At the moment this returns something like `http://cilogon.org/servera/users/NNNN` which is far from the actual username (`NNNN` is probably a UID on some machine, but not the text username). This could be probably fixed by using  `eppn` instead of `sub` in the above snipped. However, other people might want to use something different, e.g. email, so why not leaving that end-user-configurable and getting its value from an environment variable such as `CILOGON_CLAIM`? When not set, let's use `sub` (and when not found, before `return`ing `None`, let's spit a message in the logs about what was looked for and not found).

---

Somewhat related to this, I'd like to configure the skin and identity provider which seem to be possible (or at least thought as possible) looking at https://github.com/jupyterhub/oauthenticator/blob/2b462c547ed8832f7e832f094134762e7cab6cc0/oauthenticator/cilogon.py#L58-L62

However (and this could be my ignorance) I can't figure out how to set `self.authenticator.skin` and `idp`. If that part is still missing, how again using environmental variables `CILOGON_SKIN` and `CILOGON_IDP` while preserving current behavior when not set? From a user's perspective (who is not actually calling cilogon.py, but starting the jupyterhub server with various configurations) the environmental variables seem a better solution, and people are already setting `CILOGON_CLIENT_ID` and `CILOGON_CLIENT_SECRET` anyway, so this is not something different from what they are already doing.

---
If this sounds like a good approach, I'll be happy to implement these changes and make a PR myself if necessary.

Thanks @eshook @jbasney

minrk:
@davidedelvento to set these variables, in your `jupyterhub_config.py` add:

```python
c.CILogonOAuthenticator.skin = 'myskin'
c.CILogonOAuthenticator.idp = 'myidp'
```

In general, that's how all configuration of JupyterHub is done. Environment variables are available for certain auth-related configuration, such as secrets that one wants to avoid putting in files. But for the most part, configuration is done via the configuration file.

[...]

PR #134, opened 2017-10-11
Making possible to specify custom claims for username in jupytherhub_config.py

davidedelvento:
see #133

The provided tests pass, however I did not add a specific test for this per @minrk message regarding having some tests already (probably not pushed yet).

Hope I did "the right thing" since I was not familiar with the way configuration settings "pop up" from the `jupytherhub_config.py` to the classes which actually need them, so I made a few guesses.

### DIFF
--- a/oauthenticator/cilogon.py
+++ b/oauthenticator/cilogon.py
@@ -21,7 +21,7 @@ import json
 import os
 
 from tornado.auth import OAuth2Mixin
-from tornado import gen
+from tornado import gen, web
 
 from tornado.httputil import url_concat
 from tornado.httpclient import HTTPRequest, AsyncHTTPClient
@@ -59,8 +59,6 @@ class CILogonLoginHandler(OAuthLoginHandler, CILogonMixin):
             extra_params["selected_idp"] = self.authenticator.idp
         if self.authenticator.skin:
             extra_params["skin"] = self.authenticator.skin
-        if self.authenticator.username_key:
-            extra_params["claim"] = self.authenticator.username_key
 
         return super().authorize_redirect(*args, **kwargs)
 
@@ -72,10 +70,10 @@ class CILogonOAuthenticator(OAuthenticator):
     client_secret_env = 'CILOGON_CLIENT_SECRET'
     login_handler = CILogonLoginHandler
 
-    scope = List(Unicode(), default_value=['openid'],
+    scope = List(Unicode(), default_value=['openid', 'email', 'org.cilogon.userinfo'],
         config=True,
         help="""The OAuth scopes to request.
-        
+
         See cilogon_scope.md for details.
         At least 'openid' is required.
         """,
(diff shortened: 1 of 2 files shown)


oauthenticator/cilogon.py
```diff
@@ -101,6 +104,15 @@ def _validate_scope(self, proposal):
             Contact help@cilogon.org to request a custom skin.
         """,
     )
+    claim = Unicode(
+        config=True,
+        default_value = "sub", # Username is "sub" claim, or what specified in jupytherhub_config.py
```
minrk: Can we change the default to "email"? The sub claim produces an invalid username because it's a URL, which won't work.

oauthenticator/cilogon.py
```diff
@@ -59,6 +59,9 @@ def authorize_redirect(self, *args, **kwargs):
             extra_params["selected_idp"] = self.authenticator.idp
         if self.authenticator.skin:
             extra_params["skin"] = self.authenticator.skin
+        if self.authenticator.username_key:
```
minrk: username_key is undefined. Do you mean for the `claim` trait below to be called `username_key`? Or are `claim` and `username_key` different concepts? It seems that the key from which we will retrieve the username and the claim requested via the oauth flow aren't the same thing.

oauthenticator/cilogon.py
davidedelvento: I can certainly do that, however that's changing the current behavior in a backward incompatible way. Do you want me to do that?

oauthenticator/cilogon.py
davidedelvento: They are the same thing, in my mind. I was inclined to call it claim because that is what it is internally in cilogon, but I interpreted https://github.com/jupyterhub/oauthenticator/issues/133#issuecomment-335754031 as "let's call it `username_key`, so that users who do not know anything about cilogon will be less confused when needing to set that in `jupytherhub_config.py`". From what you are saying now, I think I probably misinterpreted that though

oauthenticator/cilogon.py
minrk: Well, the last version that worked used email, when we were using OAuth1. The OAuth2 update changed the key for the name to `sub` in the latest release. That has never worked due to the issues reported in #133, where it's producing an invalid username. So I don't think it's a breaking change, since we can't break something that doesn't work yet.

oauthenticator/cilogon.py
minrk: Thanks, that makes the `claim` word make sense. Since I want it to be clear what this means - that it's specifically the claim to produce the username, let's call it `username_claim`.

minrk:
Thanks! I've updated this PR to use:

- `username_claim` as the field
- default to `eppn`, which seems to be the suggested default
- add a caveat that `sub` cannot be used as the username field
- add a caveat that `email` may be used in cases where not all users will have `eppn` (e.g. login via Google)

davidedelvento:
This looks good to me, thanks. I see some builds have failed on travis, but they seem unrelated to the changes (more about the `requirements.txt` installation, AFAICT)

minrk:
Fixed it in the test!

zonca:
thanks for the fix,

I suggest to explain what `ePPN` is, possibly in the readme.

PR #137, opened 2017-10-17
parametrize Gitlab API version

### LINKED ISSUE #119: hard-coded gitlab api version 'v3' will need to handle switch to 'v4'
Just flagging that 'v3' of the gitlab api is hard-coded [here](
https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/gitlab.py#L29).

From [here](https://docs.gitlab.com/ce/api/v3_to_v4.html), 'v3' was deprecated earlier this year and support for it was removed since version 9.5 of Gitlab release in August 2017.  Apparently, only 'v4' will be handled from version 9.5 onwards.  Having said that, calls against the v3 api still appear to work for gitlab.com which is running version 9.5.4.

Users that are using a self-hosted gitlab instances may have both old and newer versions of gitlab, so perhaps parameterising the version using something like `GITLAB_API_VERSION` may make sense, with the default set to `4` of `v4`.

jbweston:
API v3 is deprecated for Gitlab 9.5+, but we also need to leave it as an option for self-hosted Gitlab installs.

Closes #119.

### DIFF
--- a/oauthenticator/gitlab.py
+++ b/oauthenticator/gitlab.py
@@ -26,7 +26,8 @@ from .oauth2 import OAuthLoginHandler, OAuthenticator
 
 # Support gitlab.com and gitlab community edition installations
 GITLAB_HOST = os.environ.get('GITLAB_HOST') or 'https://gitlab.com'
-GITLAB_API = '%s/api/v3' % GITLAB_HOST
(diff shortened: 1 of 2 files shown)


minrk:
Thanks!

PR #143, opened 2017-10-25
Allow configuration for Generic OAuthenticator login_service

fjbsantiago:
This is useful to specify the name of a custom OAuth provider.


minrk:
Thanks!

Issue #140, opened 2017-10-23
stripping the domain from cilogon usernames?


davidedelvento:
This is related to #133 @eshook @jbasney @zonca 

oauthenticator/google.py has a "hosted_domain" configuration option that strips off the domain from the username value. I think something similar would be nice to have for oauthenticator/cilogon.py when you're using a single IdP (e.g. in my case everybody has a `@xsede.org` suffix) to prevent messing up with usernames with `@` and `.` in them.

Alternatively, I could make a callback option to set up in `jupytherhub_config.py` where the user would specify which "massage" (if any) to give to the username strings.

As before, I'd be happy to provide a PR, if given some guidance on what would be best for the project and more likely to be accepted.

minrk:
That sounds like a good idea. A similar `idp_whitelist` and `strip_idp_domain` pair of configurables along the lines of:

```python
name, idp = email.split('@')
if self.idp_whitelist and idp not in self.idp_whitelist:
    # reject login
if len(self.idp_whitelist) == 1 and self.strip_idp_domain:
    username = name
else:
    username = email
```

ought to work.

davidedelvento:
see PR #145 I tested in my implementation and it works for me

davidedelvento:
see PR #145 I tested in my implementation and it works for me

davidedelvento:
@minrk I see this is merged and works fine for me. 
To properly plan my work, do you have any sense in when it might be included in an official release?
Thanks!

willingc:
Thanks @davidedelvento for letting us know it works. I've opened https://github.com/jupyterhub/oauthenticator/issues/170 to track the next official release.

PR #145, opened 2017-10-31
stripping the domain

davidedelvento:
As per discussion in issue #140

### DIFF
--- a/oauthenticator/cilogon.py
+++ b/oauthenticator/cilogon.py
@@ -164,12 +164,13 @@ class CILogonOAuthenticator(OAuthenticator):
             )
             raise web.HTTPError(500, "Failed to get username from CILogon")
 
-        gotten_name, gotten_idp = username.split('@')
-        if self.idp_whitelist and gotten_idp not in self.idp_whitelist:
-            self.log.error("Trying to login from not whitelisted domain %s", gotten_idp)
-            raise web.HTTPError(500, "Trying to login from not whitelisted domain")
-        if len(self.idp_whitelist) == 1 and self.strip_idp_domain:
-            username = gotten_name
+        if self.idp_whitelist:
+            gotten_name, gotten_idp = username.split('@')
+            if gotten_idp not in self.idp_whitelist:
+                self.log.error("Trying to login from not whitelisted domain %s", gotten_idp)
+                raise web.HTTPError(500, "Trying to login from not whitelisted domain")
+            if len(self.idp_whitelist) == 1 and self.strip_idp_domain:
+                username = gotten_name
         userdict = {"name": username}
         # Now we set up auth_state
         userdict["auth_state"] = auth_state = {}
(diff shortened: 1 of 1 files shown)


oauthenticator/cilogon.py
```diff
@@ -153,6 +163,13 @@ def authenticate(self, handler, data=None):
                 self.username_claim, sorted(resp_json.keys())
             )
             raise web.HTTPError(500, "Failed to get username from CILogon")
+
+        gotten_name, gotten_idp = username.split('@')
```
minrk: Can we add one more conditional so that this whole block is inside `if self.idp_whitelist:`? I think there might be some cases (e.g. where `ePPN` is not used) that username doesn't have `@` in it, in which case this initial split will fail.

davidedelvento: Sure, that makes sense. However it might still fail if `if self.idp_whitelist`, no? 

So maybe it's better to put it in a `try` block, which on exception will set `gotten_idp` to either empty string or `None`. In this way we don't have strange exception if the split fails (it will still failin the `HTTPError` below).

What do you think>

davidedelvento: @minrk Just pinging you to make sure this does not fall in the oblivion. I'm waiting for your feedback on it. Thanks!

minrk: Failing is okay, I think. If using an idp whitelist, it should be incumbent on the admin to choose a username field that will work, right? Or are there username fields where only some users will get names with `@` and not others?

We could also do:

```python
if '@' not in username:
    name = username
    idp = ''
else:
    name, idp = username.split('@', 1)
```

davidedelvento: Well, I don't know, since that is not something that I am in control of: the auth provider is. Maybe @jbasney has any insight? If @minrk think that is ok with failing, I'm ok with it too.

jbasney: The eppn and email claims should always have the user@domain form but the sub claim won't, and strictly speaking the sub claim is what clients should be using, so I agree it's good to put the split inside `if self.idp_whitelist`.

The idea of having multiple entries in `self.idp_whitelist` makes me nervous, because of the possibility of username collisions across multiple domains. You don't want jbasney@wisc.edu logging in to jbasney@illinois.edu's account. It seems safer to follow the example from google.py to allow only a single `self.hosted_domain` parameter when you have a single trusted domain.

minrk: > The idea of having multiple entries in self.idp_whitelist

That's why stripping the domain is only allowed when `len(idp_whitelist) == 1`

davidedelvento:
Ok, so I've made the split inside a `if self.idp_whitelist` block. 

I've also made the stylistic choice or having the whole logic inside the same block. It does not change the functionality at all, but I think it is much easier to understand for a human brain, and therefore less prone to introducing subtle bugs.

@minrk  @jbasney  please let me know if you deem if good enough to be merged or if you have additional suggestions!

minrk:
Thanks!

PR #151, opened 2017-12-07
Remove problematic defaults from generic oauthenticator

### LINKED ISSUE #147: GenericOAuthenticator has bad token_url default parameter
Hello. These lines:

https://github.com/jupyterhub/oauthenticator/blob/bd4567ceacca1dcc8122be2f96db320c4342fd02/oauthenticator/generic.py#L60-L64

`GET` is not a sensible default value for `token_url` (and the help string is a remnant of a copy-paste). Not setting `token_url` causes an uncaught exception in tornado when exchanging tokens.

I'm not sure what a sensible default value would be, though. Maybe just the empty string?

jpolchlo:
Fixes #147 

While working with the Generic OAuthenticator, I ran into mysterious errors that could have been avoided with some more useful error messages.  I've attempted to remedy that in this PR.

Signed-off-by: jpolchlo <jpolchlopek@azavea.com>

### DIFF
--- a/oauthenticator/generic.py
+++ b/oauthenticator/generic.py
@@ -81,7 +81,7 @@ class GenericOAuthenticator(OAuthenticator):
         if 'OAUTH2_TOKEN_URL' in os.environ:
             url = self.token_url
         else:
-            raise ValueError("Please set the OAUTH2_TOKEN_URL environment variable ({})".format(token_url.help))
+            raise ValueError("Please set the OAUTH2_TOKEN_URL environment variable ({})".format(self.token_url.help))
 
         b64key = base64.b64encode(
             bytes(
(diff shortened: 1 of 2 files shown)


jpolchlo:
Apologies for spamming this PR with excessive pushes, but I had trouble getting the tests to run locally.  The remaining Travis error is triggered during installation of `six`, so this should be ready for review.

oauthenticator/generic.py
```diff
@@ -78,7 +78,10 @@ def authenticate(self, handler, data=None):
             grant_type='authorization_code'
         )
 
-        url = self.token_url
+        if 'OAUTH2_TOKEN_URL' in os.environ:
```
minrk: Better to directly check `if not self.token_url`. Same below.

PR #152, opened 2018-01-05
Fix token_url default value and help

dbsanfte:
These look wrong. A default value of GET makes no sense for an URL. Looks like it was just copy pasted from USERDATA_METHOD above.

Fixing with sane values.


minrk:
Thanks! Fixed by #151

PR #154, opened 2018-01-30
Replace access_token with token_response.

athornton:
It turns out that the ID token can be useful too, so we preserve the entire token response.

### DIFF
--- a/oauthenticator/cilogon.py
+++ b/oauthenticator/cilogon.py
@@ -39,6 +39,7 @@ class CILogonMixin(OAuth2Mixin):
 
 class CILogonLoginHandler(OAuthLoginHandler, CILogonMixin):
     """See http://www.cilogon.org/oidc for general information."""
+
     def authorize_redirect(self, *args, **kwargs):
         """Add idp, skin to redirect params"""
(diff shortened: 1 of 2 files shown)


oauthenticator/cilogon.py
```diff
@@ -160,24 +162,26 @@ def authenticate(self, handler, data=None):
...
                 username = gotten_name
         userdict = {"name": username}
         # Now we set up auth_state
         userdict["auth_state"] = auth_state = {}
-        # Save the access token and full CILogon reply in auth state
+        # Save the token response and full CILogon reply in auth state
         # These can be used for user provisioning
         #  in the Lab/Notebook environment.
-        auth_state['access_token'] = access_token
+        auth_state['token_response'] = token_response
```
minrk: for backward-compatibility maybe leave access_token there?

Otherwise +1

athornton: Done.  Although I have a suspicion that I may be the only current user.

minrk: entirely possible, but still nice to do it anyway

PR #157, opened 2018-02-06
[MRG] Add refresh token to auth_state if it is available

betatim:
If the OAuth2 server sends a refresh token, add it to the `auth_state`.

### DIFF
--- a/oauthenticator/generic.py
+++ b/oauthenticator/generic.py
@@ -106,6 +106,7 @@ class GenericOAuthenticator(OAuthenticator):
         resp_json = json.loads(resp.body.decode('utf8', 'replace'))
 
         access_token = resp_json['access_token']
+        refresh_token = resp_json.get('refresh_token', None)
         token_type = resp_json['token_type']
 
         # Determine who the logged in user is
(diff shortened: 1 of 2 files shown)


oauthenticator/generic.py
```diff
@@ -134,6 +135,7 @@ def authenticate(self, handler, data=None):
             'name': resp_json.get(self.username_key),
             'auth_state': {
                 'access_token': access_token,
+                'refresh_token': refresh_token,
```
betatim: do you prefer the key to always be present (like this) but with `None` as value or only put the key in if the value is not `None`? I am easy either way.

betatim:
There is a whole collection of errors like this:
```python
_______________________ ERROR at teardown of test_auth0 ________________________
    def _close():
        io_loop.clear_current()
>       if (not tornado.ioloop.IOLoop.initialized() or
                io_loop is not tornado.ioloop.IOLoop.instance()):
E               AttributeError: type object 'IOLoop' has no attribute 'initialized'
../../../virtualenv/python3.6.3/lib/python3.6/site-packages/pytest_tornado/plugin.py:136: AttributeError
```

Any ideas what could be causing this? Recent update to tornado?

oauthenticator/generic.py
minrk: No preference from me

betatim:
🎉

PR #158, opened 2018-02-07
Add ioloop fixture that works with tornado 5

betatim:
This should fix travis failures like seen in #157


minrk:
This fixes most of it. I think there's another change in the Google OAuth urls that's sneaking past our mocking.

PR #169, opened 2018-02-24
Add OktaOAuthenticator (derived from the Auth0 provider)

emmanuel:
This is (lightly) adapted from the Auth0 provider. Main differences are the URLs, and that the token endpoint accepts its inputs (e.g., authorization code) as URL query parameters instead of in the POSTed request body.

### DIFF
--- a/oauthenticator/okta.py
+++ b/oauthenticator/okta.py
@@ -1,0 +1,122 @@
+"""
+Custom Authenticator to use Okta OAuth with JupyterHub
+
+Derived using the Auth0 OAuthenticator implementations as examples.
+
+The following environment variables may be used for configuration:
+
+    OKTA_DOMAIN - The domain for your Okta account; should end with okta.com or oktapreview.com
+    OAUTH_CLIENT_ID - Your client id
+    OAUTH_CLIENT_SECRET - Your client secret
+    OAUTH_CALLBACK_URL - Your callback handler URL
+
+Additionally, if you are concerned about your secrets being exposed by
+an env dump(I know I am!) you can set the client_secret, client_id and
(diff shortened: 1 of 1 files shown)


emmanuel:
In principle it would be possible to discover all of URLs from the base URL, using the discovery endpoint (i.e., `https://${BASE_URL}/.well-known/openid-configuration`). In practice the URL structure is stable enough that I didn't consider it worth the effort.

Also, in principle, the retrieved identity token could/should be verified locally (or remotely). In practice I didn't consider it worth the effort, especially given that the other providers don't currently verify the fetched user info.

sstarcher:
@consideRatio I'll test this out.  Anything else needed to get this merged outside of making the changes @minrk suggested?

consideRatio:
Since last review, we have put in more effort to conclude how we can make this repo more sustainable. To create a new class with new documentation for all providers isn't sustainable.

We need to go more general than we have done in the past with the code within this repository. I'm closing this PR to reflect this. Thank you all for your work invested!

sstarcher:
@consideRatio do you have any suggestions on what to do with something like this?  Should we fork the github repo or do you have an out of tree approach you would recommend/

manics:
Version 0.11.0 was recently released, with some changes that should make it easier to use and subclass. You should be able to create your own repo/package which imports some base-classes from oauthenticator, and extends them.

consideRatio:
@emmanuel thank you for making this PR!

For Okta specifically, it is possible to use the generic oauthenticator class. Due to this, I'll close this PR.

Thank you for contributing to the Jupyter ecosystem @emmanuel! :heart:

PR #172, opened 2018-03-06
Minor comment change.

tylere:
Changed "GitHub OAuth" to "OAuth", given that base class works for multiple OAuth providers.


minrk:
Thanks!

PR #173, opened 2018-03-09
add moodle OAuth2

henchc:
I've been able to get OAuth2 running for Moodle platforms and the [oauth plugin](https://github.com/projectestac/moodle-local_oauth) .

It only needed slight modifications to the generic oauth, so I'm not sure if you'd rather just generalize the generic a bit more or add this. Specifically, the `params` dict for the token and the access token for the `user_info` post.

I'm also not sure if this was all possible through only modifications to the `jupyterhub_config.py`, but I couldn't figure out how to add the above into that.

### DIFF
--- a/oauthenticator/moodle.py
+++ b/oauthenticator/moodle.py
@@ -1,0 +1,150 @@
+"""
+Custom Authenticator to use generic OAuth2 with JupyterHub
+"""
+
+
+import json
+import os
+import base64
+import urllib
+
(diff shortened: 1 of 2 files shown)


minrk:
If we could get this via modifications to the generic implementation, that would be great. Then it could be simplified to a configuration example rather than a new class.

henchc:
sounds good!

Issue #180, opened 2018-04-11
Google Oauthenticator excepts when trying to raise an exception


davidbath:
If the google `OAuthenticator` subclass doesn't get a response from the token URL it attempts to clean up the state and then 500 the user, here: https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/google.py#L73-L75

However, the `self.clear_all_cookies()` itself appears to except - I believe it might be intended to be `handler.clear_all_cookies()`.

minrk:
@davidbath I think that's right! Probably an artifact of refactoring things from Handler methods to Authenticator methods. Would you like to make a PR?

davidbath:
Sure - it'll be my first one, so might need some gentle correction through the process.

PR #181, opened 2018-04-12
Clear cookies on handler, not OAuthenticator

davidbath:
To close https://github.com/jupyterhub/oauthenticator/issues/180


minrk:
Thanks!

PR #182, opened 2018-04-12
Catch informative errors from upstream token endpoint

davidbath:
Would appreciate several eyes on this - there may be far superior ways to catch the errors that can come back from github (https://developer.github.com/apps/managing-oauth-apps/troubleshooting-oauth-app-access-token-request-errors/). As they are expected to be infrequent, I haven't attempted any auto-recovery - assume the error is either fairly permanent (client id/secret or callback URL mismatch) in which case some ops intervention will be required, or is highly transitory (code mismatch) in which case the user will just try again.

I haven't exercised the error conditions - would gladly accept some help on adding an error condition to the tests.

### DIFF
--- a/oauthenticator/github.py
+++ b/oauthenticator/github.py
@@ -16,7 +16,7 @@ from tornado.auth import OAuth2Mixin
 from tornado import gen, web
 
 from tornado.httputil import url_concat
-from tornado.httpclient import HTTPRequest, AsyncHTTPClient
+from tornado.httpclient import HTTPRequest, AsyncHTTPClient, HTTPError
 
 from jupyterhub.auth import LocalAuthenticator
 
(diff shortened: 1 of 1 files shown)


oauthenticator/github.py
```diff
@@ -116,7 +116,15 @@ def authenticate(self, handler, data=None):
         resp = yield http_client.fetch(req)
         resp_json = json.loads(resp.body.decode('utf8', 'replace'))
 
-        access_token = resp_json['access_token']
+        if 'access_token' in resp_json:
+            access_token = resp_json['access_token']
+        elif 'error_description' in resp_json:
+            raise HTTPError(403,
+                "An access token was not returned: {}".format(
```
minrk: Should this be a 403? I think typically this will be a result of a bug, so maybe leave it as a 500 with the error_description?

minrk:
I think this is a great idea. If you'd like to take a stab at looking at the request mocking and add a test with a mock error response from that page, that would be great, but if it proves infeasible, I'm happy with this as-is.

oauthenticator/github.py
davidbath: Well it’s  not necessarily a bug - if the auth code code out of sync (eg someone left a browser cache open for a long time) then a simple retry will fix it. Happy to make it a 500 if that’s more consitent with the jupyterhub style.

Issue #107, opened 2017-08-18
Working configuration for generic authenticator with Keycloak


pfisterer:
I have tried to configure Jupyterhub to use the generic OAuth2 authentication mechanism with [Keycloak](http://www.keycloak.org/) as OAuth2 sever. However, it never redirects to it for authentication. Could you please support me and provide a sample configuration that is supposed to work? Or is there anything that I'm obviously doing wrong?

Thank you very much!
Dennis

OAuthenticator configuration in jupyterhub_config.py:
```
from oauthenticator.generic import GenericOAuthenticator
c.JupyterHub.authenticator_class = GenericOAuthenticator

c.GenericOAuthenticator.oauth_callback_url = 'https://example.com:8443/auth/realms/testrealm/protocol/openid-connect/auth'
c.GenericOAuthenticator.client_id = 'oauth-secret'
c.GenericOAuthenticator.client_secret = 'asfdsdf-dfdf-fhhfh-aed7-asdafsdfsaf'
c.GenericOAuthenticator.token_url = 'https://example.com:8443/auth/realms/testrealm/protocol/openid-connect/token'
c.GenericOAuthenticator.userdata_url = 'https://141.72.18.234:8443/auth/realms/testrealm/protocol/openid-connect/userinfo'
```

Keycloak Endpoints:
```JSON
{
	"issuer": "https://example.com:8443/auth/realms/testrealm",
	"authorization_endpoint": "https://example.com:8443/auth/realms/testrealm/protocol/openid-connect/auth",
	"token_endpoint": "https://example.com:8443/auth/realms/testrealm/protocol/openid-connect/token",
	"token_introspection_endpoint": "https://example.com:8443/auth/realms/testrealm/protocol/openid-connect/token/introspect",
	"userinfo_endpoint": "https://example.com:8443/auth/realms/testrealm/protocol/openid-connect/userinfo",
	"end_session_endpoint": "https://example.com:8443/auth/realms/testrealm/protocol/openid-connect/logout",
	"jwks_uri": "https://example.com:8443/auth/realms/testrealm/protocol/openid-connect/certs",
	"check_session_iframe": "https://example.com:8443/auth/realms/testrealm/protocol/openid-connect/login-status-iframe.html",
	"grant_types_supported": ["authorization_code", "implicit", "refresh_token", "password", "client_credentials"],
	"response_types_supported": ["code", "none", "id_token", "token", "id_token token", "code id_token", "code token", "code id_token token"],
[...]

PR #183, opened 2018-04-23
Add KeycloakOAuthenticator

niveau0:
A possible implementation of a keycloak authenticator. Might be a solution for issue #107

### DIFF
--- a/oauthenticator/keycloak.py
+++ b/oauthenticator/keycloak.py
@@ -1,0 +1,110 @@
+"""
+Custom Authenticator to use Keycloak OAuth with JupyterHub
+
+Most of the code c/o Kyle Kelley (@rgbkrk)
+"""
+
+import json
+import os
+import urllib
+
+from tornado.auth import OAuth2Mixin
+from tornado import gen, web
+from tornado.httputil import url_concat
+from tornado.httpclient import HTTPRequest, AsyncHTTPClient
+
+from jupyterhub.auth import LocalAuthenticator
+from jupyterhub.handlers import LogoutHandler
+from jupyterhub.utils import url_path_join
+
+from .oauth2 import OAuthLoginHandler, OAuthenticator
+
+class KeycloakMixin(OAuth2Mixin):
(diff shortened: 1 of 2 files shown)


yuvipanda:
Thanks for the PR, @niveau0!
 
Does the generic oauthenticator not work by itself? According to https://github.com/jupyterhub/oauthenticator/pull/81 and @rafael-ladislau it does!

rafael-ladislau:
If you set these lines in your jupyterhub_config.py:

```
c.GenericOAuthenticator.login_service = 'You Service Name'
```

And set these environment variables before run JupyterHub:

```
OAUTH2_AUTHORIZE_URL=https://localhost:8080/auth/realms/master/protocol/openid-connect/auth
OAUTH2_TOKEN_URL=https://localhost:8080/auth/realms/master/protocol/openid-connect/token
OAUTH2_USERNAME_KEY=preferred_username
OAUTH_CLIENT_SECRET=<KEYCLOAK_CLIENT_SECRET>
OAUTH_CALLBACK_URL=https://<JUPYTERHUB_URL>/hub/oauth_callback
OAUTH2_USERDATA_URL=https://localhost:8080/auth/realms/master/protocol/openid-connect/userinfo
OAUTH_CLIENT_ID=<KEYCLOAK_CLIENT_ID>
```

The authentication should work with Keycloak.

niveau0:
Ah thanks, that would be great. I will try that. My implementation is an old one, so maybe something changed in the meantime? Whatever, I try it and close this if it works, thanks a lot.

minrk:
How did it go with the generic auth? Would you like to turn this into a docs PR for enabling keycloak via config?

willingc:
@niveau0 How's it going with keycloak and the generic authenticator?

niveau0:
@willingc Sorry for late reply, I will test it the next days, there was no time for now

consideRatio:
@niveau0  thank you for making this PR!

For Keycloak specifically, it seems to be possible to use the generic oauthenticator class. Due to this, I'll close this PR.

Thank you for contributing to the Jupyter ecosystem @niveau0! :heart:

PR #189, opened 2018-05-23
Bug fix for Globus when user logs in with an Identity Provider Namespace

NickolausDS:
Previously, the Globus Oauthenticator only expected usernames like `johndoe@uchicago.edu`, but it's also possible for Identity Providers to namespace their domains and log users in with usernames like `johndoe@uchicago.edu@provider.org`. 

More documentation: https://docs.globus.org/api/auth/specification/#identity_provider_namespaces

### DIFF
--- a/oauthenticator/globus.py
+++ b/oauthenticator/globus.py
@@ -166,7 +166,10 @@ class GlobusOAuthenticator(OAuthenticator):
         # Doing the code for token for id_token exchange
         tokens = client.oauth2_exchange_code_for_tokens(code)
         id_token = tokens.decode_id_token(client)
-        username, domain = id_token.get('preferred_username').split('@')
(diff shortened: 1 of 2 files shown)


oauthenticator/globus.py
```diff
@@ -166,7 +166,10 @@ def authenticate(self, handler, data=None):
         # Doing the code for token for id_token exchange
         tokens = client.oauth2_exchange_code_for_tokens(code)
         id_token = tokens.decode_id_token(client)
-        username, domain = id_token.get('preferred_username').split('@')
```
minrk: We can use `username, domain = id_token.split('@', 1)` to split only on the first occurrence of `@`.

NickolausDS: I like that much better, thanks!

PR #192, opened 2018-05-30
Deprecate GITLAB_HOST in favor of GITLAB_URL

### LINKED ISSUE #191: Rename GITLAB_HOST to GITLAB_URL
The name of env variable is misleading since it requires GitLab URL and not the host.

I would like to propose a following change with a deprecation warning.

```python
GITLAB_URL = os.getenv('GITLAB_URL')
GITLAB_HOST = os.getenv('GITLAB_HOST')

if not GITLAB_URL and GITLAB_HOST:
    if GITLAB_HOST.startswith('http'):
        warnings.warn(..., DeprecationWarning)
        GITLAB_URL = GITLAB_HOST
    else:
        GITLAB_URL = 'https://{0}'.format(GITLAB_HOST)
# elif GITLAB_URL and GITLAB_HOST:
#     if GITLAB_URL != GITLAB_HOST:
#         raise RuntimeError('Conflicting configuration of GITLAB_HOST and GITLAB_URL. Use only GITLAB_URL.')

# replace GITLAB_HOST with GITLAB_URL bellow ...
```

jirikuncar:
(closes #191)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ command line for details.
 
 
 ## [Unreleased]
+- Use GITLAB_URL instead of GITLAB_HOST to specify URL of your GitLab instance.
 
 ## 0.7
 
--- a/oauthenticator/gitlab.py
+++ b/oauthenticator/gitlab.py
@@ -9,6 +9,7 @@ based on the GitHub plugin by Kyle Kelley (@rgbkrk)
 import json
 import os
 import sys
+import warnings
 
 from tornado.auth import OAuth2Mixin
 from tornado import gen, web
@@ -23,10 +24,31 @@ from traitlets import Set
 
 from .oauth2 import OAuthLoginHandler, OAuthenticator
 
+GITLAB_URL = os.getenv('GITLAB_URL')
+GITLAB_HOST = os.getenv('GITLAB_HOST')
+
+if not GITLAB_URL and GITLAB_HOST:
+    warnings.warn('Use of GITLAB_HOST might be deprecated in the future. '
+                  'Rename GITLAB_HOST environemnt variable to GITLAB_URL.',
+                  PendingDeprecationWarning)
+    if GITLAB_HOST.startswith('https://') or GITLAB_HOST.startswith('http://'):
+        GITLAB_URL = GITLAB_HOST
+    else:
+        # Hides common mistake of users which set the GITLAB_HOST
+        # without a protocol specification.
+        GITLAB_URL = 'https://{0}'.format(GITLAB_HOST)
+        warnings.warn('The https:// prefix has been added to GITLAB_HOST.'
+                      'Set GITLAB_URL="{0}" instead.'.format(GITLAB_URL))
+
 # Support gitlab.com and gitlab community edition installations
-GITLAB_HOST = os.environ.get('GITLAB_HOST') or 'https://gitlab.com'
+if not GITLAB_URL:
+    GITLAB_URL = 'https://gitlab.com'
+
+# Use only GITLAB_URL in the code bellow.
+del GITLAB_HOST
+
 GITLAB_API_VERSION = os.environ.get('GITLAB_API_VERSION') or '4'
-GITLAB_API = '%s/api/v%s' % (GITLAB_HOST, GITLAB_API_VERSION)
+GITLAB_API = '%s/api/v%s' % (GITLAB_URL, GITLAB_API_VERSION)
 
 
 def _api_headers(access_token):
@@ -37,8 +59,8 @@ def _api_headers(access_token):
 
 
 class GitLabMixin(OAuth2Mixin):
-    _OAUTH_AUTHORIZE_URL = "%s/oauth/authorize" % GITLAB_HOST
-    _OAUTH_ACCESS_TOKEN_URL = "%s/oauth/access_token" % GITLAB_HOST
+    _OAUTH_AUTHORIZE_URL = "%s/oauth/authorize" % GITLAB_URL
+    _OAUTH_ACCESS_TOKEN_URL = "%s/oauth/access_token" % GITLAB_URL
 
 
 class GitLabLoginHandler(OAuthLoginHandler, GitLabMixin):
@@ -81,7 +103,7 @@ class GitLabOAuthenticator(OAuthenticator):
 
         validate_server_cert = self.validate_server_cert
 
-        url = url_concat("%s/oauth/token" % GITLAB_HOST,
+        url = url_concat("%s/oauth/token" % GITLAB_URL,
                          params)
 
         req = HTTPRequest(url,


betatim:
@minrk do you have any opinion/guidance on how to handle the deprecation?

oauthenticator/gitlab.py
```diff
@@ -23,10 +24,26 @@
...
 from .oauth2 import OAuthLoginHandler, OAuthenticator
 
+GITLAB_URL = os.getenv('GITLAB_URL')
+GITLAB_HOST = os.getenv('GITLAB_HOST')
+
+if not GITLAB_URL and GITLAB_HOST:
+    if GITLAB_HOST.startswith('https://') or \
+            GITLAB_HOST.startswith('http://'):
+        warnings.warn('Use GITLAB_URL instead GITLAB_HOST.',
+                      DeprecationWarning)
```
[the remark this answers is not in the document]
minrk: Putting the "deprecated in v0.8" is good. I don't like "will be removed", as removal of deprecated features should only occur when the presence of the deprecated feature causes a problem, which for something like this is likely ~never.

oauthenticator/gitlab.py
```diff
@@ -23,10 +24,26 @@
...
+        GITLAB_URL = GITLAB_HOST
+    else:
+        GITLAB_URL = 'https://{0}'.format(GITLAB_HOST)
```
minrk: This is adding a new feature to GITLAB_HOST while deprecating it, which we probably don't need to do. We can just have:

```python
if GITLAB_HOST and not GITLAB_URL:
    warnings.warn(...)
    GITLAB_URL = GITLAB_HOST
```

without looking at the contents of GITLAB_HOST.

jirikuncar: I will add the warning, but will leave the `https://` prefix. I've burned myself few times by setting `GITLAB_HOST` as host without `https://` prefix, so while the `GITLAB_HOST` is not completely removed I would keep it there for people that don't read docs very carefully ;)

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ command line for details.
 
 
 ## [Unreleased]
+- Use GITLAB_URL instead of GITLAB_HOST to specify URL of your GitLab instace.
```
minrk: s/instace/instance/

oauthenticator/gitlab.py
minrk: In that case, this is adding a *new feature* to fully supported GITLAB_HOST, instead of deprecating it at all. If supporting `GITLAB_HOST='mygitlab.org'` is a desired goal, I would propose this change:

- Don't define `GITLAB_URL`
- just add this `if '://' not in GITLAB_HOST: GITLAB_HOST = "https://GITLAB_HOST"`

I don't think it's useful to have separate environment variables for the same value with/without `https`. If that should be optional, it should be optional, but there is no need for different environment variables where it's optional in one, but not the other.

CHANGELOG.md
jirikuncar: Thanks!

oauthenticator/gitlab.py
```diff
@@ -23,10 +24,30 @@
...
+    warnings.warn('Use of GITLAB_HOST has been deprecated in v0.8. '
+                  'Rename GITLAB_HOST environemnt variable to GITLAB_URL.',
+                  DeprecationWarning)
+    if GITLAB_HOST.startswith('https://') or \
+            GITLAB_HOST.startswith('http://'):
+        GITLAB_URL = GITLAB_HOST
+    else:
+        # Hides common mistage of users which set the GITLAB_HOST
+        # without a protocol specification.
+        GITLAB_URL = 'https://{0}'.format(GITLAB_HOST)
```
jirikuncar: The warning is always displayed. This only helps new developers that don't read docs carefully and they will get the warning anyway.

oauthenticator/gitlab.py
```diff
@@ -23,10 +24,32 @@
...
+        warnings.warn('The https:// prefix has been added to GITLAB_HOST.'
+                      'Set GITLAB_URL="{0}" instead.'.format(GITLAB_URL))
+
 # Support gitlab.com and gitlab community edition installations
-GITLAB_HOST = os.environ.get('GITLAB_HOST') or 'https://gitlab.com'
+if not GITLAB_URL:
+    GITLAB_URL = 'https://gitlab.com'
+
+# Use only GITLAB_URL in the code bellow.
+del GITLAB_HOST
```
betatim: What do others think of this pattern? I've not seen it before. I can see the motivation for it but it still feels weird :)

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ command line for details.
 
 
 ## [Unreleased]
+- Use GITLAB_URL instead of GITLAB_HOST to specify URL of your GitLab instance.
```
willingc: I think it's ok to add GITLAB_URL but let's leave GITLAB_HOST as it currently is. According to the [GITLAB docs](https://docs.gitlab.com/ee/administration/environment_variables.html#doc-nav), GITLAB_HOST is the recommended environment variable and should include "The full URL of the GitLab server (including http:// or https://)"

oauthenticator/gitlab.py
willingc: This line will generate a duplication of https:// if GITLAB_HOST is set correctly.

oauthenticator/gitlab.py
jirikuncar: I wanted to make it explicit so the global variable is not used in the code bellow.

CHANGELOG.md
jirikuncar: I might submit a PR also to GitLab too. It is really confusing to call a variable `HOST` if it **must** contain an URL.

jirikuncar:
> I understand the desire to add GITLAB_URL, but it would be helpful to only fall back to it if GITLAB_HOST is not set.

@willingc That's exactly the current behaviour 😉

CHANGELOG.md
willingc: Let's do this to minimize user confusion. State in the release notes that the full URL should be set in both GITLAB_URL and GITLAB_HOST (current GitLab documentation).

We should not redefine what GITLAB_HOST means in this PR.

oauthenticator/gitlab.py
```diff
@@ -23,10 +24,32 @@
...
 from .oauth2 import OAuthLoginHandler, OAuthenticator
 
+GITLAB_URL = os.getenv('GITLAB_URL')
+GITLAB_HOST = os.getenv('GITLAB_HOST')
+
+if not GITLAB_URL and GITLAB_HOST:
+    warnings.warn('Use of GITLAB_HOST has been deprecated in v0.8. '
+                  'Rename GITLAB_HOST environemnt variable to GITLAB_URL.',
+                  DeprecationWarning)
+    if GITLAB_HOST.startswith('https://') or \
```
willingc: Let's make this one line and avoid the \. It's fine if it is over 80 characters.

oauthenticator/gitlab.py
willingc: I do not think that we should perform a destructive action on the user's system, especially without any warning.

jirikuncar:
@willingc can you please explain me how is the source code of oauthenticator related to GitLab? I don't see why we should repeat same naming mistake. I am completely open to change the warning messages or formatting, but I really think that giving correct names to variables is important.

oauthenticator/gitlab.py
```diff
@@ -23,10 +24,32 @@
 
 from .oauth2 import OAuthLoginHandler, OAuthenticator
 
+GITLAB_URL = os.getenv('GITLAB_URL')
+GITLAB_HOST = os.getenv('GITLAB_HOST')
+
+if not GITLAB_URL and GITLAB_HOST:
+    warnings.warn('Use of GITLAB_HOST has been deprecated in v0.8. '
```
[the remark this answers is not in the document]
jirikuncar: I have changed it to `PendingDeprecationWarning`.

CHANGELOG.md
jirikuncar: Let's start with simply by recommending correct naming. The `GITLAB_HOST` will work and it will warn developers that use common sense and set service hostname to variable that it is called `_HOST`.

oauthenticator/gitlab.py
jirikuncar: It's not affecting user's system. It's keeping the global namespace in the python file clean.

minrk:
Thanks for the PR, I think this is all set now.

PR #194, opened 2018-06-11
Add possibility to use self signed certificates with generic auth

mklan:
I have a local keycloak service running and could not authenticate due to self signed certificates.

### DIFF
--- a/oauthenticator/generic.py
+++ b/oauthenticator/generic.py
@@ -7,6 +7,7 @@ import json
 import os
 import base64
 import urllib
+import ast
 
 from tornado.auth import OAuth2Mixin
 from tornado import gen, web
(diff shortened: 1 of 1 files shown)


oauthenticator/generic.py
```diff
@@ -75,6 +76,8 @@ def authenticate(self, handler, data=None):
         code = handler.get_argument("code")
         # TODO: Configure the curl_httpclient for tornado
         http_client = AsyncHTTPClient()
+        
+        tls_verify = ast.literal_eval(os.environ.get('OAUTH2_TLS_VERIFY', 'True'))
```
minrk: A few things:

- this should be a Bool(True, config=True) traitlet above, not a local variable
- instead of `literal_eval`, check for specific values in the env, e.g. (v.lower() in {'true', '1'})

mklan: Thank you for the review. I am new to Python so I need to read things about traitlets first.

mklan: Ok I implemented a new traitlet, but I somehow edited an older version of the file so the commits are a bit messed up.

minrk:
Thanks!

PR #198, opened 2018-06-25
[generic] remove unused userdata_method body

### LINKED ISSUE #197: Potential problem in Generic OAuth implementation.
The generic OAuth implementation has the code:

```
        req = HTTPRequest(url,
                          method=self.userdata_method,
                          headers=headers,
                          validate_cert=self.tls_verify,
                          body=urllib.parse.urlencode({'access_token': access_token})
                          )
```

The ``userdata_method`` is set as:

```
    userdata_method = Unicode(
        os.environ.get('OAUTH2_USERDATA_METHOD', 'GET'),
        config=True,
        help="Userdata method to get user data login information"
    )
```

The problem is that Tornado implements sanity checks and will error if you try and use ``GET`` method where you supply a body for the request. The result is an exception of:

```
ValueError: Body must be None for method GET (unless allow_nonstandard_methods is true)
```

The workaround was to set in the configuration ``userdata_method`` to ``POST`` as the target URL for getting user info must allow both ``GET`` and ``POST``. 

* http://openid.net/specs/openid-connect-core-1_0.html#UserInfo

So this should perhaps default to using ``POST`` for ``userdata_method`` since is accepted anyway. Alternatively, need to pass ``allow_nonstandard_methods=True`` to ``HTTPRequest`` class when initialised.

minrk:
since userdata_method is GET, having a body doesn't make sense. The auth info is already passed in the Authorization header.

closes #197

### DIFF
--- a/oauthenticator/generic.py
+++ b/oauthenticator/generic.py
@@ -137,7 +137,6 @@ class GenericOAuthenticator(OAuthenticator):
                           method=self.userdata_method,
                           headers=headers,
                           validate_cert=self.tls_verify,
-                          body=urllib.parse.urlencode({'access_token': access_token})
                           )
         resp = yield http_client.fetch(req)
         resp_json = json.loads(resp.body.decode('utf8', 'replace'))


guimou:
Hi! Will you merge this also on jupyter-on-openshift/oauthenticator?
I'm using jupyterhub-quickstart from @GrahamDumpleton but OAuthenticator still fails with error 500 as the correction did not make it yet to the repo. Thanks!

GrahamDumpleton:
I can. Have sort of been hoping there would be a new oauthenticator release so not relying on development version.

The quick workaround is to add to the JupyterHub config snippet:

```
os.environ['OAUTH2_USERDATA_METHOD'] = 'POST'
```

before importing setting up any authenticator stuff.

guimou:
OK, great. I had seen your comment regarding the configuration workaround that could be done, but did not know exactly how to proceed. Thanks!

GrahamDumpleton:
I have merged it from master of upstream, although have actually updated packages in ``jupyterhub-quickstart`` to use specific commit from master repo.

Do note you will need to remove:

```
os.environ['OAUTH2_USERDATA_METHOD'] = 'POST'
```

if you added it else it will fail because body is now not supplied and Tornado doesn't like that either.

See https://github.com/jupyterhub/oauthenticator/pull/199#issuecomment-405422705

PR #199, opened 2018-07-10
Removed incorrect body being sent with GET request to USERDATA_URL

robdrysdale:
Last change is creating an issue as USERDATA_URL is by default a GET.  This fixes it for me.  Perhaps more logic around support for whether config is POST or GET and set as parameter or body

### DIFF
--- a/oauthenticator/generic.py
+++ b/oauthenticator/generic.py
@@ -136,9 +136,7 @@ class GenericOAuthenticator(OAuthenticator):
         req = HTTPRequest(url,
                           method=self.userdata_method,
                           headers=headers,
-                          validate_cert=self.tls_verify,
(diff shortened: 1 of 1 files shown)


minrk:
This was merged as #198

GrahamDumpleton:
Just as warning. This change will cause failure for anyone who had previously worked around the problem by using:

```
os.environ['OAUTH2_USERDATA_METHOD'] = 'POST'
```

This is because will now get:

```
    ValueError: Body must not be None for method POST (unless allow_nonstandard_methods is true)
```

The solution obviously is to remove the workaround of setting ``OAUTH2_USERDATA_METHOD``.

Does the configuration setting for ``userdata_method`` serve any purpose now if it always has to be ``GET``. Should ``userdata_methed`` be deprecated and removed?

PR #80, opened 2017-05-10
Allow for a tuple of whitelisted Google Apps domains

gforsyth:
This is an attempt to address #63 

I should note that making `hosted_domain` a `tuple` might mess up existing installations, since traitlets will be angry when it gets a `str` instead of a `tuple`.

Issue #63, opened 2017-01-31
Allowing multiple hosted domain subdomains


gforsyth:
I manage a JupyterHub at George Washington University and we currently use GitHub oauth for logins but we're looking at switching to Google oauth and using the university's google apps accounts. 

Due to some historical quirk, students are assigned a `gwmail.gwu.edu` address while faculty/staff have `email.gwu.edu` -- we want faculty and students to both be able to access the hub. Are there security concerns I'm missing if I change https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/google.py#L111 and remove the `@` from the `endswith` condition? 

I'm not sure if this is a particularly common problem, so I'm happy to maintain a fork for our needs, but if it would be useful I'm also happy to contribute it back in, be it the potential solution above or something else.

minrk:
I think this would be a fine addition.

You'd need to be a little more careful than just removing the @, because`notgwu.edu` also `endswith('gwu.edu')`, but I think this is sensible. I see two options:

1. allow hosted_domain to be a tuple. endswith accepts tuples of strings (`endswith(('a', 'b'))`)
2. allow subdomain matching, such as:

   ```python
   # adding '.' ensures matching only applies to whole domain parts
   ('.' + domain).endswith('.' + ref)
   ```

I can't think of any cases where subdomains *aren't* wanted, but maybe it's more prudent to go the tuple way.

gforsyth:
Hey @minrk -- sorry, just starting to get to this now. 

Should this still use `traitlets` to verify the input? I'm having an issue there since `endswith` recognizes a `tuple` of strings but not a `Tuple` of strings...

minrk:
`Tuple` declared on the class will be a `tuple` on any instance of that class. traitlets would only be used for exposing this as configuration, not the runtime validation of values. Where are you seeing a `Tuple` instance?

I was thinking something along the lines of:

```python

def _with_dot(domain):
    if domain.startswith('.'):
        return domain
    return '.' + domain

class GoogleOAuthenticator(OAuthenticator):
    domains = Tuple(help="""Tuple of domains to match...""", config=True)
    ...
    def authenticate(self):
        ...
        if not self.domains or _with_dot(domain).endswith((_with_dot(d) for d in self.domains)):
            # valid
```

gforsyth:
Ahh, ok, I understand a bit better now.  The way it's set up now just relies on the default value being set when the `Tuple` is defined.

[...]

PR #206, opened 2018-08-16
Allow a list of email domains with Google OAuth 2.0

Misteur-Z:
Rework of #80 
Addressing issue #63 

New features:
* Allows multiple Google OAuth email domains
* Checks for `verified_email` in Google OAuth response
* Log unauthorized attempts as warnings (and not errors)

Sidenotes:
* User `john.doe@domain1.com` and `john.doe@domain2.org` end up with the same jupyterhub user `john.doe` to avoid branking changes

### DIFF
--- a/oauthenticator/google.py
+++ b/oauthenticator/google.py
@@ -6,12 +6,14 @@ Derived from the GitHub OAuth authenticator.
 
 import os
 import json
+import sys
 
 from tornado import gen
 from tornado.auth import GoogleOAuth2Mixin
 from tornado.web import HTTPError
+from tornado.log import app_log
 
-from traitlets import Unicode, default
+from traitlets import Unicode, Tuple, default
 
 from jupyterhub.auth import LocalAuthenticator
 from jupyterhub.utils import url_path_join
@@ -40,10 +42,9 @@ class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
     def _scope_default(self):
         return ['openid', 'email']
 
-    hosted_domain = Unicode(
-        os.environ.get('HOSTED_DOMAIN', ''),
+    hosted_domain = Tuple(
         config=True,
-        help="""Hosted domain used to restrict sign-in, e.g. mycollege.edu"""
+        help="""Tuple of domains used to restrict sign-in, e.g. mycollege.edu"""
     )
     login_service = Unicode(
         os.environ.get('LOGIN_SERVICE', 'Google'),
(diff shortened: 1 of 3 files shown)


betatim:
We should continue to support using a string to set a single domain in addition to allowing people to set it to a tuple/list via the configuration. I think it would be nice to maintain backwards compatibility as there is no good reason to break it here.

What motivated your choice to strip the domain names in the case of multiple domains being allowed? Not allowing joe@fake.com to impersonate joe@email.com seems like a desirable feature to me.

oauthenticator/google.py
```diff
@@ -10,8 +10,9 @@
 from tornado import gen
 from tornado.auth import GoogleOAuth2Mixin
 from tornado.web import HTTPError
+from tornado.log import app_log
```
betatim: this isn't used anywhere

oauthenticator/google.py
```diff
@@ -53,6 +53,11 @@ def _scope_default(self):
 
     @gen.coroutine
     def authenticate(self, handler, data=None):
+        if not self.hosted_domain:
```
betatim: Not requiring a hosted domain to be configured was a feature I think

betatim:
The docs in the README need updating as well.

oauthenticator/google.py
david-dumas: My logic was you should not allow any domain to login if you setup Google Oauth.

betatim: I find it convenient that I can use this to require you have a "google" account but I don't actually care you have a particular domain. For example for workshops or other open events where people have all sorts of "google" accounts.

david-dumas:
@betatim I agree maintaining backwards compatibility is better, my python knowledge is limited and I don't know how to solve those two issues cleanly:
* `os.environ.get('HOSTED_DOMAIN', '')` should be converted to a tuple of one string and we have to chose if it overrides `c.GoogleOAuthenticator.hosted_domain` from configuration file
* `hosted_domain` could already be a string or a tuple, the code in `authenticate()` would already work

I will update the README once this is done.

david-dumas:
* Used a List() instead of a Tuple()
* Updated the README
* Backwards compatible

joe@fake.com can impersonate joe@email.com but if we add the user's email domain to it's username [here](https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/google.py#L89) to avoid it we create a breaking change...

cmseal:
+1 for this... I had to do something similar, as we've deployed JupyterHub across as group of companies, each with their own domain, but I kept the domain in the username, and just replaced the '@' with another '.' so that joe.bloggs@test.com is different from joe.bloggs@test.co.uk.

minrk:
Thanks for the PR! I added a couple of commits to avoid breaking changes:

1. resolve conflicts
2. only an empty list means no setting, not `['']` (and default HOSTED_DOMAIN='' env results in an empty list now)
3. only strip domain when hosted_domain has only one item, so that collisions cannot occur when allowing multiple domains, like what @cmseal described

PR #209, opened 2018-08-24
fix parsing of dict env variables

### LINKED ISSUE #208: Environment variable OAUTH2_AUTHENTICATION_PARAMS is parsed incorrectly
It is parsed here:
https://github.com/jupyterhub/oauthenticator/blob/8277ae61f471fac5c054b1e7980da00c84ed3c4e/oauthenticator/generic.py#L62-L65
I see no way to set this, os.environ.get will always return a string (if the variable exists) and 
the Dict constructor expects a python dict

Is the some special string formatting i have to use?
I think the code should be something like this:
```python
userdata_params = Dict( 
     json.loads(os.environ.get('OAUTH2_USERDATA_PARAMS', '{}')), 
     help="Userdata params to get user data login information" 
 ).tag(config=True) 
```

mhuttner:
fixes #208

### DIFF
--- a/oauthenticator/generic.py
+++ b/oauthenticator/generic.py
@@ -50,7 +50,6 @@ class GenericOAuthenticator(OAuthenticator):
         help="Access token endpoint URL"
     )
     extra_params = Dict(
-        json.loads(os.environ.get('OAUTH2_AUTHENTICATION_PARAMS', '{}')),
         help="Extra parameters for first POST request"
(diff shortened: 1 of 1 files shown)


minrk:
Since this has never worked, can we instead remove the environment variables here? Not everything should be settable via environment variables, instead opting for using the standard `jupyterhub_config.py`.

PR #223, opened 2018-10-24
[GitHubOAuthenticator] Implement a team/repo whitelist

tmshn:
⚠️ This is still wip implementation.
Before going on, I want to hear from maintainers and some users.

# Background

On Github, organization is not the only way to handle permission, but also we usually create teams or set per-repository collaborators. It varies on each companies/universities.

Using those resources in jupyterhub's authentication will help us a lot.

# Changed

Support team-level or repository-level whitelisting.

Instead of doing

```python
GitHubOAuthenticator.organization_whitelist(
    'foo',
    'bar',
)
```

you can do

```python
GitHubOAuthenticator.whitelist(
    'orgs/foo',       # `foo` organization member
    'repo/baz:write', # Users with write permission to `baz` repo
    'teams/123'       # Member of team which id is 123
)
```

# Status

Currently I implemented the main logic, but not fixed the option parser etc. Tests aren't written yet too.

Before going on the implementation, I want to hear from maintainers and users.

- Does this idea seems good and rational?
- Is there any implementation concern?

### DIFF
--- a/oauthenticator/github.py
+++ b/oauthenticator/github.py
@@ -166,23 +166,64 @@ class GitHubOAuthenticator(OAuthenticator):
         return userdict
 
     @gen.coroutine
-    def _check_organization_whitelist(self, org, username, access_token):
+    def _check_whitelist(self, target, username, access_token):
+        # Check if the user has permission against the target.
+
+        # If target is specified with `:`, treat the latter part as level specifier.
+        if ':' in target:
+            target, level = target.split(':', 1)
+        else:
+            level = None
+
+        if target.beginswith('orgs/'):
+            # https://developer.github.com/v3/orgs/members/#check-membership
+            # With empty scope (even if authenticated by an org member), this
+            #  will only check public org members.  You want 'read:org' in order
+            #  to be include private members.
+            resp = yield self._call_github(target, username, 'members/{username}', access_token)
+            if resp.get('state') != 'active':
+                return False
+            if level == 'admin':
+                return resp.get('role') == 'admin'
+            return True
+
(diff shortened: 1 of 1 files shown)


manics:
Team auth definitely sounds nice!

I've never thought of using repo-level auth for JupyterHub, but if you've got a direct link between users and a repo with write auth you could potentially pre-configure git in jupyter to push direct to the repo?

tmshn:
> if you've got a direct link between users and a repo with write auth you could potentially pre-configure git in jupyter to push direct to the repo?

Actually I think this can be done without this PR.
Just adding `public_repo` or `repo` scope will give you token which you can use to push to github. You can pre-configure it to user environment.
See this example:
https://github.com/jupyterhub/oauthenticator/tree/master/examples/auth_state

(N.B.: I don't think we can get "repository-specific" token...token with `repo` scope can push to any repository that the user has write access.)

sodre:
I think this would be a great addition, but I am not in love with the notation. It would be nice to use some of the existing syntax from GitHub to control access, e.g. 

```
GitHubOAuthenticator.whitelist(
    'foo',       # `foo` organization member, this is still backwards compatible.
    '@foo/awesome-team'       # Member of awesome-team in org foo.
    'foo/baz', # Users with write permission to `baz` repo. I am not sure we really need this one
)
```

consideRatio:
Is this PR still relevant, does this functionality perhaps already exist within the GitHub authenticator?

manics:
I'm closing this as it seems to be stale. Please feel free to open a new PR, or start a discussion on https://discourse.jupyter.org/

Issue #213, opened 2018-09-06
Do not use Azure AD display name as username


thomas-rabiller-azimut:
While using the AzureAdOAuthenticator to authenticate with Oauth to the Azure AD provider, the property "name" from the requested JWT token is used as username in JupyterHub. According to the [Azure AD documentation](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-id-and-access-tokens):

> The name claim provides a human-readable value that identifies the subject of the token. The value is not guaranteed to be unique, it is mutable, and it's designed to be used only for display purposes

It is therefore not reliable as a username (not unique, can contain spaces and special chars...)

I suggest using the object id instead, which is exactly the common unix username max size (32 chars long after removing the dashes in guid)

This is what I am doing currently in a project by overriding the behaviour in a inherited class, but it should probably be applied in future versions by default.

vwoloszyn:
I open an issue regarding this behavior of using the AD display name as username. In short, jupyterhub is not allowing to autheticate users because the display name contain an illegal char "/"
https://github.com/jupyterhub/jupyterhub/issues/3996

PR #224, opened 2018-11-05
AzureAD: use AD OID as username to ensure unicity

thomas-rabiller-azimut:
This PR requires admin review for compliance with project architecture and potential further testing

### Background
This fix is a suggestion to solve issue #213 concerning Azure AD authentication.

### Changed
- Use `OID` from AD as JupyterHub username instead of `name` property as it is today because the `name` value is not immutable
- Apply normalization when used with local authenticator to comply with unix username restrictions

### DIFF
--- a/oauthenticator/azuread.py
+++ b/oauthenticator/azuread.py
@@ -10,6 +10,7 @@ import re
 import string
 import urllib
 import sys
+import unidecode
 
 from tornado.auth import OAuth2Mixin
 from tornado.log import app_log
@@ -102,15 +103,23 @@ class AzureAdOAuthenticator(OAuthenticator):
         id_token = resp_json['id_token']
         decoded = jwt.decode(id_token, verify=False)
 
-        userdict = {"name": decoded['name']}
+        # use AD OID as username to ensure unicity (name is not immutable)
+        # https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-id-and-access-tokens
+        userdict = {"name": decoded['oid']} 
         userdict["auth_state"] = auth_state = {}
         auth_state['access_token'] = access_token
         # results in a decoded JWT for the user data
         auth_state['user'] = decoded
 
         return userdict
+    
 
 
 class LocalAzureAdOAuthenticator(LocalAuthenticator, AzureAdOAuthenticator):
+    def normalize_username(self, username):
+        # fix azure ad guid to match unix username standards (no accents or dashes)
+        unaccented_string = unidecode.unidecode(username)
+        return unaccented_string.lower().replace('-', '')
+
     """A version that mixes in local system user creation"""
     pass
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,2 @@
 jupyterhub>=0.5
+Unidecode>=1.0.22


sdementen:
Hello, could it be possible to provide a function to extract the name from the decoded JWT ?
The oid is a uuid that is not human friendly when digging into the username/folders/etc created for jupyterhub.
In my case, I would rather prefer to use the 'unique_name' (userid@blable.com) and extract the userid, i.e. something like:
`userdict = {"name": decoded['unique_name'].split("@")[0]}`
So if we could provide a python expression in the config to extract the username, would be great!

thomas-rabiller-azimut:
I agree the `oid` isn't very human friendly when looking at the folder structure. However it looks like the `unique_name` property is not safe to use, according to [AAD documentation](https://docs.microsoft.com/en-gb/azure/active-directory/develop/access-tokens):

> unique_name: Only present in v1.0 tokens. Provides a human readable value that identifies the subject of the token. This value is not guaranteed to be unique within a tenant and should be used only for display purposes.

This has similar risks as the current `name` property:

> name: Only present in v1.0 tokens. Provides a human readable value that identifies the subject of the token. This value is not guaranteed to be unique within a tenant and should be used only for display purposes.

sdementen:
I would rather leave this choice to the user/configuration.
We just deployed a jupyterhub and as admin I see all users by their internal id (ie a part of the unique_name) which is not convenient either.
We will probably move to a mix of user friendly name + unique id.
Today, we are patching the azuread.py. would be great if tomorrow we could do this by configuration

jfleury-eidos:
@thomas-rabiller-azimut, are you still working on this PR? I'd really like to see it merged.

thomas-rabiller-azimut:
@jfleury-eidos The PR is complete as it is as far as I am concerned (fix is as I expect and tests have passed) but I am not sure what needs to be done for the jupyter Hub team to review and integrate it in the official repo...

jfleury-eidos:
@sdementen asked that the username value be configurable.

dwaltrip:
Using the `oid` seems very suboptimal, as it becomes much more difficult for users to see who they are logged in as, as well as for admins trying to look at user info.

I understand the documented warnings around `unique_name`, but in practice how likely is it for the values to not be unique? Perhaps we could append `-1`, `-2`, whenever a conflict is found?

Also, a nice configurable escape hatch might be a value such as, `azuread_username_key`, which determines which key to use to set the username. One could set `c.AzureAdOAuthenticator.azuread_username_key = 'oid'` if they choose.

timlod:
We are facing the same issues.
I'm now patching to use unique_name, which in all of the current cases is equivalent to using email.

An interesting note to add to the discussion:
Azure offers the AADLoginForLinux extension for Azure Virtual Machines.
If you login using your AD credentials, at least in my case the system uses the email address prefix (firstname.lastname) as the username.
By using the email address in JupyterHub we get the advantage of being able to log into the same account using SSH AD login as well as JupyterHub Azure AD authentication.

The problem seems indeed to be that v2.0 tokens don't have an email field. The example says that preferred_name can be an email address, but it is mutable.

I agree that giving an option would be the best way to fix this.
If I can find time in the near future I may submit a PR.

Issue #226, opened 2018-11-08
GitHubOAuthenticator: another way to check org_whitelist


xuwaters:
Currently the algorithm of checking organization whitelist is to call
api `/orgs/%s/members` which requires the organization members to be publicly visible.

I think we can call api `/user/orgs` to get current user's orgs and check whether the `login` field of each org item is in the configured `org_whitelist`

Pros and Cons of this method:
Pros: 
- only one api call (`/user/orgs`) to check user's membership instead of call (`/orgs/%s/members`) for each configured organization
- even organization members is not publicly visible, user's membership of organization can be checked correctly

Cons:
- require permission of `user` scope instead of `read:user` scope of the login user

xuwaters:
Sorry, I don't think `user` scope is a good idea, which requires read/write access to profile info.

manics:
You can already get access to private GitHub org memberships by changing `.scope`. The exact rules are a bit complicated though, for some discussion see:
- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/741
- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/687

xuwaters:
Thank you @manics 
I followed your links and tested `read:org` scope, it works.

And I found another api `GET /orgs/:org/public_members/:username`, 
I think this api is more efficient than get all members and check member login one by one.

PR #227, opened 2018-11-08
GitHubOAuthenticator: check org_whitelist using check-membership api

xuwaters:
Implement org_whitelist check using GitHub api [check-membership](https://developer.github.com/v3/orgs/members/#check-membership)
#226

### DIFF
--- a/oauthenticator/github.py
+++ b/oauthenticator/github.py
@@ -169,19 +169,15 @@ class GitHubOAuthenticator(OAuthenticator):
     def _check_organization_whitelist(self, org, username, access_token):
         http_client = AsyncHTTPClient()
         headers = _api_headers(access_token)
-        # Get all the members for organization 'org'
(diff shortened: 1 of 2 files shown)


manics:
Unfortunately I haven't got time to test this at the moment. As mentioned in https://github.com/jupyterhub/oauthenticator/issues/226#issuecomment-436933318 the rules for getting the existing org authentication to work was complicated and took a lot of investigation with regards to public/private membership, different authenticator scopes, and permissions on the GitHub org. Could you outline what configs you have or haven't tested?

minrk:
I was able to test this and it worked even better than I thought. In requesting `read:org` access I specifically elected not to grant read-access to the orgs that require such opt-ins to see how that would fail. I was able to still login based on my membership in those organizations, even though I hadn't granted jupyterhub read-access to those orgs, so I couldn't see the full user list. So even without granting read access to the org membership list, you can check the authorized user's membership in the organization.

I went ahead and added some debug-logging to the process, which helped me verify that this is working.

PR #233, opened 2018-12-19
Customize the "Sign in" button

T0mWz:
Make it possible to customize the "Sign in" button, with the name of your login service.
```
LOGIN_SERVICE="Miskatonic account"
```

### DIFF
--- a/oauthenticator/generic.py
+++ b/oauthenticator/generic.py
@@ -33,8 +33,9 @@ class GenericLoginHandler(OAuthLoginHandler, GenericEnvMixin):
 class GenericOAuthenticator(OAuthenticator):
 
     login_service = Unicode(
-        "GenericOAuth2",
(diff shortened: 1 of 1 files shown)


minrk:
The `config=True` makes this already configurable via `c.GenericOAuthenticator.login_service = 'my service'`. Does that not work for you?

T0mWz:
Thanks, you're right! Pff. I was struggling with environment variables, no idea why I didn't get this working. This helps, so pull request can be closed.  

But maybe an idea to make this variable configurable too via the bash environment variable, like the other options?

Thanks for your help! 👍 😃

consideRatio:
I'll go ahead and close this PR as it is not important this is configurable through an environment variable. Thanks for contributing @T0mWz !! :heart:

PR #236, opened 2019-02-01
Allow supplementary username_claim list

athornton:
The new functionality hasn't been tested; I'm waiting on some client OAuth work at CILogon to do it.  It doesn't break anything when called *without* the new parameter.

Once I get that go-ahead and test against CILogon I will update the comments, and update the sources if need be.


minrk:
Nice!

PR #241, opened 2019-02-27
add entrypoints for authenticators

minrk:
so that they can be loaded by shortcut in jupyterhub 1.0


willingc:
Thanks @minrk. Pretty cool to be preparing for 1.0 ;-)

Issue #250, opened 2019-03-18
OAuth2Mixin mixed in too late (in generic.py / google.py etc instead of in oauth2.py)


consideRatio:
The OAuthLoginHandler defined in `oauth2.py` is using `self.authorize_redirect` (Line 85), but that is a function not defined in the class, or its base class `jupyterhub.handlers.BaseHandler`, or its base class `tornado.web.RequestHandler`. It is simply assumed that the `OAuthLoginHandler` class will be subclassed and get the `tornado.auth.OAuth2Mixin`. This took a while for me to understand ;D

https://github.com/jupyterhub/oauthenticator/blob/5b0b395326292d526a4c924c34c6d9d1d06818af/oauthenticator/oauth2.py#L59-L90

I figure it would be fine to simply mixin the OAuth2Mixin to the `OAuthLoginHandler` as it actually needs it, and remove the OAuth2Mixin from here for example.

https://github.com/jupyterhub/oauthenticator/blob/5b0b395326292d526a4c924c34c6d9d1d06818af/oauthenticator/generic.py#L24-L30

minrk:
Closing in favor of #312 which should solve this and we shouldn't need two issues to track it.

PR #251, opened 2019-03-18
Use OAuth2Mixin on base OAuthLoginHandler

consideRatio:
I struggled with understanding the Authenticator classes and ran into #250. This PR removes mixing in the OAuth2Mixin from tornado.auth unless it is actually used by the classes, and makes sure to mix it in when it is actually needed.

The PR seems functional to me without breaking any tests. But, I have not tested things on a JupyterHub and made an actual login to google etc though. Speaking of that, what would the best way to do that be? It is unreasonable to me to test all authenticator classes but perhaps I could setup and run a E2E test with say google auth. Please help me decide on the level of verification needed before merge.

### DIFF
--- a/oauthenticator/auth0.py
+++ b/oauthenticator/auth0.py
@@ -32,7 +32,6 @@ jupyterhub_config.py :
 import json
 import os
 
-from tornado.auth import OAuth2Mixin
 from tornado import gen, web
 
 from tornado.httpclient import HTTPRequest, AsyncHTTPClient
@@ -43,7 +42,7 @@ from .oauth2 import OAuthLoginHandler, OAuthenticator
 
 AUTH0_SUBDOMAIN = os.getenv('AUTH0_SUBDOMAIN')
 
-class Auth0Mixin(OAuth2Mixin):
+class Auth0Mixin():
     _OAUTH_AUTHORIZE_URL = "https://%s.auth0.com/authorize" % AUTH0_SUBDOMAIN
     _OAUTH_ACCESS_TOKEN_URL = "https://%s.auth0.com/oauth/token" % AUTH0_SUBDOMAIN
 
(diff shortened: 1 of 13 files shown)


betatim:
I'm not sure I have an opinion on this :-/ I think it would be good to understand why it was done the way it is currently. It makes some amount of sense to my programmer brain to do things the original way (especially if we didn't cut off the comment in one of the docstrinngs ...).

Overall I wouldn't want to decide either way without having some input from the original creators @minrk and/or @rgbkrk. However they are both on leave at the moment :-/

minrk:
It's a little confusing because we are mixing in generic implementations from two sources:

1. OAuth2Mixin which is a necessarily incomplete base class from tornado for a mixin that defines the API for `authorize_redirect` and `_OAUTH_AUTHORIZE_URL` etc.
2. OAuthLoginHandler, which is also an incomplete base class that calls `authorize_redirect`, thereby assuming final classes inherit from a completed OAuth2Mixin subclass
3. provider-specific mixins that should ~only set the values for `_OAUTH_AUTHORIZE_URL`

I think this PR is a step in the right direction, but it adds an additional complexity that the final Handler classes are now sensitive to inheritance order, because both parents (ProviderMixin and OAuthLoginHandler) define `_OAUTH_AUTHORIZE_URL`. I think it probably makes sense to go one step further and eliminate the ProviderMixins entirely and define the provider-specific `_OAUTH` attributes directly on the final Handler classes. Does that sound good?

consideRatio:
@minrk a very concrete idea that currently makes sense to me! I'll work on it! I appreciate how clearly you can summarize and describe things :D

consideRatio:
Closed in favor of suggested approach, which can be opened in another PR. See issue #312 to remember this going onwards.

PR #257, opened 2019-04-08
OpenShift 4.0 splits OAuth provider out of REST API URL host.

GrahamDumpleton:
From OpenShift 4.0, the OAuth provider and Kube REST API are not on the same URL. It is necessary to specify them separately via distinct environment variables.

Prior to OpenShift 4.0, you set the the single `OPENSHIFT_URL` environment variable. For OpenShift 4.0+, you now would need to set `OPENSHIFT_AUTH_API_URL` and `OPENSHIFT_REST_API_URL`.

This change also updates the URL path for the REST API query for user information, to use new namespaced URL path convention supported by Kubernetes. This is because old URL path no longer works in OpenShift 4.0+.

This change will mean the current code will no longer work for OpenShift versions older than 3.6. That OpenShift version is quite old now, and so seen as acceptable to make change. If someone really needs to use this with OpenShift 3.5 or older, then can override:

```
OpenShiftOAuthenticator.users_rest_api_path
```

from JupyterHub config file, setting it back to older path of:

```
/oapi/v1/users/~
```

Note that the Jupyter on OpenShift project's deployment method for JupyterHub likely will not work for versions older that OpenShift 3.6, so unlikely this situation would arise, unless someone has been maintaining there own way of deploying JupyterHub to OpenShift.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,8 +204,13 @@ c.GoogleOAuthenticator.login_service = 'My College'
 In case you have an OpenShift deployment with OAuth properly configured (see the
 following sections for a quick reference), you should set the client ID and
 secret by the environment variables `OAUTH_CLIENT_ID`, `OAUTH_CLIENT_SECRET` and
-`OAUTH_CALLBACK_URL`. The OpenShift API URL can be specified by setting the
-variable `OPENSHIFT_URL`.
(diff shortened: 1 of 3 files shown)


minrk:
Thanks!

PR #261, opened 2019-05-14
Allow option to disable basic auth header in generic oauth

sunnielyu:
- Basic authentication still enabled by default
- Adds configuration parameter to remove Authorization header (basic_auth)
- User can send client_id and client_secret in POST body via extra_params configuration parameter


minrk:
Makes sense, thanks!

PR #263, opened 2019-05-15
Add support for nested userdata in generic authenticator.

janLo:
Services like NextCloud return a nested data-structure at the userdata
endpoint. A single key to get the username is not sufficient here.
This adds a new envvar "OAUTH2_USERNAME_KEY_NESTED" to tell that the
username-key is a dot-separated sequence of keys to access nested
userdata responses. For Nextcloud this would be "ocs.data.id".

Unittests are implemented to test also with nested data.

### DIFF
--- a/oauthenticator/generic.py
+++ b/oauthenticator/generic.py
@@ -58,6 +58,11 @@ class GenericOAuthenticator(OAuthenticator):
         config=True,
         help="Userdata username key from returned json for USERDATA_URL"
     )
+    username_key_nested = Bool(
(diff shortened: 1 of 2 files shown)


janLo:
Hello, any requests or news on this matter?

yuvipanda:
Apologies for the very late response, @janLo. I just merged https://github.com/jupyterhub/oauthenticator/pull/305 which should support this use case, but in a slightly different way. Let me know if that works for you?

PR #268, opened 2019-06-01
[GenericOAuthenticator] Usage example for Nextcloud

arneki:


arneki:
Oh, I just realized that #263 also solves the problem with nested userdata, even including tests. I will adjust my pull request if this is merged first...

arneki:
Thanks for the review!
Stripped out everything except of the documentation and adjusted it to the changes introduced in #305.

minrk:
Sorry for leaving you hanging, @arneki, but can you re-issue this PR against the new docs which are now in docs/source/getting-started.rst? If not, I'm happy to translate the example to rst and add it.

minrk:
Thanks!

PR #275, opened 2019-06-28
adopt async def instead of gen.coroutine

### LINKED ISSUE #246: github closes stream after token request
When I try to authenticate using GitHubOAuthenticator I always get Exception "Stream closed".
It happens at step 2 of OAuth2 process - "https://github.com/login/oauth/authorize" executes normally and returns code, but when authenticator tries to fetch "https://github.com/login/oauth/access_token?client_id=..." exception is thrown. I found out that it happens in "tornado/simple_httpclient.py" when `_write_body` method tries to `await self.connection.read_response(self)`.

Although:
1) I tried to fetch other arbitrary URLs from within GitHubOAuthenticator' `authenticate` method - it works fine.
2) I tried to fetch GitHub's token manually with the help of curl - it works fine.
3) I even found and tried to use YandexPassportOAuthenticator which has almost identical code - it works fine!

Does anyone have any ideas why this can happen?

minrk:
avoid deprecated interface which has become buggy with tornado 6,
causing things like StreamClosedError on GitHub oauth (#246)

closes #246
closes #270

This bumps our minimum Python to 3.5, which may have been true already, but makes it official in package metadata

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,10 @@
 language: python
 sudo: false
 cache: pip
+dist: xenial
 python:
   - nightly
+  - 3.7
   - 3.6
   - 3.5
 install:
@@ -24,5 +26,5 @@ after_success:
 
 matrix:
   include:
-    python: 3.6
-    env: TEST_LINT=1
+    - python: 3.7
+      env: TEST_LINT=1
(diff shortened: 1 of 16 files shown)


betatim:
Merging now as I think this isn't controversial and if it turns out to be we can easily revert it.

PR #276, opened 2019-07-17
[GenericOAuthenticator] Add group support

rochaporto:
Update the generic oauthenticator to support groups.

This includes:
* a new group_whitelist config param, taking a list of allowed groups
* a new groupdata_url param, similar to userdata_url but retrieving group info
* a new groupdata_method param, similar to userdata_method

### DIFF
--- a/oauthenticator/generic.py
+++ b/oauthenticator/generic.py
@@ -16,7 +16,7 @@ from tornado.httpclient import HTTPRequest, AsyncHTTPClient
 
 from jupyterhub.auth import LocalAuthenticator
 
-from traitlets import Unicode, Dict, Bool
+from traitlets import Unicode, Dict, Bool, Set
 
 from .oauth2 import OAuthLoginHandler, OAuthenticator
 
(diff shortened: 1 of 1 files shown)


oauthenticator/generic.py
```diff
@@ -170,6 +192,25 @@ class GenericOAuthenticator(OAuthenticator):
...
+            raise ValueError("Please set the OAUTH2_GROUPDATA_URL environment variable")
+
+        req = HTTPRequest(url,
+                          method=self.groupdata_method,
+                          headers=headers,
+                          validate_cert=self.tls_verify,
+                          )
+        resp = await http_client.fetch(req)
+        resp_json = json.loads(resp.body.decode('utf8', 'replace'))
+        user_groups = resp_json.get('groups')
```
minrk: Is the 'groups' field and structure generic to oauth? What provider have you used this with?

gweis: I think this is really provider specific. Even fetching data from group_data_url is provider specific. Usually group member ship is either in an id_token (OIDC), can be fetched from a user info endpoint or maybe taken from an access token (or token introspection endpoint). Anything else is outside of the oauth2 or OIDC spec.

consideRatio:
@minrk - could you look at this PR?

minrk:
@consideRatio I don't think this is ready to go because it makes assumptions about groups (i.e. a 'groups' field is defined and is a list of group names). I think we need to show at least two or three real cases of group membership that this will work for with existing providers to show that it really is generic.

rochaporto:
I'm closing this as indeed it implies semantics on how groups are defined. We will handle this in another way internally.

PR #280, opened 2019-08-05
add configurable username claim

jeff-sternberg:
This is a suggestion to fix https://github.com/jupyterhub/oauthenticator/issues/213 that is similar to https://github.com/jupyterhub/oauthenticator/pull/224, but provides a configurable `username_claim` attribute.

Logic:
1. Read value of `c.AzureAdOAuthenticator.username_claim`
1. Fall back to <del>`oid`</del>name if config isn't set

closes #224

### DIFF
--- a/oauthenticator/azuread.py
+++ b/oauthenticator/azuread.py
@@ -51,6 +51,7 @@ class AzureAdOAuthenticator(OAuthenticator):
     login_handler = AzureAdLoginHandler
 
     tenant_id = Unicode(config=True)
+    username_claim = Unicode(config=True)
 
     def get_tenant(self):
         if hasattr(self, 'tenant_id') and self.tenant_id:
@@ -63,6 +64,16 @@ class AzureAdOAuthenticator(OAuthenticator):
                 help="Tenant")
             app_log.info('ID4: {0}'.format(tenant_id))
             return tenant_id
+
+    def get_username_claim(self):
+        """
+        The claim to map to the jupyter username, such as `upn` or `unique_name`
+        See https://docs.microsoft.com/en-gb/azure/active-directory/develop/id-tokens
+        """
+        if hasattr(self, 'username_claim') and self.username_claim:
+            app_log.info('ID5: {0}'.format(self.username_claim))
+            return self.username_claim
+        return 'oid'
 
     async def authenticate(self, handler, data=None):
         code = handler.get_argument("code")
@@ -101,7 +112,7 @@ class AzureAdOAuthenticator(OAuthenticator):
         id_token = resp_json['id_token']
         decoded = jwt.decode(id_token, verify=False)
 
-        userdict = {"name": decoded['name']}
+        userdict = {"name": decoded[self.get_username_claim()]}
         userdict["auth_state"] = auth_state = {}
         auth_state['access_token'] = access_token
         # results in a decoded JWT for the user data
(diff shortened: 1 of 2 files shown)


oauthenticator/azuread.py
```diff
@@ -64,6 +65,16 @@ def get_tenant(self):
...
 
+    def get_username_claim(self):
+        """
+        The claim to map to the jupyter username, such as `upn` or `unique_name`
+        See https://docs.microsoft.com/en-gb/azure/active-directory/develop/id-tokens
+        """
+        if hasattr(self, 'username_claim') and self.username_claim:
+            app_log.info('ID5: {0}'.format(self.username_claim))
+            return self.username_claim
+        return 'oid'
```
betatim: The default used to be `'name'`, unless there is a good reason I wouldn't change this. If we do change it we need to think about if or not this will break current users (and then how to deal with that)

oauthenticator/azuread.py
```diff
@@ -64,6 +65,16 @@ def get_tenant(self):
             app_log.info('ID4: {0}'.format(tenant_id))
             return tenant_id
 
+    def get_username_claim(self):
+        """
+        The claim to map to the jupyter username, such as `upn` or `unique_name`
+        See https://docs.microsoft.com/en-gb/azure/active-directory/develop/id-tokens
+        """
+        if hasattr(self, 'username_claim') and self.username_claim:
```
betatim: I'd use the possibility to set a default value for the traitlet (https://traitlets.readthedocs.io/en/stable/using_traitlets.html#default-values-and-checking-type-and-value) to remove the need for testing for attributes etc. Probably means we can even remove all of `get_username_claim` and directly use `self.username_claim` further down

oauthenticator/azuread.py
jeff-sternberg: Makes sense, changed.

oauthenticator/azuread.py
jeff-sternberg: Sounds good, please take a look at the updated version. Should we do the same for `get_tenant`?

oauthenticator/azuread.py
```diff
@@ -6,23 +6,16 @@
...
-
-from tornado.httputil import url_concat
 from tornado.httpclient import HTTPRequest, AsyncHTTPClient
 
 from jupyterhub.auth import LocalAuthenticator
 
-from traitlets import List, Set, Unicode
+from traitlets import Unicode, default
 
-from .common import next_page_from_links
```
betatim: Just to double check: all these removed imports were unused ones or is there something to double check here?

betatim:
Thanks for the quick turn around! This PR looks ready to go after my double check comment. I'd merge this as is and then we can have a second PR to switch over `get_tenant` (or make an issue for later).

oauthenticator/azuread.py
jeff-sternberg: yeah they were all unused - at least according to PyCharm! Thanks

minrk:
> Fall back to oid if config isn't set

Looks like the fallback is to `name` if it isn't set. Does this PR mean to change this to oid?

jeff-sternberg:
@minrk originally yes that was the intent, but as @betatim suggested, if we fall back to `name` that will be backwards compatible. Also with this change, a user can configure `oid` if desired.

minrk:
Sorry, should have read through better. Thanks!

jeff-sternberg:
No worries, and thanks for merging!

PR #282, opened 2019-08-22
Use traitlet.default for Azure AD tenant_id

jeff-sternberg:
This implements the suggestion in https://github.com/jupyterhub/oauthenticator/pull/280 to simplify `tenant_id` by using `traitlet.default`. This allows removing the `get_tenant()` function entirely. As before, the default value is retrieved from the `AAD_TENANT_ID` environment variable.


minrk:
Nice!

PR #283, opened 2019-08-23
Include inherited members in GitLab auth checks

vindvaki:
When triaging a customer issue with the JupyterHub GitLab integration, 
we noticed that some members with access developer or above were
getting 403s when trying to log in via GitLab. The problem, we found, is
that the API call used by JupyterHub only retrieves direct group or project
members, as opposed to all members including inherited ones.

This is easily fixed by adding /all to the call.

See https://docs.gitlab.com/ee/api/members.html#list-all-members-of-a-group-or-project-including-inherited-members

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ c.MyOAuthenticator.client_secret = 'your-client-secret'
 > export AAD_TENANT_ID='{AAD-TENANT-ID}'
 ```
 
-* Sample code is provided for you in `examples > azuread > sample_jupyter_config.py` 
+* Sample code is provided for you in `examples > azuread > sample_jupyter_config.py`
 * Just add the code below to your `jupyterhub_config.py` file
 * Making sure to replace the values in `'{}'` with your APP, TENANT, DOMAIN, etc. values
 
@@ -180,6 +180,16 @@ You can also use `LocalGitLabOAuthenticator` to map GitLab accounts onto local u
 
 You can use your own GitLab CE/EE instance by setting the `GITLAB_HOST` environment
 flag.
+
+You can restrict access to only accept members of certain projects or groups by setting
+```
+c.GitLabOAuthenticator.gitlab_project_id_whitelist = [ ... ]
+```
+and
+```
+c.GitLabOAuthenticator.gitlab_group_whitelist = [ ... ]
+```
+but be aware that each entry incurs a separate API call, increasing the risk of rate limiting and timeouts.
 
 ## Google Setup
 
--- a/oauthenticator/gitlab.py
+++ b/oauthenticator/gitlab.py
@@ -8,6 +8,7 @@ based on the GitHub plugin by Kyle Kelley (@rgbkrk)
 
 import json
 import os
+import re
 import sys
 import warnings
 
@@ -127,7 +128,7 @@ class GitLabOAuthenticator(OAuthenticator):
         # memoize gitlab version for class lifetime
         if self.gitlab_version is None:
             self.gitlab_version = await self._get_gitlab_version(access_token)
-            self.member_api_variant = 'all/' if self.gitlab_version >= '12.4' else ''
+            self.member_api_variant = 'all/' if self.gitlab_version >= [12, 4] else ''
 
         # Determine who the logged in user is
         req = HTTPRequest("%s/user" % GITLAB_API,
@@ -180,7 +181,9 @@ class GitLabOAuthenticator(OAuthenticator):
                           validate_cert=self.validate_server_cert)
         resp = await AsyncHTTPClient().fetch(req, raise_error=True)
         resp_json = json.loads(resp.body.decode('utf8', 'replace'))
-        return resp_json['version']
+        version_string = re.sub(r'-pre$', '', resp_json['version'])
+        return list(map(int, str.split(version_string, '.')))
+
 
     async def _check_group_whitelist(self, user_id, access_token):
         http_client = AsyncHTTPClient()
--- a/oauthenticator/gitlab.py
+++ b/oauthenticator/gitlab.py
@@ -10,7 +10,6 @@ import json
 import os
 import sys
 import warnings
-import urllib
 
 from tornado.auth import OAuth2Mixin
 from tornado import web
--- a/oauthenticator/gitlab.py
+++ b/oauthenticator/gitlab.py
@@ -186,7 +186,7 @@ class GitLabOAuthenticator(OAuthenticator):
         headers = _api_headers(access_token)
         # Check if user has developer access to any project in the whitelist
         for project in self.gitlab_project_id_whitelist:
-            url = "%s/projects/%s/members/%d" % (GITLAB_API, project, user_id)
+            url = "%s/projects/%s/members/%d/all" % (GITLAB_API, project, user_id)
             req = HTTPRequest(url, method="GET", headers=headers)
             resp = await http_client.fetch(req, raise_error=False)
 
--- a/oauthenticator/gitlab.py
+++ b/oauthenticator/gitlab.py
@@ -144,12 +144,12 @@ class GitLabOAuthenticator(OAuthenticator):
 
         if self.gitlab_group_whitelist:
             is_group_specified = True
-            user_in_group = await self._check_group_whitelist(user_id, access_token)
+            user_in_group = await self._check_group_whitelist(username, access_token)
 
         # We skip project_id check if user is in whitelisted group.
         if self.gitlab_project_id_whitelist and not user_in_group:
             is_project_id_specified = True
-            user_in_project = await self._check_project_id_whitelist(user_id, access_token)
+            user_in_project = await self._check_project_id_whitelist(username, access_token)
 
         no_config_specified = not (is_group_specified or is_project_id_specified)
 
@@ -168,12 +168,12 @@ class GitLabOAuthenticator(OAuthenticator):
             return None
 
 
-    async def _check_group_whitelist(self, user_id, access_token):
+    async def _check_group_whitelist(self, username, access_token):
         http_client = AsyncHTTPClient()
         headers = _api_headers(access_token)
         # Check if user is a member of any group in the whitelist
         for group in map(url_escape, self.gitlab_group_whitelist):
-            url = "%s/groups/%s/members/%d/all" % (GITLAB_API, group, user_id)
+            url = "%s/groups/%s/members/all?query=%s" % (GITLAB_API, group, username)
             req = HTTPRequest(url, method="GET", headers=headers)
             resp = await http_client.fetch(req, raise_error=False)
             if resp.code == 200:
@@ -181,12 +181,12 @@ class GitLabOAuthenticator(OAuthenticator):
         return False
 
 
-    async def _check_project_id_whitelist(self, user_id, access_token):
+    async def _check_project_id_whitelist(self, username, access_token):
         http_client = AsyncHTTPClient()
         headers = _api_headers(access_token)
         # Check if user has developer access to any project in the whitelist
         for project in self.gitlab_project_id_whitelist:
-            url = "%s/projects/%s/members/%d/all" % (GITLAB_API, project, user_id)
+            url = "%s/projects/%s/members/all?query=%s" % (GITLAB_API, project, username)
             req = HTTPRequest(url, method="GET", headers=headers)
             resp = await http_client.fetch(req, raise_error=False)
 
--- a/oauthenticator/gitlab.py
+++ b/oauthenticator/gitlab.py
@@ -134,7 +134,6 @@ class GitLabOAuthenticator(OAuthenticator):
         resp_json = json.loads(resp.body.decode('utf8', 'replace'))
 
         username = resp_json["username"]
-        user_id = resp_json["id"]
         is_admin = resp_json.get("is_admin", False)
 
         # Check if user is a member of any whitelisted groups or projects.
@@ -144,12 +143,12 @@ class GitLabOAuthenticator(OAuthenticator):
 
         if self.gitlab_group_whitelist:
             is_group_specified = True
-            user_in_group = await self._check_group_whitelist(user_id, username, access_token)
+            user_in_group = await self._check_group_whitelist(username, access_token)
 
         # We skip project_id check if user is in whitelisted group.
         if self.gitlab_project_id_whitelist and not user_in_group:
             is_project_id_specified = True
-            user_in_project = await self._check_project_id_whitelist(user_id, username, access_token)
+            user_in_project = await self._check_project_id_whitelist(username, access_token)
 
         no_config_specified = not (is_group_specified or is_project_id_specified)
 
@@ -168,7 +167,7 @@ class GitLabOAuthenticator(OAuthenticator):
             return None
 
 
-    async def _check_group_whitelist(self, user_id, username, access_token):
+    async def _check_group_whitelist(self, username, access_token):
         http_client = AsyncHTTPClient()
         headers = _api_headers(access_token)
         # Check if user is a member of any group in the whitelist
@@ -179,12 +178,12 @@ class GitLabOAuthenticator(OAuthenticator):
             if resp.code == 200 and resp.body:
                 resp_json = json.loads(resp.body.decode('utf8', 'replace'))
                 for user in resp_json:
-                    if user['id'] == user_id and user['username'] == username:
+                    if user['username'] == username:
                         return True  # user _is_ in group
         return False
 
 
-    async def _check_project_id_whitelist(self, user_id, username, access_token):
+    async def _check_project_id_whitelist(self, username, access_token):
         http_client = AsyncHTTPClient()
         headers = _api_headers(access_token)
         # Check if user has developer access to any project in the whitelist
@@ -198,7 +197,7 @@ class GitLabOAuthenticator(OAuthenticator):
                 for user in resp_json:
                     # We only allow access level Developer and above
                     # Reference: https://docs.gitlab.com/ee/api/members.html
-                    if user['id'] == user_id and user['username'] == username and user['access_level'] >= 30:
+                    if user['username'] == username and user['access_level'] >= 30:
                         return True
         return False
 
--- a/oauthenticator/gitlab.py
+++ b/oauthenticator/gitlab.py
@@ -173,7 +173,7 @@ class GitLabOAuthenticator(OAuthenticator):
         headers = _api_headers(access_token)
         # Check if user is a member of any group in the whitelist
         for group in map(url_escape, self.gitlab_group_whitelist):
-            url = "%s/groups/%s/members/%d" % (GITLAB_API, group, user_id)
+            url = "%s/groups/%s/members/%d/all" % (GITLAB_API, group, user_id)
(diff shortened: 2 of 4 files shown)


vindvaki:
See https://gitlab.com/gitlab-org/gitlab-ce/issues/66458 for added context

vindvaki:
Something seems wrong with the MediaWiki based tests. Another pull request is failing in the same way.

vindvaki:
Thanks @consideRatio! Reading the API docs again and my proposed changes, I see I was a bit hasty 😅 

I'm tempted to actually make the changes on the GitLab API instead. Nothing explicitly states that these responses should not include inherited members, and `members/all` can get really big.

What do you think?

vindvaki:
Ok, so changing the GitLab API is maybe not ideal, because there's no distinction between inherited and direct members in the response. 

Also, there's no need for that, since `/all` has an optional param called `query` that can filter by username, so we can change both of these to `/all?query=$USERNAME` instead of the `/:user_id` endpoints.

I think we need `/all` for both projects and groups, because both groups and projects can be arbitrarily nested, and users can be granted access at each level.

consideRatio:
=) no worries! Oh sure it makes sense to improve the GitLab API itself!

Are you suggesting to add a :user-id field to the /all request or modifying the existing `/groups|projects/:id/members` to include ancestor groups? I think the latter would be a breaking change due to...

> ## List all members of a group or project
> Gets a list of group or project members viewable by the authenticated user. **Returns only direct members and not inherited members through ancestors groups.**
> --- [GitLab Documentation about /groups|projects/:id/members](https://docs.gitlab.com/ee/api/members.html#list-all-members-of-a-group-or-project-including-inherited-members)

---

Ah see your new comment now, using query=$USERNAME makes sense!

oauthenticator/gitlab.py
```diff
@@ -168,25 +168,25 @@ class GitLabOAuthenticator(OAuthenticator):
...
 
 
-    async def _check_group_whitelist(self, user_id, access_token):
+    async def _check_group_whitelist(self, username, access_token):
         http_client = AsyncHTTPClient()
         headers = _api_headers(access_token)
         # Check if user is a member of any group in the whitelist
         for group in map(url_escape, self.gitlab_group_whitelist):
-            url = "%s/groups/%s/members/%d" % (GITLAB_API, group, user_id)
+            url = "%s/groups/%s/members/all?query=%s" % (GITLAB_API, group, username)
```
vindvaki: Oh, and I believe we need to encode at least `group` here and `project` below, as these these can contain `/` (e.g. the project `gitlab-org/gitlab` 😉 ) which need to be encoded as `%2F`.

vindvaki: I was just blind. Missed the `map(url_escape, ...)` part in the line above

vindvaki:
@consideRatio The GitLab tests are now passing again. Would you mind giving this another look?

vindvaki:
Ah, there's one more bug in there. `/all` returns an array, so we need to take the first result if any

vindvaki:
Hmm, and the query also does partial matches, so actually we need to filter on the results to match on the id after all

consideRatio:
@vindvaki good catch again with the partial match. I failed to find a workaround while still using the `?query=username` solution. I tried making it match exactly, but apparently it matches with wildcard as default. thats bad =/

vindvaki:
@consideRatio I think I finally have it working the way we want. Just to make sure, I set up a local jupyterhub development environment and made it use this branch of `oauthenticator`, connected it to as an oauth app to my GitLab.com account, and tested it with a test group/project hierarchy I have there:

### Project id whitelist
**criteria:** only developers and above should be accepted
- direct project membership: Tested with a direct project member and varied the privileges
- inherited project membership: Tested with a parent group owner and a non-owner with varying privileges

### Group whitelist
**criteria:** all group members should be accepted
- direct group membership: Tested with a direct member
- inherited group membership: Tested using a subgroup with a user defined on a parent group

oauthenticator/gitlab.py
```diff
@@ -144,12 +144,12 @@ class GitLabOAuthenticator(OAuthenticator):
...
         if self.gitlab_group_whitelist:
             is_group_specified = True
-            user_in_group = await self._check_group_whitelist(user_id, access_token)
+            user_in_group = await self._check_group_whitelist(user_id, username, access_token)
 
         # We skip project_id check if user is in whitelisted group.
         if self.gitlab_project_id_whitelist and not user_in_group:
             is_project_id_specified = True
-            user_in_project = await self._check_project_id_whitelist(user_id, access_token)
+            user_in_project = await self._check_project_id_whitelist(user_id, username, access_token)
```
consideRatio: Are these two conditionals for group/project whitelists ordered deliberately already? If not, consider what is most likely to reduce the amount of requests. I think it is good as it is though as I imagine groups have a bigger user base within them and the potential API rate limiting is probably relating to the amount of requests rather than how much data passed back.

oauthenticator/gitlab.py
```diff
@@ -168,36 +168,38 @@ class GitLabOAuthenticator(OAuthenticator):
...
-            url = "%s/groups/%s/members/%d" % (GITLAB_API, group, user_id)
+            url = "%s/groups/%s/members/all?query=%s" % (GITLAB_API, group, username)
             req = HTTPRequest(url, method="GET", headers=headers)
             resp = await http_client.fetch(req, raise_error=False)
-            if resp.code == 200:
-                return True  # user _is_ in group
+            if resp.code == 200 and resp.body:
+                resp_json = json.loads(resp.body.decode('utf8', 'replace'))
+                for user in resp_json:
+                    if user['id'] == user_id and user['username'] == username:
```
consideRatio: If is redundant to check both these, isn't it?

oauthenticator/gitlab.py
```diff
@@ -168,36 +168,38 @@ class GitLabOAuthenticator(OAuthenticator):
...
-
-                # We only allow access level Developer and above
-                # Reference: https://docs.gitlab.com/ee/api/members.html
-                if resp.code == 200 and access_level >= 30:
-                    return True
+                for user in resp_json:
+                    if user['id'] == user_id and user['username'] == username:
+                        # We only allow access level Developer and above
+                        # Reference: https://docs.gitlab.com/ee/api/members.html
+                        return user['access_level'] >= 30
```
consideRatio: Same question as above regarding the redundancy, and :heart: for including a good reference pointer to the access_level details!

oauthenticator/gitlab.py
vindvaki: Yes, I suppose checking just one should suffice as these are both values we get back from GitLab. I would pick the `id` for no particular reason, though `username` is probably also safe.

oauthenticator/gitlab.py
vindvaki: Good question!

It's hard to say about ideal ordering, because it largely depends on which list is longer, which seems to be entirely up to the user. And then there's considerations like that larger groups first to increase chances of an early return

For JupyterHub installed as a GitLab-managed app on a Kubernetes cluster, one of these lists will generally be empty, and the other will include exactly the project or group associated with the cluster.

In any case, I think it's safe to say neither before nor after this MR would it be recommended to have more than a handful of whitelisted groups and projects, especially given the [10 requests per second per IP address](https://docs.gitlab.com/ee/user/gitlab_com/index.html#haproxy-api-throttle) limitation on GitLab.com

vindvaki: Thinking more about deliberate ordering, I'm not sure if we should bother until we get actual user complaints. And the same for the rate limits.

oauthenticator/gitlab.py
vindvaki: To be fair, the link to the docs was already there 😉

vindvaki:
Thanks for the thorough review @consideRatio ❤️ 

---

> - Does this code risk running into pagination issues, for example we could imagine a group contains 105 members, and will fail due to pagination challenges.

While I had not considered pagination until now, I had considered the possibility of a username matching a large amount of group members. Given that 3 characters are required for an inexact match to be attempted, I did not think there would ever be *that* many matches. I tested it on the `gitlab-org` group, which is moderately big, and responses are still fast, even for usernames with 20 matches. And I had a hard time constructing a query with more matches than that.

But with pagination in the mix, *this MR is wrong as-is* and could introduce a regression for groups where a username matches a large amount of group members. Asking for 100 per page would limit the chances of needing multiple requests, but **I'm reconsidering if we should add a new pair of endpoints to GitLab that account for ancestor groups in membership checks**.

What do you think? The appeal of not using a new GitLab endpoint is primarily for backwards compatibility with older versions of GitLab. 

----

> - Could this start failing due to API limitations
> - If we fail due to API rate limiting, how would be good to fail and how would we fail right now?

Yes, it can fail, but this MR should usually not change the number of calls. The exception being when there is pagination and the match is not on the first page. The number of calls per IP over time is what counts, assuming I'm reading the docs for [EE rate limiting](https://docs.gitlab.com/ee/security/rate_limits.html) and [GitLab.com-specific rate limits](https://docs.gitlab.com/ee/user/gitlab_com/index.html#gitlabcom-specific-rate-limits) correctly.

The limit most likely to hit us seems *10 requests per second per IP address* for the API, which I think is unlikely to affect *concurrent* users of the same JupyterHub instance, but might very well affect individual users of installations with large whitelists.

oauthenticator/gitlab.py
vindvaki: Actually, one idea that is simple enough not to require additional testing: We could loop over the whitelists in a random order (different order each time). That way, at least the problem has a chance of getting better on a refresh for legitimate users.

vindvaki:
I'm in the progress of [updating the GitLab API](https://gitlab.com/gitlab-org/gitlab/merge_requests/17744), because I don't like the thought of having to paginate for something that should be a simple "is this user a member of this group/project?" call.

However, I'm struggling to make changes that are both sensible at an API level and allow upgrading of this library without immediately requiring a GitLab upgrade. See [this comment](https://gitlab.com/gitlab-org/gitlab/merge_requests/17744#note_224339866) for details.

@consideRatio do you think this is a worthwhile concern?

We could add a version [check call](https://docs.gitlab.com/ee/api/version.html) to the start of the flow to decide whether to use `/all` or not.

vindvaki:
I should note that compatibility would not be an issue for GitLab-managed JupyterHub installations; only independent JupyterHub installations relying on GitLab for authentication would be affected.

oauthenticator/gitlab.py
consideRatio: It's fine to keep both, but the fact that both were checked made me consider if i made a faulty assumption only one check was needed. That would be the main reason for me to only use one.

consideRatio:
Could you add a note that it isn't recommended to use a very large whitelist because of potential rate limits?

Regarding a check call, there are three situations as I understand it.
1. The GitLab instance has the required API
2. The GitLab instance doesn't have the required API
3. The GitLab instance doesn't have the required API, but that would mean additional calls with potential pagination problems.

I suggest making a cached (in memory) version check in order to determine if the new API is available, and if it isn't just print a warning that this new functionality coming from the /all endpoint won't be available. In other words, ignore point 3 above and only focus on 1 + 2, assuming that's easier.

---

@vindvaki thanks for doing this inter-project work so thorough! I really like having a new GitLab API endpoint to resolve the challenges you faced trying to implement this originally. I think it makes great sense to have on its own separate from this specific need as well!

vindvaki:
Thanks @consideRatio! Really appreciate the feedback ❤️ 

I just had another idea as an alternative to the API call for the version check: We can make this a configuration option which is opt-in by default for the next few of GitLab releases. This is a bit easier to code and more robust (e.g. what do we do if the version check fails? And do we block startup until version check is done?)

I will go for the configuration option unless you prefer the API call approach.

I'll ping you again once I've made these changes and the rest of your suggestions (update docs, update changelog)

consideRatio:
:sunflower: :heart:

I like the version check call because I don't see when someone would want some other logic than "use the new API if it is available, but don't if it isnt".

I'm thinking we can have a helper function that is called synchronously, but is returning a cached value if it has been called once before. If the request fails due to failure to communicate with the GitLab API, then let it crash because a failure here indicate something will fail later anyhow so why right away.

oauthenticator/gitlab.py
vindvaki: I think the note in the readme should be sufficient for now

vindvaki:
@consideRatio I've added a changelog entry, updated the readme, and switched to the new endpoints from https://gitlab.com/gitlab-org/gitlab/merge_requests/17744 (which I tested with local development instances of GitLab and JupyterHub). 

We still have to wait for the GitLab change to be merged before this one can be merged, but maybe you could give this another look in the meantime?

I had a hard time getting the version check to work on startup, so I made it a lazily memoized check instead.

oauthenticator/gitlab.py
```diff
@@ -124,6 +125,11 @@ class GitLabOAuthenticator(OAuthenticator):
 
         access_token = resp_json['access_token']
 
+        # memoize gitlab version for class lifetime
+        if self.gitlab_version is None:
+            self.gitlab_version = await self._get_gitlab_version(access_token)
+            self.member_api_variant = 'all/' if self.gitlab_version >= '12.4' else ''
```
consideRatio: I considered this string based comparison, but its probably good and robust enough to not fail.

consideRatio:
> @consideRatio I've added a changelog entry, updated the readme, and switched to the new endpoints from https://gitlab.com/gitlab-org/gitlab/merge_requests/17744 (which I tested with local development instances of GitLab and JupyterHub).

:heart:

> We still have to wait for the GitLab change to be merged before this one can be merged, but maybe you could give this another look in the meantime?

I don't see why we need to wait since you do the check for another version.

> I had a hard time getting the version check to work on startup, so I made it a lazily memoized check instead.

I think that's fine!

---

I'll review the changes very soon, is there something left you want to do with this PR as far as you know?

oauthenticator/gitlab.py
vindvaki: Ah, I was too fixated on `11` vs `12` that I forgot about cases like `'9.5' > '12.4'`. And if the code lives for long enough, then we'd also run into `'111.0' < '12.4'` 😉 

The last problematic version, `9.5`, came out in August 2017, and `10.0` one month later. Neither version is still supported, and it's an unlikely combination to have such an old version in use for JupyterHub authentication. 

We can parse the output of `/api/v4/version` to be on the safe side

vindvaki:
> I don't see why we need to wait since you do the check for another version.

It's just a slight possibility the change won't be included in 12.4.0, in which case we'd need to bump this to whichever version of GitLab first includes the new endpoints (e.g. 12.5 or 12.4.1)

> is there something left you want to do with this PR as far as you know?

Yeah I'll fix the version parsing as mentioned above

oauthenticator/gitlab.py
vindvaki: Updated the version check to parse to a list instead

oauthenticator/gitlab.py
```diff
@@ -167,13 +173,24 @@ class GitLabOAuthenticator(OAuthenticator):
...
+    async def _get_gitlab_version(self, access_token):
+        url = '%s/version' % GITLAB_API
+        req = HTTPRequest(url,
+                          method="GET",
+                          headers=_api_headers(access_token),
+                          validate_cert=self.validate_server_cert)
+        resp = await AsyncHTTPClient().fetch(req, raise_error=True)
+        resp_json = json.loads(resp.body.decode('utf8', 'replace'))
+        version_string = re.sub(r'-pre$', '', resp_json['version'])
+        return list(map(int, str.split(version_string, '.')))
```
consideRatio: What do you think about skipping the use of a regular expression here, and instead do something like this?

I know that my gitlab version returned was `12.2.5-ee`, so it makes sense to not lock into specifically stripping `-pre` or similar but instead truncating at anything after `-` that separates the `major.minor.patch` part of the string from the other details.

```python
        version_strings = resp_json["version"].split("-")[0].split(".")[:3]
        version_ints = list(map(int, version_strings))
        return version_ints
```

oauthenticator/tests/test_gitlab.py
```diff
@@ -37,9 +37,18 @@ def gitlab_client(client):
...
+        ret = { 'version': version, 'revision': "f79c1794977" }
+        return HTTPResponse(request, 200,
+                            headers={'Content-Type': 'application/json'},
+                            buffer=BytesIO(json.dumps(ret).encode('utf-8')))
+    regex = re.compile(API_ENDPOINT + '/version')
+    client.hosts['gitlab.com'].append((regex, mock_version_response))
 
 async def test_gitlab(gitlab_client):
     authenticator = GitLabOAuthenticator()
+    mock_api_version(gitlab_client, '12.3')
```
consideRatio: Mock with a realistic version response, such as `12.2.5-ee`.

This was what my /api/v4/version endpoint returned for example.
```json
{"version":"12.2.5-ee","revision":"e817f2e2"}
```

oauthenticator/tests/test_gitlab.py
```diff
@@ -58,6 +67,7 @@ def make_link_header(urlinfo, page):
 async def test_group_whitelist(gitlab_client):
     client = gitlab_client
     authenticator = GitLabOAuthenticator()
+    mock_api_version(client, '12.4')
```
consideRatio: Same as above

oauthenticator/gitlab.py
vindvaki: Great catch!

README.md
```diff
@@ -181,6 +181,16 @@ You can also use `LocalGitLabOAuthenticator` to map GitLab accounts onto local u
...
 
+You can restrict access to only accept members of certain projects or groups by setting
+```
+c.GitLabOAuthenticator.gitlab_project_id_whitelist = [ ... ]
+```
+and
+```
+c.GitLabOAuthenticator.gitlab_group_whitelist = [ ... ]
+```
+but be aware that each entry incurs a separate API call, increasing the risk of rate limiting and timeouts.
```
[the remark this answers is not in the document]
vindvaki: I was thinking as a separate concern, as if there are a lot of API calls during the authorization check, then we could end up with a client-side timeout. Does the text need additional clarification?

oauthenticator/gitlab.py
vindvaki: Fixed

oauthenticator/tests/test_gitlab.py
vindvaki: Thanks! Updated

vindvaki:
Looks like the changes made it to 12.4 on the GitLab side, so we're good on that end

consideRatio:
LGTM! Thank you for your thorough work on this @vindvaki!

vindvaki:
Thanks @consideRatio! Now we need to get this to [zero-to-jupyterhub-k8s
](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/6290e23602225b543850f5567c77ddd6c3662297/images/hub/requirements.txt#L10)

Any thoughts on cutting a new release of each?

vindvaki:
I would be happy to assist with those releases in any way I can

PR #287, opened 2019-09-04
[GenericOAuthenticator] Implement refresh_user

guimou:
This PR implements the refresh_user function for the GenericOauthenticator, as defined in https://github.com/jupyterhub/jupyterhub/blob/e4d4e059bd6ecc749a6276a80eada8f0de8ad206/jupyterhub/auth.py#L429

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
 jupyterhub>=0.5
-pyjwt==1.7.1
+pyjwt
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,2 @@
 jupyterhub>=0.5
+pyjwt==1.7.1
--- a/oauthenticator/generic.py
+++ b/oauthenticator/generic.py
@@ -7,6 +7,8 @@ import json
 import os
 import base64
 import urllib
+import time
+import jwt
 
 from tornado.auth import OAuth2Mixin
 from tornado import web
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -1 +1,2 @@
-jupyterhub>=0.5
\ No newline at end of file
+jupyterhub>=0.5
+pyjwt==1.7.1
```
consideRatio: We should not pin on the latest release only, what about pyjwt>=1.7.1 ?

```suggestion
pyjwt>=1.7.1
```

consideRatio: Also, remove pyjwt from test-requirements.txt is needed if this is added in requirements.txt

consideRatio:
## Review help
We have currently not added dependency parts for various authenticators, but the GenericOAuthenticator is subclassed by other authenticators and should really be allowed to have a dependency at least - right?

I know AzureAdB2COAuthenticator added notes about needing pyjwt recently, and could be updated to not declare that needed if we add this pyjwt dependency.

manics:
https://github.com/jpadilla/pyjwt/blob/1.7.1/setup.py
Looks like there are no compulsory dependencies, and it's a pure Python package (easy to pip install), so if this is all that's required I think it's fine to add as a dependency.

The [optional dependency](https://github.com/jpadilla/pyjwt/blob/1.7.1/setup.py#L75) adds cryptographic support.

requirements.txt
guimou: I don't see why it could not be changed. During the project we were pinning most of the packages to guarantee some reproducibility, but we're not using some very specific function here, just decoding the JWT.

consideRatio: It is fine to pin packages, but we should not make oauthenticator as a package require specifically 1.7.1, then you wont be able to install both oauthenticator and a package that require a version of 1.7.2 or later.

guimou:
Requested changes are done!

consideRatio:
Thanks @guimou! I merged a PR and there is now a merge conflict, can you resolve it?

minrk:
Sorry for not reviewing promptly. This makes an assumption that access tokens and refresh tokens are jwt, which is not generally true (quite rare in fact, with providers I have encountered). Unfortunately, GenericOAuthenticator truly does need to be generic, and thus I think getting some of this info using jwt methods needs to be abstracted away via configuration to support the apparent majority of providers where tokens are not jwt.

guimou:
@minrk That makes a lot of sense. Thanks for the review though. I guess we may close the PR then, the code would have to go a Keycloak-specific authenticator. Maybe I'll do it if I find the time.

manics:
As suggested in https://github.com/jupyterhub/oauthenticator/pull/287#issuecomment-568076150 I'm closing this PR.

sidhantswami:
Hello, I had quick question with your implementation for refresh_user. How were you able to decode your refresh token and get the expiry? I am currently using the GenericAuth template as well and using Okta as the authentication server.

guimou:
@sidhantswami Woo, that's an old thing :smile: 
Quickly looking back at the code what you're looking for seems to be on line 196 for file `oauthenticator/generic.py` from this PR.
I found out that it has also be implemented in the same way but much more properly on this other project: [ref1](https://github.com/swan-cern/jupyterhub-extensions/blob/88b512202ffa6ff9e398891ab33e24b4bb7c4a1f/KeyCloakAuthenticator/keycloakauthenticator/auth.py#L271) and [ref2](https://github.com/swan-cern/jupyterhub-extensions/blob/88b512202ffa6ff9e398891ab33e24b4bb7c4a1f/KeyCloakAuthenticator/keycloakauthenticator/auth.py#L173)
I hope it was what you were looking for.

Issue #285, opened 2019-08-27
Globus logout raises exception


NickolausDS:
On a fresh install of Jupyterhub and Oauthenticator, Logout on Globus raises the following exception: 
```
/home/vagrant/anaconda3/envs/jupyterhub/lib/python3.7/site-packages/oauthenticator/globus.py:51: RuntimeWarning: coroutine 'LogoutHandler.get' was never awaited
  super().get()
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
/home/vagrant/anaconda3/envs/jupyterhub/lib/python3.7/site-packages/tornado/web.py:1699: RuntimeWarning: coroutine 'BaseHandler.get_current_user' was never awaited
  result = await result
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

This was using: 
* `oauthenticator==0.9.0`
* `jupyterhub==1.0.0`

Comparing the [Jupyterhub module](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/handlers/login.py#L80) with the [Globus Module](https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/globus.py#L51), it looks like a get call is missing an `await`. Any thoughts?

NickolausDS:
This was fixed last month with #288. Thanks everyone!

PR #288, opened 2019-09-11
Fixed Globus Logout Handler, added test

NickolausDS:
Fix for #285. Previously, logout would not `await` the `super().get()` line, causing the exception. The result was logout succeeding but failing to render the page for users. 

The simple fix was simply to `await super().get()`, but I also wanted to be more respectful of the new `handle_logout` for the custom token revocation behavior. However, I don't think it's possible to completely remove `get()` due to the `self.redirect()` line raising an exception if called after the template rendering code in the superclass.

In addition to these changes I also fixed up the tests and added a new unrelated test to increase coverage. I can split that out into a separate PR if that would be preferred. 

Tested on Jupyterhub and Zero to Jupyterhub.

### DIFF
--- a/oauthenticator/tests/test_globus.py
+++ b/oauthenticator/tests/test_globus.py
@@ -60,12 +60,10 @@ def mock_globus_user(mock_globus_sdk):
         name = 'Wash'
         state = {'tokens': mock_globus_sdk.by_resource_server}
 
-        @gen.coroutine
-        def get_auth_state(self):
+        async def get_auth_state(self):
             return self.state
 
-        @gen.coroutine
-        def save_auth_state(self, state):
+        async def save_auth_state(self, state):
             self.state = state
     return User()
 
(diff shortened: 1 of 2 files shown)


oauthenticator/tests/test_globus.py
```diff
@@ -54,6 +54,22 @@ def globus_client(client):
...
+    class User:
+        name = 'Wash'
+        state = {'tokens': mock_globus_sdk.by_resource_server}
+
+        @gen.coroutine
+        def get_auth_state(self):
+            return self.state
+
+        @gen.coroutine
+        def save_auth_state(self, state):
```
consideRatio: Can this perhaps be async instead of @gen.coroutine functions?

NickolausDS: Absolutely, I'll update that now.

NickolausDS:
@consideRatio I added your suggestion, and also rebased to be up-to-date with the latest oauthenticator master branch.

minrk:
Excellent, thanks!

PR #290, opened 2019-09-16
Update generic.py

hmumtaz:
Fixes TypeError: get_argument() missing 1 required positional argument: 'name'

### DIFF
--- a/oauthenticator/generic.py
+++ b/oauthenticator/generic.py
@@ -86,7 +86,7 @@ class GenericOAuthenticator(OAuthenticator):
     )
 
     async def authenticate(self, handler, data=None):
-        code = handler.get_argument("code")
+        code = handler.get_argument(name="code")
         # TODO: Configure the curl_httpclient for tornado
         http_client = AsyncHTTPClient()
 


consideRatio:
What object type is the handler parameter? From the JupyterHub defined base class's docstring in the authenticate() function, i see that handler is a `tornado.web.RequestHandler` (source: [auth.py](https://github.com/jupyterhub/jupyterhub/blob/9990100f896afbd80526e020905bec0a746f0a24/jupyterhub/auth.py#L474-L490))

So I traced down handler.get_argument function definition in both [tornado 6.0](https://www.tornadoweb.org/en/branch6.0/web.html#tornado.web.RequestHandler.get_argument) and [tornado 5.1](https://www.tornadoweb.org/en/branch5.1/web.html#tornado.web.RequestHandler.get_argument) and concludes that tornado 6.0 made a change in the signature that indeed may require us to name the argument passed like this PR does.

This LGTM!

minrk:
Something's funky here, because this is already a positional argument and works with tornado 5 and 6 as-is. I'm guessing something else peculiar has happened in this particular environment, e.g. if `handler` were a handler *class* instead of the handler instance, this error would occur, because the first positional argument will be consumed as `self`.

consideRatio:
Closing for now, feel free to reopen!

PR #297, opened 2019-10-28
mediawiki: utf-8 > binary strings, req. mwoauth>=0.3.7

consideRatio:
Awaits merge and release of https://github.com/mediawiki-utilities/python-mwoauth/pull/39, and would very much benefit review. I've only tried to resolve the test failures but don't have a clue about auth with mediawiki in reality.

@yuvipanda, oauthenticator is broken with mediawiki 0.3.5 and 0.3.6, but may be fixed with the referened PR in 0.3.7. I verified that that PR + this PR combined made tests succeed, but I don't properly grasp what I did except that remove all decode/encode stuff going on, as it seems that all interaction with mwoauth from now can be done without binary strings.


chicocvenancio:
LGTM.

Issue #299, opened 2019-11-05
Make a release


consideRatio:
I'd like to make a release of oauthenticator. My motivation is mainly to do it alongside a z2jh release I'd like to make. Ping @jupyterhub/jupyterhubteam!

Related instructions on how to do it are drafted in #294.

manics:
Can we use this as an opportunity to automate the deployment? https://github.com/jupyterhub/team-compass/issues/213#issuecomment-548124874

I _think_ this repo should be straightfoward to convert since it's a pure python package.

choldgraf:
I'm all for it

willingc:
Before making the release, are there any open PRs that should go into the release? cc/ @minrk

minrk:
Looking through PRs, I think we can make a release from master as it is now. There are always future releases.

I'd be happy with following the instructions in #294, or attempting to implement the automated release process. One tool that avoids the objections to versioneer is [bumpversion](https://pypi.org/project/bumpversion/) (or [punch](https://punch.readthedocs.io), but I haven't tried punch before). I've used this in various projects, and it's pretty nice. Versioneer is the only game in town if we want to be able to publish releases entirely from the GitHub website (i.e. create the tag and that's it), but it has its downsides (git archive doesn't work, install from forks can report the wrong version, etc.). With bumpversion, it can at least be entirely scriptable, but does require someone with push access to run some code.

The release process  with bumpversion (after initial configuration) looks like:

```bash
bumpversion --tag release  # creates commit with 1.2.3-dev -> 1.2.3 and tag 1.2.3
python3 setup.py sdist bdist_wheel
twine upload dist/*
bumpversion --no-tag patch # creates commit with 1.2.3 -> 1.2.4-dev
git push && git push --tags 
```

This automates only the editing of the version file and creation of tags, but at least enables everything to be fully automatic without editing any files. Push  to PyPI from the tag can still be automatic.

manics:
This is automated 😀
https://github.com/jupyterhub/oauthenticator/blob/397a3ec6b489635ceb70cfd40059d8ee55bddd6d/.travis.yml#L37-L48

PR #294, opened 2019-10-25
Add RELEASE.md

consideRatio:
Mostly copied from https://github.com/jupyterhub/chartpress, but slightly adjusted relating to detailed version update instructions.

PR #301, opened 2019-11-10
[MRG] Travis: Deploy releases to pypi

manics:
Release tagged builds to PyPi using Travis.

A matrix build will normally deploy multiple times. The usual way of deploying once is to use either [`skip_existing: true`](https://docs.travis-ci.com/user/deployment/pypi/#upload-artifacts-only-once) or a conditional that matches one of the jobs. The downside is that the deploy only depends on whether that particular job passes, other jobs in the matrix may fail.

I think using build stages is better because the deploy only happens if all matrix jobs pass. It's also clearer in travis since the deploy runs as it's own job.

I've tested this on testpypi:
- https://github.com/manics/oauthenticator/commit/aa53d6d1408d4881b499b48ab23c9869df6b2b6e
- https://test.pypi.org/project/oauthenticator-testpypi/0.0.1.dev3/
-  https://travis-ci.com/manics/oauthenticator/builds/135826346
manics-oauthenticator-testpypi-travis

Next steps:
- [ ] Either add a secret environment variable `PYPI_TOKEN` to Travis or add an encrypted secret to travis.yml

Related:
- https://github.com/jupyterhub/oauthenticator/issues/299
- https://github.com/jupyterhub/oauthenticator/pull/294 (I'll help update this if we decide to go ahead with the automatic deploy for this release)
- https://github.com/jupyterhub/team-compass/issues/213#issuecomment-548124874

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,14 +21,17 @@ after_success:
   - codecov
 
 jobs:
+  allow_failures:
+    - python: nightly
+  fast_finish: true
   include:
     # Default stage: test
     - python: 3.7
       env: TEST_LINT=1
-    - python: nightly
     - python: 3.7
     - python: 3.6
     - python: 3.5
+    - python: nightly
     # Only deploy if all test jobs passed
     - stage: deploy
       python: 3.7
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,4 +39,4 @@ jobs:
       deploy:
         provider: pypi
         user: __token__
-        password: $PYPI_TOKEN
+        # password: see secret PYPI_PASSWORD variable
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,7 @@ language: python
 sudo: false
 cache: pip
 dist: xenial
-python:
-  - nightly
-  - 3.7
-  - 3.6
-  - 3.5
+
 install:
   - pip install --upgrade pip
   - pip install --upgrade --pre -r test-requirements.txt .
@@ -24,7 +20,22 @@ script:
 after_success:
   - codecov
 
-matrix:
+jobs:
   include:
+    # Default stage: test
     - python: 3.7
       env: TEST_LINT=1
+    - python: nightly
+    - python: 3.7
+    - python: 3.6
+    - python: 3.5
+    # Only deploy if all test jobs passed
+    - stage: deploy
+      python: 3.7
+      if: tag IS present
+      deploy:
+        provider: pypi
+        user: __token__
+        password: $PYPI_TOKEN
+        on:
+          tags: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,5 +37,3 @@ jobs:
         provider: pypi
         user: __token__
         password: $PYPI_TOKEN
-        on:
-          tags: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,3 +40,4 @@ jobs:
         provider: pypi
         user: __token__
         # password: see secret PYPI_PASSWORD variable
+        distributions: sdist bdist_wheel


betatim:
Looks good to me!

.travis.yml
```diff
@@ -24,7 +20,22 @@ script:
...
+    # Only deploy if all test jobs passed
+    - stage: deploy
+      python: 3.7
+      if: tag IS present
+      deploy:
+        provider: pypi
+        user: __token__
+        password: $PYPI_TOKEN
+        on:
+          tags: true
```
consideRatio: I don't think it is relevant to have the on: tags: true section assuming we have `if: tag IS present` already. TravisCI is very hard to comprehend though so I'm certainly not certain.

manics:
@consideRatio How does this look?

consideRatio:
@manics this LGTM!

manics:
@consideRatio No! Someone with access tothe PyPi project needs to add the token either in the Travis web ui or by adding a secret to `.travis.yml`. We can either wait, or merge and sort out the token separately since it only takes effect on a tag. I tested it on testpypi but probably best to test again with an RC just in case.

consideRatio:
@manics i was a maintainer but not an owner so I cannot create such token. What path do you recommend to take regarding adding it to the web ui of travis or adding it as a encrypted secret in .travis.yml ?

@choldgraf or @minrk can perhaps help out with permissions

choldgraf:
@consideRatio you are now an owner!

manics:
@consideRatio 
Creating the token:
Goto https://pypi.org/manage/project/{pypi-repo}/settings/
`API Token` → `Create a token for jupyter-notebookparams`, Name can be anything, Scope: just the repository for now

Either add in Travis web UI:
Go to `https://travis-ci.com/{organisation}/{repo}/settings` and scroll down to ` Environment Variables`. Add `PYPI_TOKEN`:
Travis Environtment Variables

Or use an encrypted secret inside `.travis.yml
https://docs.travis-ci.com/user/environment-variables#defining-encrypted-variables-in-travisyml

There's an ongoing discussion on https://github.com/jupyterhub/team-compass/issues/213#issuecomment-548124874 about pypi accounts and token scopes, but if you're ready to release might as well use your own token and update later?

consideRatio:
Important insights:
1. Tag builds, which are triggered by a refs/tags/<tag> object in git is pushed, does not belong to a branch. That makes sense, as a tag is a tagged commit, but this can be part of multiple branches or none at all.
2. If we limit an env var to to a certain branch, that means we exclude tagged builds as they are not part of any branch. It is not possible in the TravisCI web settings to declare if they should be part of tagged builds but not branch builds.
3. With `travis encrypt` we could get a nice entry in TravisCI relating to an encrypted environment variable that would allow us to specifically make the PYPI_PASSWORD environment available when it is needed, on tagged builds. But, we can't use it due to [a limiting behavior we experience](https://travis-ci.community/t/pypi-upload-with-api-token-not-working/5338/5).

Anyhow, we now have a PYPI_PASSWORD setup, which is making the variable accessible on all pushes to the git repo. It is tied to a deployment token specifically for jupyterhub/oauthenticator. Note that I opted for PYPI_PASSWORD over PYPI_TOKEN as this documentation suggest that: https://docs.travis-ci.com/user/deployment-v2/providers/pypi. You can therefore remove the password: $PYPI_TOKEN fields and instead document that the PYPI_PASSWORD will be set indirectly by the PYPI_PASSWORD environment variable.

It sais that the default distribution is sdist, perhaps it should be bdist_wheel or similar also? I'm super noob about these matters really but I think it is very relevant that we inspect the varioable options for PyPI deployment here: https://docs.travis-ci.com/user/deployment-v2/providers/pypi#known-options

manics:
@consideRatio Done!

consideRatio:
@manics :heart:

PR #303, opened 2019-11-18
Add scopes documentation to auth0 example

jbradenbrown:
Had a lot of trouble figuring out why Auth0 would break if there were any cookies. Without editing scopes, the response body of /userinfo would be empty. This fixed my problem.


yuvipanda:
Thanks for the PR, @jbradenbrown!

PR #305, opened 2019-11-19
Support callable for GenericOAuthenticator username_key

eslavich:
We have an OAuth provider that returns the username in a nested object.  This PR updates GenericOAuthenticator's username_key traitlet to support a callable that can be used to extract a username from anywhere in the response.

### DIFF
--- a/oauthenticator/traitlets.py
+++ b/oauthenticator/traitlets.py
@@ -1,0 +1,16 @@
+from traitlets import TraitType
+
+class Callable(TraitType):
+    """
+    A trait which is callable.
+    Classes are callable, as are instances
+    with a __call__() method.
+    """
+
+    info_text = 'a callable'
+
+    def validate(self, obj, value):
+        if callable(value):
+            return value
+        else:
+            self.error(obj, value)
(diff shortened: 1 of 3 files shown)


oauthenticator/traitlets.py
```diff
@@ -0,0 +1,16 @@
+from traitlets import TraitType
```
yuvipanda: We depend on JupyterHub, so you might be able to just import https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/traitlets.py#L93

eslavich: Ah!  I didn't notice that, thanks.

The requirements.txt allows jupyterhub 0.5, which doesn't include Callable -- Callable first shows up in 0.9.0.  Is it okay to stop supporting jupyterhub < 0.9.0?

yuvipanda: Oh, that's interesting. I *think* we passed that bridge a long time ago...

Regardless, let's leave it as is and pick that up in a separate PR.

yuvipanda:
Thanks, @eslavich!

PR #307, opened 2019-11-21
Add AzureAdB2COAuthenticator (AzureAdB2COAuthenticator)

linkcd:
Hi 
I would like to adding support for Azure AD B2C (https://azure.microsoft.com/en-us/services/active-directory-b2c/). 

# Problem
The current Azure AD works fine by providing a simple tenant ID, then all OAuth endpoint was generated based on the tenant ID as these AAD endpoints are in the same format.

However,  config Azure AD B2C requires more configuration than a simple tenant ID: You will need to provide policy (ie. the p parameter in the query string in OAUTH_AUTHORIZE_URL and OAUTH_ACCESS_TOKEN_URL). 

# Possible solutions
1. Use generic implementation: 
Not possible, as it requires JWT , and also there is no userdata_url for it out of box.

2. Modify the current AAD for supporting both AAD and AAD B2C
Possible, as they are sharing lots of similar logic. However, most likely it will be a breaking change for current AAD users. 

3. Create a brand new plugin for AAD B2C specific
Created a new implementation that is specifically for AAD B2C. It fully supports all different configurations that you can have in AAD B2C, such as customized policy. It will be easier to read and maintain. After all, MS have different roadmap for AAD and AAD B2C as they are two separated products. 

# Solution
Created this PR as solution 3 from above. Added new files as the implementation, sample code. Updated readme for documentation.  Tested and it works successfully. 

thank you
Feng


README.md
```diff
@@ -148,6 +150,16 @@ sudo jupyterhub -f ./path/to/jupyterhub_config.py
...
 
+#### For Azure AD B2C
+
+> See `run.sh` for an [example](./examples/azureadb2c/)
+
+* [Source Code](oauthenticator/azureadb2c.py)
+
+
+
+##
```
consideRatio:
```suggestion
```

README.md
```diff
@@ -148,6 +150,16 @@ sudo jupyterhub -f ./path/to/jupyterhub_config.py
...
+#### For Azure AD B2C
+
+> See `run.sh` for an [example](./examples/azureadb2c/)
+
+* [Source Code](oauthenticator/azureadb2c.py)
+
+
+
+## 
+
```
consideRatio: Just some cleanup here and I think we are good!

linkcd: thanks for the review. Now i have updated the readme as well as the sample config file for better documentation. If everything looks good, can you (or someone else) merge the PR, so we can use the updated package in our project? thanks!

linkcd: Also the new PR included the latest update from master branch, just fyi.

consideRatio:
@linkcd nice thanks for your work on this!

I created a new issue to represent the fact that this code isn't tested anywhere in the CI system yet. If you would want to help out with that, it would be big!

linkcd:
@consideRatio thanks for review and merge this PR. I have commented in the issue. 

Meanwhile, when can we expect the pip package https://pypi.org/project/oauthenticator/ get updated, in order to include the AAD B2C implementation? 

It would be great to get the pip package updated, so we can continue the work here :)

many thanks,
Feng

consideRatio:
Begore next week is my aim

PR #323, opened 2019-12-18
Remove mixins, per-Authenticator LoginHandler classes

### LINKED ISSUE #286: Entrypoints causes issues with JupyterHub.
A number of the authenticators require environment variables to be set up in order to be used. For example `openshift`:

```
os.environ['OPENSHIFT_URL'] = oauth_issuer_address

os.environ['OPENSHIFT_REST_API_URL'] = kubernetes_server_url
os.environ['OPENSHIFT_AUTH_API_URL'] = oauth_issuer_address

from oauthenticator.openshift import OpenShiftOAuthenticator
c.JupyterHub.authenticator_class = OpenShiftOAuthenticator
```

but also `generic` OAuth authenticator. Even things like GitHub authenticator rely on environment variables for service host.

The setting of the environment variables used to be able to be done in the `jupyterhub_config.py` file by setting them via `os.environ` and then importing the authenticator.

This pattern no longer works because the addition of the entrypoints for the authenticators results in the modules for the authenticators seem to be preloaded prior to the `jupyterhub_config.py` file being processed.

Change where entry points was added is:

* https://github.com/jupyterhub/oauthenticator/commit/5a6e94352e5c3940ee0e55594e8146f6b0985cb8

The only workaround at the moment seems to be to use:

```
del sys.modules['oauthenticator.openshift']

from oauthenticator.openshift import OpenShiftOAuthenticator
c.JupyterHub.authenticator_class = OpenShiftOAuthenticator
```

That is, delete the imported module from `sys.modules` and thus force reloading of it.

Long term, all the authenticator modules that need environment variables for some things, if you want to be able to configure them from `jupyterhub_config.py` will need to support configuration using traitlets. Whether that is possible is all cases isn't clear though.

The only other choice seems to be to set the environment variables before JupyterHub is started, which isn't necessarily convenient when `jupyterhub_config.py` was the only place provided to configure anything.

Is there perhaps a way to tell JupyterHub not to preload the modules for all the authenticators?

### LINKED ISSUE #312: Refactor away some mixins
I tried to improve some logic in https://github.com/jupyterhub/oauthenticator/pull/251 that I had found problematic from a developer experience view, but minrk came with a good approach that I want to switch to instead. So I'm closing that PR and opening this issue to not forget about that potential work.

https://github.com/jupyterhub/oauthenticator/pull/251#issuecomment-475230267

## Related
- https://github.com/jupyterhub/oauthenticator/issues/250

minrk:
This PR simplifies oauthenticator quite a bit by reducing a lot of the boilerplate involved in defining/maintaining implementations.

- [x] authorize_url, token_url, userdata_url are traits on base OAuthenticator (closes #312)
- [x] eliminates all env-only module-level config, moving everything to Authenticator traits (closes #286)
- [x] removes never-released Authenticators that could be implemented via configuration of existing ones (AzureAd or Generic)

Implementation progress:

- [x] github
- [x] gitlab
- [x] google
- [x] bitbucket
- [x] azuread
- [x] azureadb2c (never released, removed and replaced with azuread config)
- [x] auth0
- [x] aws (never released, removed and replaced with generic config)
- [x] cilogon
- [x] generic
- [x] globus
- [x] mediawiki
- [x] okpy
- [x] openshift
- [x] yandex (never released, removed and replaced with generic config)

In the future, I think we should perhaps promote more aspects of GenericOAuthenticator to the base class, and implement most of our providers via override of more granular methods than `authenticate`. I'm pretty sure several can be pretty basic already (openshift, okpy, auth0, bitbucket, google).


oauthenticator/oauth2.py
```diff
@@ -268,7 +302,7 @@ def login_url(self, base_url):
 
     login_handler = "Specify login handler class in subclass"
```
NickolausDS: I saw these changes removed custom login handlers for some classes. In that case, they will default to the `login_handler` value defined here. Should this instead fall back to `OAuthLoginHandler` if the subclass doesn't define a custom Login handler?

minrk: > Should this instead fall back to OAuthLoginHandler if the subclass doesn't define a custom Login handler?

Yes, absolutely!

oauthenticator/oauth2.py
```diff
@@ -223,10 +223,11 @@ class OAuthenticator(Authenticator):
     Subclasses must override:
 
     login_service (string identifying the service provider)
-    login_handler (likely a subclass of OAuthLoginHandler)
     authenticate (method takes one arg - the request handler handling the oauth callback)
     """
 
+    login_handler = OAuthLoginHandler
+
```
NickolausDS: Another line further down redefines `login_handler`,  nullifying this one. After removing that line, I did a quick test (using Globus) and everything looked good.

oauthenticator/oauth2.py
```diff
@@ -268,7 +303,7 @@ def login_url(self, base_url):
 
     login_handler = "Specify login handler class in subclass"
```
consideRatio: As noted by @NickolausDS this overrides the earlier statement setting the login_handler.

oauthenticator/oauth2.py
minrk: Good catch, fixed!

NickolausDS:
I went through and tested the latest changes with both GitHub and Globus, and both worked great. It looks good to me!

consideRatio:
@minrk this LGTM to the extent i can evaluate it, and i think it is an essential PR.

I'll go ahead and merge, thank you for review and testing @nickolausds!

PR #328, opened 2020-01-22
[AzureAD] Don't pass resource when requesting a token

craigminihan:
* `AzureAdOAuthenticator` fails when calling into Azure AD when the `resource` field is present in the token body.
* Updated the example config.py to show how `username_claim` can be used when the `name` returned by AD isn't compliant with JH.
* Minor typo change


oauthenticator/azuread.py
```diff
@@ -77,7 +76,7 @@ def _username_claim_default(self):
 
         headers = {
             'Content-Type':
-            'application/x-www-form-urlencoded; ; charset=UTF-8"'
```
craigminihan: This looks malformed so updated

oauthenticator/azuread.py
```diff
@@ -67,7 +67,6 @@ def _username_claim_default(self):
             client_secret=self.client_secret,
             grant_type='authorization_code',
             code=code,
-            resource=self.client_id,
```
craigminihan: Azure responds with 400 if this field is present

oauthenticator/azuread.py
```diff
@@ -19,13 +19,13 @@
 from .oauth2 import OAuthLoginHandler, OAuthenticator
 
 
-def azure_token_url_for(tentant):
```
craigminihan: super minor spelling issue

examples/azuread/sample_jupyter_config.py
```diff
@@ -9,3 +9,8 @@
 c.AzureAdOAuthenticator.oauth_callback_url = 'http://{your-domain}/hub/oauth_callback'
 c.AzureAdOAuthenticator.client_id = '{AAD-APP-CLIENT-ID}'
 c.AzureAdOAuthenticator.client_secret = '{AAD-APP-CLIENT-SECRET}'
+
+# if the user's name returned by Azure isn't acceptable to JupyterHub then we can use an alternate field,
+# uncomment the line below to use 'unique_name' rather than the default 'name'. Consult the Azure
+# documentation for other field names.
+#c.AzureAdOAuthenticator.username_claim = 'unique_name'
```
craigminihan: For me the `name` returned by AD is bounced by JH so I had to use another value. This field was already present in the class, setting its value to `unique_name` solved my problem. I added this to the example code for others to more easily find than me.

See: https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/auth.py#L185

craigminihan:
I created a custom image based on `jupyterhub/k8s-hub:0.8.2` with the `azurerd.py` from this PR. I have been able to log into JH via the OAuth redirect to a corporate sync'd AD instance in Azure.

The Dockerfile is as follows:
```
FROM jupyterhub/k8s-hub:0.8.2

USER root
RUN apt-get update && \
    apt-get install -y wget

USER jovyan

RUN pip3 install --user --no-deps --no-compile oauthenticator==0.8.0 && \
    wget https://raw.githubusercontent.com/craigminihan/oauthenticator/remove_resource/oauthenticator/azuread.py -O ~/.local/lib/python3.6/site-packages/oauthenticator/azuread.py
```
The updated `azurerd` is loaded by JH from `.local` here.

minrk:
Thanks!

PR #330, opened 2020-01-30
changelog for 0.11

minrk:
/cc @consideRatio

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -8,6 +8,41 @@ command line for details.
 
 ## [Unreleased]
 
+## 0.11
+
+## [0.11.0] - 2020-01-30
+
+The main change in 0.11 is a refactoring of classes to remove mixins,
+reducing the amount of boilerplate needed.
+In addition, there are some fixes to the Azure AD Authenticator.
+
+This release also *removes* the latest Authenticators added in 0.10
+(AzureAdB2COAuthenticator, AWSCognitoAuthenticator),
+which were released without being fully supported and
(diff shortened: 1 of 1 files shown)


manics:
Are there any breaking changes in https://github.com/jupyterhub/oauthenticator/pull/323 that implementers of external Oauthenticator sub-classes need to be aware of?

docs/source/changelog.md
```diff
@@ -8,6 +8,41 @@ command line for details.
...
+## 0.11
+
+## [0.11.0] - 2020-01-30
+
+The main change in 0.11 is a refactoring of classes to remove mixins,
+reducing the amount of boilerplate needed.
+In addition, there are some fixes to the Azure AD Authenticator.
+
+This release also *removes* the latest Authenticators added in 0.10
+(AzureAdB2COAuthenticator, AWSCognitoAuthenticator),
```
consideRatio: I confirmed that YandexPassportOAuthenticator was also part of 0.10.0 and removed.

minrk:
> Are there any breaking changes in #323 that implementers of external OAuthenticator sub-classes need to be aware of?

There aren't any known, intentional breakages here. Subclasses should be fine. One exception, perhaps if any subclasses were importing the removed mixins.

minrk:
I added a note about subclasses possibly importing the removed mixins and the fact that env-only config should be gone.

minrk:
@consideRatio I think so, yes!

consideRatio:
I pushed a 0.11.0 git tag!

PR #331, opened 2020-01-30
hosted_domain can contain domains not hosted with Google but have accounts

cmseal:
Very specific use case:

A domain, i.e. business.com, uses self-hosted email... but some users have created Google accounts using @business.com email addresses.

These accounts don't have the hd key in bodyjs[] dict from Google oauth2, but are valid Google accounts and authenticate correctly using the domain name alone.

This allows such Google accounts to auth alongside domains that may be hosted with Google.

### DIFF
--- a/oauthenticator/google.py
+++ b/oauthenticator/google.py
@@ -115,7 +115,7 @@ class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
         if self.hosted_domain:
             if (
                 user_email_domain not in self.hosted_domain
-                or bodyjs['hd'] not in self.hosted_domain
+                or ('hd' in bodyjs and bodyjs['hd'] not in self.hosted_domain)
             ):
                 self.log.warning(
                     "Google OAuth unauthorized domain attempt: %s", user_email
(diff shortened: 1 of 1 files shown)


oauthenticator/google.py
```diff
@@ -115,7 +115,7 @@ def _cast_hosted_domain(self, proposal):
         if self.hosted_domain:
             if (
                 user_email_domain not in self.hosted_domain
-                or bodyjs['hd'] not in self.hosted_domain
+                or ('hd' in bodyjs and bodyjs['hd'] not in self.hosted_domain)
```
consideRatio: Reading in https://developers.google.com/identity/protocols/OpenIDConnect#hd-param I wonder if this logic really is relevant. I see that the hd value can even be an `*` to indicate something entirely different.

I guess that you ran into an issue where the hd parameter wasn't passed, and you ended up with an error of trying to access the `hd` key of bodyjs, and that led to this PR?

Is the right fix perhaps to remove this section? I don't fully grasp what the hd (hosted domain) parameter indicates though.

betatim: From my reading of the docs you linked and what I know about GSuites I would say we can drop this. It is a hint to the UI of Google which G Suite hosted domain you are looking for.

cmseal: @consideRatio Your guess is correct... removing the requirement for the `hd` value, and just using the `user_email_domain` feels like a cleaner way to filter.

@betatim I'll amend my PR accordingly.

Thanks both!

consideRatio:
Thank you @cmseal for resolving this issue! :heart: :tada:

PR #333, opened 2020-02-04
Improve gitlab subgroup support

akhmerov:
When testing the usage of a gitlab subgroup we found two problems:

- Members with access to the group inherited from the parent one aren't listed in the api endpoint that is currently used
- Resolving group by name doesn't work if the name contains slashes (all subgroups do).

This fixes both problems.

### DIFF
--- a/oauthenticator/gitlab.py
+++ b/oauthenticator/gitlab.py
@@ -8,6 +8,7 @@ import os
 import re
 import sys
 import warnings
+from urllib.parse import quote
 
 from tornado.auth import OAuth2Mixin
 from tornado import web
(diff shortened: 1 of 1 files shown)


akhmerov:
Ah, the endpoint is already treated correctly, only the escaping the subgroup is needed. This is a quality of life improvement—in principle users may already provide the numerical group name.

consideRatio:
@vindvaki resolved the issue about group inheritance in https://github.com/jupyterhub/oauthenticator/pull/283 I believe, and this is now part of the 0.10.0 oauthenticator release. This fix requires GitLab 2.14 and higher.

Regarding group escaping, it makes sense to me. It raised the question of escaping the username also, I found the username currently allow a-Z + 0-9 and `.` `-` `_` - we don't need to urlescape these I think.

```ipython
In [1]: from urllib.parse import quote

In [2]: quote("testdot.testdash-testunderscore_okok")                                                                                               
Out[2]: 'testdot.testdash-testunderscore_okok'
```

This LGTM.

consideRatio:
I verified this locally, and this is indeed needed!

Issue #311, opened 2019-11-26
Best practice for picking user name for creating account on linux?


linkcd:
When using OAuthenticator, the user name is extracted from a selected field in the Id_token/access_token from idp. Then the name is used for creating the linux system level user account, then login to notebook. 

The field is often one of the following:
- email
- full name
- user id (such as guid)

# Problem
There are some challenges/consideration of which field to be used.
## 1. Use full name
Not OK, as we might have two users with the same name.

## 2. Use email 
Email  (e.g. **tom@tomsen.com**) normally ensures uniqueness, but it has two issues:
1. email is not a OK linux user name. You can force linux to use them by using --force-badname, but it is not ideal.
```bash
c.LocalAzureAdB2COAuthenticator.add_user_cmd = ['adduser', '-q', '--gecos', '""', '--disabled-password', '--force-badname']
```
2. In addition, even the badname was created by force, the jupyterhub URL exposes this username in URL, e.g. _notebook.domain.com/user/**tom@tomsen.com**/tree_ , which is an issue from privacy point of view (e.g. GDPR)

### 3. Use internal ID (i.e GUID)
This one does not have privacy issue, but it has the same problem as issue 2.1: it is not a OK linux user name. 

# Idea for solution
Inside of the authorization implementation, we convert the email or guid into a string that can be used as linux user name. 

> (from https://unix.stackexchange.com/questions/157426/what-is-the-regex-to-validate-linux-users)
> It is usually recommended to only use usernames that begin with a lower case letter or an underscore, followed by lower case letters, digits, underscores, or dashes. They can end with a dollar sign. In regular expression terms: [a-z_][a-z0-9_-]*[$]?
> 
> On Debian, the only constraints are that usernames must neither start with a dash ('-') nor contain a colon (':') or a whitespace (space: ' ', end of line: '\n', tabulation: '\t', etc.). Note that using a slash ('/') may break the default algorithm for the definition of the user's home directory.
> 
> Usernames may only be up to 32 characters long.

# The missing puzzle part
Therefore, the missing part of the solution, is to find an algorithm that can do this job. The most important rule of this algorithm is to make sure we do not have collision (ie. 2 different email/guid does not generate the same result string, resulting the same user name). 

Based on this [list](https://en.wikipedia.org/wiki/Comparison_of_cryptographic_hash_functions), looks like all well-known hash algorithms gererate too long hash values. 

[...]

PR #334, opened 2020-02-11
Truncate username to max 30 chars when fetching from AzureAD

jamescross91:
One of our client deployments has extremely long username_claim names, exceeding 63 characters.  This results in numerous cascading failures downstream, since the name field is leveraged in a number of places - such as pod name, volume name and as part of the URL

### DIFF
--- a/oauthenticator/azuread.py
+++ b/oauthenticator/azuread.py
@@ -81,7 +81,7 @@ class AzureAdOAuthenticator(OAuthenticator):
         id_token = resp_json['id_token']
         decoded = jwt.decode(id_token, verify=False)
 
-        userdict = {"name": decoded[self.username_claim]}
+        userdict = {"name": decoded[self.username_claim][0:30]}
         userdict["auth_state"] = auth_state = {}
         auth_state['access_token'] = access_token
         # results in a decoded JWT for the user data


consideRatio:
Is this issue specific only to zero-to-jupyterhub-k8s deployments, or general? I want to ensure we fix this in the right place

jamescross91:
Hi @consideRatio - we're experiencing it in zero-to-jupyterhub-k8s but my guess is any component using this code have the same problem since we're using display name here instead of username.  The display name limit is 256 characters which could blow up lots of things

manics:
Is there a guarantee that the first 30 chars are always unique amongst users?

jamescross91:
@manics no but since this is actually the display name field I don’t think it’s enforced by AD anyway.  I think if we don’t truncate we should consider using a different field

manics:
OK. I'm a bit confused now because in the description you said 
> since the name field is leveraged in a number of places - such as pod name, volume name

which implies the name needs to be unique- or do you think there's a bug somewhere else?

jamescross91:
> OK. I'm a bit confused now because in the description you said
> 
> > since the name field is leveraged in a number of places - such as pod name, volume name
> 
> which implies the name needs to be unique- or do you think there's a bug somewhere else?

@manics - so I think zero-to-jupyterhub will _fail_ if the display name is not unique - I'm just saying that I don't think the identity provider (Azure AD) is enforcing this - although I haven't checked.  The max display name length in Azure AD is 256 characters - currently this library blindly reads it in, causing downstream failures.

Hence the suggestion to truncate here for safety - actually anything up to about 50 characters should be fine I think.

manics:
Thanks for clarifying. I agree with @consideRatio, we should figure out if it's a problem specific to this authenticator or if it's a deeper problem in Jupyterhub/Kubespawner

betatim:
Do you know where it leads to problems in Z2JH? For example pod names are limited in length but we have code that should take care of that (truncate a overly long pod name and stick a unique ID at the end to keep it unique). This means it would be good to know where things break as it is likely a bug that (also) needs fixing.

betatim:
If we do have to limit (very) long names by truncation I'd suggest that we use something like
```python
if len(name) > max_name_length:
  name = "%s-%s" % (name[:23], str(hash(name))[:6])
```
so that the resulting names remain unique.

The other question is: why is it 30? and not 35 or 40 or 25? And does that mean we need to make it configurable?

Overall I think if we can get a bit more detail about what breaks when we could fix that instead of having to limit username lengths.

jamescross91:
@betatim I'm happy to update the PR to prepend a random hash.  You're right - 30 is fairly arbitrary.  The exact issue we're seeing is relating to creating a per-user persistent volume - we get an invalid value "must be no more than 63 characters" error.  30 chars was my estimate of what would be safe given that spaces are replaced by `20` and other special characters are also swapped out.

Sounds like we have two options - truncate here, or add some additional safety in Z2JH.

I'm happy with either and will take your steer - if you point me in the right direction I can raise a similar PR for Z2JH so we gain the same protection for either IDPs.

manics:
This is a related issue concerning the general problem of invalid usernames such as uniqueness, invalid characters, and length restrictions: https://github.com/jupyterhub/oauthenticator/issues/311

jamescross91:
@manics - give that we're running into this now, are we ok to truncate here for safety using @betatims approach? If so il update the PR

manics:
@jamescross91 I don't know how safe it is. I mentioned that other issue in case it was useful.
If you need a quick fix you could try overriding `normalize_username`:
- https://jupyterhub.readthedocs.io/en/stable/reference/authenticators.html#normalize-usernames
- https://github.com/jupyterhub/jupyterhub/blob/22f86ad76c437d37bc26c28d9fecfa5457ef5974/jupyterhub/auth.py#L317

For example, in Z2JH something like (I haven't tested this):
```yaml
  hub:
    extraConfig:
      custom-authenticator: |
        from oauthenticator.azuread import AzureAdOAuthenticator

        class CustomAzureAdOAuthenticator(AzureAdOAuthenticator):
            def normalize_username(self, username):
                return username[0:30] + '<hash> or something else'

        c.JupyterHub.authenticator_class = CustomAzureAdOAuthenticator
```
may work?

jamescross91:
@manics - thanks will give this a go - but I think at the very least we should be truncating in Z2JH and prepending a hash on the first part of a UUID.  Sure there isn't a 100% chance it will be unique, but a collision is way less likely than an overflow.

manics:
I agree this should be handled, or clearly documented if there's no single solution, but that might take some time to figure out so hopefully the above is a short-term solution for you.

jamescross91:
Thanks @manics - we will give it a go

linkcd:
Hi
i think the RP code must have a hash. Simply truncate original username to first 30 chars max is collision grarateed. 

e.g. you have two users
domain\verylongprefixTomHanks
domain\verylongprefixTomHanksJunior

can result same username in logging in process, then they will have access to the same instance of jupyternotebook....

manics:
After thinking about this a bit more I don't think this is the right place to truncate or hash the name. One of the reasons JupyterHub is split into several components is to allow flexibility- you can pick and mix your spawner, authenticator, etc. Each spawner or authenticator has different constraints, for example length of username or supported characters. In addition any change here may break existing users.

minrk:
I agree with @manics that the truncate+hash implementation probably belongs in KubeSpawner, since it should be used regardless of the Authenticator, and which has several similar "convert this arbitrary string into a Kubernetes-safe field" operations, but they don't include length handling at the moment. Truncating for length+hash [such as we do in binderhub](https://github.com/jupyterhub/binderhub/blob/3cb82fafa55d25e8fb6c6559423ffdef57919aa3/binderhub/builder.py#L166-L182) makes sense for KubeSpawner in general. To avoid always having these hashes, we can have KubeSpawner only truncate for names longer than `max_len - 1` (truncated+hash length must be longer than max length before truncating to avoid collisions)

manics:
Hi @jamescross91, thanks for your contribution. As you can tell from the discussion this isn't the right place to add this feature so I'm closing this PR. If you'd like to update Kubespawner instead this is the relevant file: https://github.com/jupyterhub/kubespawner/blob/master/kubespawner/spawner.py

If you'd like to chat with others in the Jupyter community and perhaps get some more input on this issue before doing further work please say hello on https://discourse.jupyter.org/ if you haven't already.

Issue #332, opened 2020-02-02
Update repos README.md - links failing etc.


consideRatio:
The README for the repo has become outdated, and links are failing.

I'd like to see us have:
- nice badges from team-compass
- a good intro to what this project is, and how it fits into the JupyterHub ecosystem of repos.
- install section (that also describes where it should be installed: next to jupyterhub)
- some quick onboarding on what OAuth is
- general notes on how to configure it, and specific examples.

rpwagner:
@NickolausDS can you watch this issue, since it's looks like the docs are being updated and maybe [moved to individual pages](https://github.com/jupyterhub/oauthenticator/tree/master/docs/source)? We may need to contribute to that for our plugin.

NickolausDS:
Yes, I'd be happy to.

dmpe:
It would be also nice to have automatic builds for docs from this repository, with a link!
https://jupyterhub.readthedocs.io/en/stable/reference/authenticators.html is only partially helpful.

PR #336, opened 2020-03-05
Update docs

GeorgianaElena:
This revamps the docs a bit. I didn't manage to cover everything, but I reordered the sections a bit, added the Jupyter logo and the nice badges from team-compass.

Closes https://github.com/jupyterhub/oauthenticator/issues/332


README.md
```diff
@@ -1,25 +1,56 @@
...
+* [General Setup](https://oauthenticator.readthedocs.io/en/latest/getting-started.html#general-setup)
+* [Azure AD](https://oauthenticator.readthedocs.io/en/latest/getting-started.html#azure-ad-setup)
+* [GitHub](https://oauthenticator.readthedocs.io/en/latest/getting-started.html#github-setup)
+* [GitLab](https://oauthenticator.readthedocs.io/en/latest/getting-started.html#gitlab-setup)
+* [Google](https://oauthenticator.readthedocs.io/en/latest/getting-started.html#google-setup)
+* [OpenShift](https://oauthenticator.readthedocs.io/en/latest/getting-started.html#openshift-setup)
+* [Okpy](https://oauthenticator.readthedocs.io/en/latest/getting-started.html#okpyauthenticator)
+* [Globus](https://oauthenticator.readthedocs.io/en/latest/getting-started.html#globus-setup)
+* [Moodle](https://oauthenticator.readthedocs.io/en/latest/getting-started.html#moodle-setup)
+* [Yandex](https://oauthenticator.readthedocs.io/en/latest/getting-started.html#yandex-setup)
```
consideRatio: Woops the Yandex authenticator is removed I think, so the documentation should also be removed.

consideRatio:
:heart: wooooo excellent work! This has been one of my "oh this really has to get done" mental notes I've had for a while!

README.md
GeorgianaElena: I know, but the example setup for Yandex is using GenericAuthenticator :yellow_heart:

GeorgianaElena: And I added a similar example for AWSCognito. I didn't tested it though :eyes:

consideRatio: Ah!

README.md
```diff
@@ -1,25 +1,56 @@
...
+plugged in and used with JupyterHub.
+
+The following authentication services are supported through their own authenticator: [Auth0](oauthenticator/auth0.py),
+[Azure AD](oauthenticator/azuread.py), [Bitbucket](oauthenticator/bitbucket.py), [CILogon](oauthenticator/cilogon.py),
+[GitHub](oauthenticator/github.py), [GitLab](oauthenticator/gitlab.py), [Globus](oauthenticator/globus.py),
+[Google](oauthenticator/google.py), [MediaWiki](oauthenticator/mediawiki.py), [Okpy](oauthenticator/okpy.py),
+[OpenShift](oauthenticator/openshift.py).
+
+There is also a [GenericAuthenticator](oauthenticator/generic.py)
+that can be used to write your own Authenticators with any OAuth 2.0 provider.
```
consideRatio: > can be used to write your own Authenticators with any OAuth 2.0 provider.

This sounds to me like we can only use the generic authenticator to create a custom class from it, but it can also be used on its own through configuration with any OAuth 2.0 provider. Sometimes though, we need something additional, and then a custom class is needed that can use the generic authenticator as a base class.

docs/source/getting-started.rst
```diff
@@ -1,65 +1,30 @@
...
+Get started
+===========
 
+The general steps to take when using OAuthenticator:
 
 1. pick your provider
 2. register with your provider
 3. configure JupyterHub (if supported, pick oauthenticator class, otherwise see :doc:`writing-an-oauthenticator`).
 4. common configuration (client_id, client_secret, callback_url, whitelist, etc.)
 5. specific configuration for your provider
```
consideRatio: - I think it would make sense to say `identity provider` throughout instead of just provider.
- I think point 3 is a bit too vague, it could be choose authenticator class or something like that instead, and avoid suggesting the creation of a custom authenticator before falling back to the generic authenticator first, and only if that isn't enough, consider creating your own. I think, not suggesting to create an own at all is best here but instead suggesting to choose a specific or the generic authenticator class.
- "common configuration" could be reworded as "configure the authenticator class" or similar to assume less from the reader.

docs/source/getting-started.rst
```diff
@@ -1,65 +1,30 @@
...
-
-A `generic implementation <oauthenticator.generic.GenericOAuthenticator>`, which you can
-use with any provider, is also available.
-
-Examples
---------
-
-For an example docker image using OAuthenticator, see the
-`examples <examples>`__ directory.
+OAuthenticator currently supports the following **authentication services**:
```
consideRatio: Here "authentication services" are used, hmmm.. Since "provider" (which is referring to "identity provider") has been used before, I like to use that here as well.

docs/source/getting-started.rst
```diff
@@ -1,65 +1,30 @@
...
-
-Or clone the repo and do a dev install:
-
-::
-
-   git clone https://github.com/jupyterhub/oauthenticator.git
-   cd oauthenticator
-   pip3 install -e .
+A `generic implementation <oauthenticator.generic.GenericOAuthenticator>`__, which you can
+use with **any provider**, is also available.
```
consideRatio: any "OAuth2 identity provider"

GeorgianaElena:
Thanks for the feedback @consideRatio :heart:. I've just added a new commit that addresses your comments.

I know the test failures are unrelated. I think it has something to do with the [`force_instace = True`](https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/generic.py#L74) just added but I don't know why. Do you have any ideas on how to fix them?

consideRatio:
Sorry for slow followup @GeorgianaElena, thanks so much for putting in this work! It makes all the difference to have good documentation! :heart: :tada: I'm very happy about this PR!!

PR #337, opened 2020-03-06
Streamline Globus requirements

NickolausDS:
The Globus OAuthenticator currently requires a custom dependency
in order to authenticate users. The Globus SDK was originally
chosen for its convenience functions in generating URLs and
automatically structuring tokens. However, it's not necessary
and can be replaced with generic components included in this
repo.

With these changes, the globus_sdk can also be removed from the
Zero to Jupyterhub 'hub' dependencies. While the globus_sdk is
extremely useful for transferring data within the Single User
Server, in the hub it is only used for the OAuthenticator.

### DIFF
--- a/docs/source/getting-started.rst
+++ b/docs/source/getting-started.rst
@@ -388,12 +388,11 @@ Set the above settings in your ``jupyterhub_config``:
 .. code:: python
 
    # Tell JupyterHub to create system accounts
-   from oauthenticator.globus import LocalGlobusOAuthenticator
-   c.JupyterHub.authenticator_class = LocalGlobusOAuthenticator
-   c.LocalGlobusOAuthenticator.enable_auth_state = True
-   c.LocalGlobusOAuthenticator.oauth_callback_url = 'https://[your-host]/hub/oauth_callback'
-   c.LocalGlobusOAuthenticator.client_id = '[your app client id]'
-   c.LocalGlobusOAuthenticator.client_secret = '[your app client secret]'
+   from oauthenticator.globus import GlobusOAuthenticator
+   c.JupyterHub.authenticator_class = GlobusOAuthenticator
+   c.GlobusOAuthenticator.oauth_callback_url = 'https://[your-host]/hub/oauth_callback'
+   c.GlobusOAuthenticator.client_id = '[your app client id]'
+   c.GlobusOAuthenticator.client_secret = '[your app client secret]'
 
 Alternatively you can set env variables for the following:
 ``OAUTH_CALLBACK_URL``, ``OAUTH_CLIENT_ID``, and
@@ -406,13 +405,6 @@ settings related to User Identity, Transfer, and additional security.
 User Identity
 ~~~~~~~~~~~~~
 
-By default, all users are restricted to their *Globus IDs*
-(example@globusid.org) with the default Jupyterhub config:
-
-.. code:: python
-
-   c.GlobusOAuthenticator.identity_provider = 'globusid.org'
-
 If you want to use a *Linked Identity* such as
 ``malcolm@universityofindependence.edu``, go to your `App Developer
 page <http://developers.globus.org>`__ and set *Required Identity
@@ -421,12 +413,22 @@ in the config:
 
 .. code:: python
 
-   c.GlobusOAuthenticator.identity_provider = 'universityofindependence.edu'
+   c.GlobusOAuthenticator.identity_provider = 'uchicago.edu'
+
+**Pitfall**: Don't set 'Required Identity Provider' on pre-existing apps!
+Previous user login consents will be tied to the identity users initially used
+to login, and will continue to be tied to that identity after changing this
+setting. Create a new Globus App with your preferred 'Required Identity Provider'
+to avoid this problem.
 
 Globus Scopes and Transfer
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The default configuration will automatically setup user environments
+The following shows how to get tokens into user Notebooks. `You can see how users
+use tokens here <https://github.com/globus/globus-jupyter-notebooks/blob/master/JupyterHub_Integration.ipynb>`__.
+If you want a demonstration, you can visit `The Jupyter Globus Demo Server <https://jupyter.demo.globus.org>`__.
+
+The default server configuration will automatically setup user environments
 with tokens, allowing them to start up python notebooks and initiate
 Globus Transfers. If you want to transfer data onto your JupyterHub
 server, it’s suggested you install `Globus Connect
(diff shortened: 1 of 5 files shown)


docs/source/getting-started.rst
```diff
@@ -388,12 +388,11 @@ Set the above settings in your ``jupyterhub_config``:
...
-   c.JupyterHub.authenticator_class = LocalGlobusOAuthenticator
-   c.LocalGlobusOAuthenticator.enable_auth_state = True
-   c.LocalGlobusOAuthenticator.oauth_callback_url = 'https://[your-host]/hub/oauth_callback'
-   c.LocalGlobusOAuthenticator.client_id = '[your app client id]'
-   c.LocalGlobusOAuthenticator.client_secret = '[your app client secret]'
+   from oauthenticator.globus import GlobusOAuthenticator
+   c.JupyterHub.authenticator_class = GlobusOAuthenticator
+   c.GlobusOAuthenticator.oauth_callback_url = 'https://[your-host]/hub/oauth_callback'
+   c.GlobusOAuthenticator.client_id = '[your app client id]'
+   c.GlobusOAuthenticator.client_secret = '[your app client secret]'
```
NickolausDS: The docs here tweak a couple things. `GlobusOAuthenticator` is probably better since it assumes less about the context the admin is using the Authenticator. In Zero to Jupyterhub, for example, I don't think LocalGlobusOAuthenticator is useful. This makes the config a little more copy-pastable.

I also removed the `c.LocalGlobusOAuthenticator.enable_auth_state = True`, since it isn't needed in the basic configuration, only when injecting tokens into user notebooks.

oauthenticator/globus.py
```diff
@@ -36,27 +28,36 @@ class GlobusLogoutHandler(LogoutHandler):
...
         if state:
-            self.authenticator.revoke_service_tokens(state.get('tokens'))
+            await self.authenticator.revoke_service_tokens(state.get('tokens'))
             self.log.info(
                 'Logout: Revoked tokens for user "{}" services: {}'.format(
                     user.name, ','.join(state['tokens'].keys())
                 )
             )
-            state['tokens'] = ''
+            state['tokens'] = {}
```
NickolausDS: This is a minor change, unrelated to removing the Globus SDK. Unit testing caught that it's possible to get tripped up by the type changing from dict to str, although in practice I think it would be difficult to get into a state where this caused an error. This change keeps types consistent, just in case.

oauthenticator/globus.py
```diff
@@ -79,7 +96,7 @@ def _authorize_url_default(self):
     ).tag(config=True)
 
     def _identity_provider_default(self):
-        return os.getenv('IDENTITY_PROVIDER', 'globusid.org')
+        return os.getenv('IDENTITY_PROVIDER', '')
```
NickolausDS: This is another change unrelated to streamlining dependencies, but I think it's preferable. The previous value of `globusid.org` essentially locks down the default installation to an identity almost nobody uses. Setting this to the empty string allows the user to login with any IdP initially.

oauthenticator/globus.py
```diff
@@ -96,16 +113,6 @@ def _scope_default(self):
...
-    allow_refresh_tokens = Bool(
-        help="""Allow users to have Refresh Tokens. If Refresh Tokens are not
-        allowed, users must use regular Access Tokens which will expire after
-        a set time. Set to False for increased security, True for increased
-        convenience."""
-    ).tag(config=True)
-
-    def _allow_refresh_tokens_default(self):
-        return True
-
```
NickolausDS: I think the last refactor pulled out the code that could request refresh tokens, so this feature is no longer usable. In order to request refresh tokens, an additional parameter needs to be sent through the first leg of the OAuth flow.

oauthenticator/globus.py
```diff
@@ -174,44 +185,67 @@ def globus_portal_client(self):
...
         return {
             'name': username,
             'auth_state': {
                 'client_id': self.client_id,
-                'tokens': {
-                    tok: v
-                    for tok, v in tokens.by_resource_server.items()
-                    if tok not in self.exclude_tokens
-                },
+                'tokens': by_resource_server,
```
NickolausDS: The whole `authenticate()` method is very similar to the `GenericOAuthenticator`, where the only functional difference is the shape of tokens it returns. When the admin sets custom scopes, including scopes for third party resource servers secured with Globus Auth, the access/refresh tokens will come through the `other_tokens` key. If the admin didn't enable `auth_state`, this method behaves much the same as the `GenericOAuthenticator`

oauthenticator/tests/test_globus.py
```diff
@@ -177,26 +241,22 @@ def test_revoke_tokens(monkeypatch):
...
 
     logout_handler.clear_login_cookie = Mock()
-    authenticator.revoke_service_tokens = Mock()
     authenticator.revoke_tokens_on_logout = True
 
     await logout_handler.get()
-    assert authenticator.revoke_service_tokens.called
     auth_state = await mock_globus_user.get_auth_state()
-    assert auth_state == {'tokens': ''}
+    assert auth_state == {'tokens': {}}
```
NickolausDS: Globus Test coverage is now 100%! 

Unfortunately, these changes don't fix the GenericOAuthenticator tests.

NickolausDS:
Rebased to include fixes for Generic tests #339. All tests pass!

NickolausDS:
Rebased on #343, which better organized the globus_sdk dependency install for Globus. These changes remove that dependency entirely to bring the Globus OAuthenticator more inline with how the other OAuthenticators work, instead of relying on oauth2 shortcuts within the globus_sdk. The implementation here is very similar to the Generic OAuthenticator, with some extensions to token handling to allow Globus Transfer tokens and any other third party tokens for user-created services.

consideRatio:
@NickolausDS I ran out of time for the moment to review the full PR, but I read through 50% and think everything has made sense so far. I also concluded that you are in this PR changing files in a way that will only influence globus.

NickolausDS:
@consideRatio Correct, this only modifies Globus. 

#338 does affect this PR (it brings back refresh tokens), but I didn't want to add major changes outside of `globus.py` and `test_globus.py`, and so put it in a separate PR. Happy to do the same with other changes here, if you think they warrant further discussion. 

Thanks for reviewing!

docs/source/getting-started.rst
```diff
@@ -436,21 +438,20 @@ other behavior, you can modify the defaults below:
...
 .. code:: python
 
-   # Allow Refresh Tokens in user notebooks. Disallow these for increased security,
-   # allow them for better usability.
-   c.LocalGlobusOAuthenticator.allow_refresh_tokens = True
+   # Allow saving user tokens to the database
+   c.GlobusOAuthenticator.enable_auth_state = True
    # Default scopes are below if unspecified. Add a custom transfer server if you have one.
-   c.LocalGlobusOAuthenticator.scope = ['openid', 'profile', 'urn:globus:auth:scope:transfer.api.globus.org:all']
+   c.GlobusOAuthenticator.scope = ['openid', 'profile', 'urn:globus:auth:scope:transfer.api.globus.org:all']
```
minrk, line 444: 👍

setup.py
```diff
@@ -113,7 +113,6 @@ def run(self):
 
 setup_args['extras_require'] = {
     'googlegroups': ['google-api-python-client==1.7.11', 'google-auth-oauthlib==0.4.1'],
-    'globus': ['globus_sdk[jwt]>=1.0.0,<2.0.0']
```
minrk, line 116: Awesome!

Issue #298, opened 2019-11-01
Google OAuth: prompt=select_account 


bmshouse:
Situation: User attempts to login to a service while already authenticated to their personal email address. The service is restricted to Internal Email addresses only and the users is immediately redirected to a Google forbidden (403) page.  

Potential Solution: Google offers an [Authorization Parameter](https://developers.google.com/identity/protocols/OAuth2WebServer#creatingclient) "prompt" that when set to "select_account" will allow users to pick/login to another account. 

Problem: I have failed to identify where I can add this authorization paramater.  It seems like extra_params mentioned in #244 could address this also. 

I appreciate any assistance you can offer.  Thank you.

manics:
Solved by https://github.com/jupyterhub/oauthenticator/pull/338

Issue #244, opened 2019-03-08
[Generic] extra_params not exposed for the authorize request


consideRatio:
We are making various requests while working with _authz_ (authorization) and _authn_ (authentication). A common request is to do the initial authorization request with _client_id_,_scope_ and _response_type_, _redirect_uri_ etc. This is done in the `oauth2.py` file by relying on a tornado function that the class has inherited and can be read about [here](https://www.tornadoweb.org/en/stable/auth.html).

https://github.com/jupyterhub/oauthenticator/blob/5b0b395326292d526a4c924c34c6d9d1d06818af/oauthenticator/oauth2.py#L80-L90

As we can see, we use the `extra_params` in a hardcoded fashion passing only a state, something very important to protect against CSRF to be explained elsewhere. But, we may want to pass additional parameters in this request, for example I'm now required by Okta to pass along one more to work well with them (which sadly is absurd).

Anyhow, we should allow this parameters to be configurable I think. But, also note that since we will make multiple types of request, we must scope the parameters to the associated request. I know the `generic.py` file defines a `extra_params` traitlet:

https://github.com/jupyterhub/oauthenticator/blob/5b0b395326292d526a4c924c34c6d9d1d06818af/oauthenticator/generic.py#L52-L54

But this extra_params is only used in the second request associated with OAUTH2_TOKEN_URL, the request of a token which can be used to fetch user information from the userinfo endpoint.

---

## Conclusion
We can now using `generic.py`'s `GenericOAuthenticator` configure the extra params passed to the `OAUTH2_TOKEN_URL`, but we cannot configure the params passed to `OAUTH2_USERDATA_URL` or the initial request to `OAUTH2_AUTHORIZE_URL`.

## Suggestion
We add traitlets to configure this, and name them related to AUTHORIZE, TOKEN, USERDATA.

> NOTE: USERDATA is typically referred to as USERINFO in the Open ID Connect (OIDC) specifications (`authorization_endpoint`, `token_endpoint`, `userinfo_endpoint`)

erowan:
I need to pass extra info in here too

d0m84:
I faced the same challenge today and it's not straight forward. Finally I fixed it by reading the code.
My impression was that `extra_params` will be used for the authorization request as well, but it's not.

Finally I fixed it by adding my own custom parameter to the environment variable. 

`OAUTH2_AUTHORIZE_URL="https://auth.example.com/authorize?my_param=example"`

[...]

Issue #211, opened 2018-08-30
refresh token on OAuth2 module in Jupyter


michec81:
referring to this old issue:

https://github.com/jupyterhub/oauthenticator/issues/203#issuecomment-417228318

 I find a small problem after doing some test on my application.
I realized that the pre_spawn_start method configures correctly the token bearer in the spawner environment, but the token after a while expires as we expect. Since I pass the token to a remote application service the authentication fails. So I suppose the token should be refreshed someway but I don't know how this could be done. Moreover, for what I know, I can pass the token to the spanwer only using the pre_spawn_start(self, user, spawner) method (/usr/local/lib/python3.6/dist-packages/oauthenticator/generic.py).
Could you give me any help on this?

Thank you again,

Cheers,
Michele

michec81:
I think the problem arises as the pre_spanwer method is executed only once and the spawner_environment variable is not refreshed anymore.

  #added by CINECA
    @gen.coroutine
    def pre_spawn_start(self, user, spawner):
        auth_state = yield user.get_auth_state()
        if not auth_state:
             return 
        spawner.environment['OAUTH2_AUTHORIZE_TOKEN'] = auth_state['access_token']
        self.log.info("Passing access token variable to user spawn process %s", auth_state['access_token'])

gweis:
The Access token is usually meant to be a short lived token. In general it has a validity time of a couple of minutes. 

Usually can use a refresh token to renew an access token, or you might be able to configure the access token validity time.

Also the environment variable inside your notebook container can't be refreshed by an outside process. Your only chance is to pass along a refresh token and have some sort of  process in your notebook server to regularly update the access token.

michec81:
I know, but the point is that if I logout and login again, the pre_spawn_start is not called again and old token is still present in the environment variable.

Is it possible to access user.get_auth_state() directly from the notebook so I can be sure to use the current token, rather then previously set and stored in the environment?

is the same problem like this:
https://stackoverflow.com/questions/51287958/how-to-access-current-user-auth-state-from-notebook

Thank you again!

michec81:
maybe a solution could be here:

https://github.com/jupyterhub/jupyterhub/issues/337 

but I still not understand what to do

PaulMazzuca:
[...]

PR #338, opened 2020-03-06
Extend oauth2 params sent to provider on first leg of oauth2 flow

NickolausDS:
Changes for #244. These changes expose a generic config option for passing in arbitrary options alongside the `state` parameter. One example is requesting refresh tokens, as in #211, which should look like this:

```
c.JupyterHub.authenticator_class.extra_authorize_params = {'access_type': 'offline'}
```

This may also solve #298 in order to select arbitrary accounts for Google. I created an app and tested the following:

```
c.LocalGoogleOAuthenticator.extra_authorize_params = {'prompt': 'select_account'}
```
Which generated the following URL: 
```
302 GET /hub/oauth_login?next= -> 
https://accounts.google.com/o/oauth2/v2/auth?
response_type=code&
redirect_uri=http%3A%2F%2Flocalhost%3A8000%2Fhub%2Foauth_callback&
client_id=[client]&
prompt=select_account&
state=[secret]&
scope=openid+email
```
This appeared to work, although I'm not familiar enough with Google to know if that worked exactly as intended. I tried requesting Google refresh tokens as well, but it didn't seem to want to give me any. I also tested with Globus, and was able to successfully request refresh tokens simply by setting `{'access_type': 'offline'}`.

### DIFF
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -18,7 +18,7 @@ from jupyterhub.handlers import BaseHandler
 from jupyterhub.auth import Authenticator
 from jupyterhub.utils import url_path_join
 
-from traitlets import Unicode, Bool, List, default
+from traitlets import Unicode, Bool, List, Dict, default
 
 
 def guess_callback_uri(protocol, host, hub_server_url):
@@ -72,6 +72,10 @@ class OAuthLoginHandler(OAuth2Mixin, BaseHandler):
     def _OAUTH_USERINFO_URL(self):
         return self.authenticator.userdata_url
 
+    @property
+    def _OAUTH_EXTRA_AUTHORIZE_PARAMS(self):
+        return self.authenticator.extra_authorize_params
+
     def set_state_cookie(self, state):
         self.set_secure_cookie(STATE_COOKIE_NAME, state, expires_days=1, httponly=True)
 
@@ -100,14 +104,16 @@ class OAuthLoginHandler(OAuth2Mixin, BaseHandler):
 
     def get(self):
         redirect_uri = self.authenticator.get_callback_url(self)
+        extra_params = self.authenticator.extra_authorize_params.copy()
         self.log.info('OAuth redirect: %r', redirect_uri)
         state = self.get_state()
         self.set_state_cookie(state)
+        extra_params['state'] = state
         self.authorize_redirect(
             redirect_uri=redirect_uri,
             client_id=self.authenticator.client_id,
             scope=self.authenticator.scope,
-            extra_params={'state': state},
+            extra_params=extra_params,
             response_type='code',
         )
 
@@ -259,6 +265,16 @@ class OAuthenticator(Authenticator):
         See the OAuth documentation of your OAuth provider for options.
         For GitHub in particular, you can see github_scopes.md in this repo.
         """,
+    )
+
+    @default("extra_authorize_params")
+    def _extra_authorize_params(self):
+        return os.environ.get("OAUTH2_EXTRA_AUTHORIZE_PARAMS", {})
+
+    extra_authorize_params = Dict(
+        config=True,
+        help="""Extra GET params to send along with the initial OAuth request
+        to the OAuth provider.""",
     )
 
     login_service = 'override in subclass'
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -72,10 +72,6 @@ class OAuthLoginHandler(OAuth2Mixin, BaseHandler):
     def _OAUTH_USERINFO_URL(self):
         return self.authenticator.userdata_url
 
-    @property
-    def _OAUTH_EXTRA_AUTHORIZE_PARAMS(self):
-        return self.authenticator.extra_authorize_params
-
     def set_state_cookie(self, state):
         self.set_secure_cookie(STATE_COOKIE_NAME, state, expires_days=1, httponly=True)
 
@@ -266,10 +262,6 @@ class OAuthenticator(Authenticator):
         For GitHub in particular, you can see github_scopes.md in this repo.
         """,
     )
-
-    @default("extra_authorize_params")
-    def _extra_authorize_params(self):
-        return os.environ.get("OAUTH2_EXTRA_AUTHORIZE_PARAMS", {})
 
     extra_authorize_params = Dict(
         config=True,


missingcharacter:
Personally, I'd like for `extra_params` to be set at the `Oauthenticator` class level, something like this https://github.com/missingcharacter/oauthenticator/commit/62352ab7254e6779c39c8de9658740dcfbc87e39

BTW, tests pass

```
$ pytest -v ./oauthenticator/tests/
============================================================================================================== test session starts ===============================================================================================================
platform darwin -- Python 3.8.2, pytest-5.4.1, py-1.8.1, pluggy-0.13.1 -- /Users/ricardo/src/oauthenticator/venv/bin/python3
cachedir: .pytest_cache
rootdir: /Users/ricardo/src/oauthenticator
plugins: requests-mock-1.7.0, asyncio-0.10.0, cov-2.8.1
collected 35 items

oauthenticator/tests/test_auth0.py::test_auth0 PASSED                                                                                                                                                                                      [  2%]
oauthenticator/tests/test_azuread.py::test_tenant_id_from_env PASSED                                                                                                                                                                       [  5%]
oauthenticator/tests/test_bitbucket.py::test_bitbucket PASSED                                                                                                                                                                              [  8%]
oauthenticator/tests/test_bitbucket.py::test_team_whitelist PASSED                                                                                                                                                                         [ 11%]
oauthenticator/tests/test_cilogon.py::test_cilogon PASSED                                                                                                                                                                                  [ 14%]
oauthenticator/tests/test_cilogon.py::test_cilogon_alternate_claim PASSED                                                                                                                                                                  [ 17%]
oauthenticator/tests/test_cilogon.py::test_cilogon_additional_claim PASSED                                                                                                                                                                 [ 20%]
oauthenticator/tests/test_cilogon.py::test_cilogon_missing_alternate_claim PASSED                                                                                                                                                          [ 22%]
oauthenticator/tests/test_generic.py::test_generic PASSED                                                                                                                                                                                  [ 25%]
oauthenticator/tests/test_generic.py::test_generic_callable_username_key PASSED                                                                                                                                                            [ 28%]
oauthenticator/tests/test_github.py::test_github PASSED                                                                                                                                                                                    [ 31%]
oauthenticator/tests/test_github.py::test_org_whitelist PASSED                                                                                                                                                                             [ 34%]
oauthenticator/tests/test_gitlab.py::test_gitlab PASSED                                                                                                                                                                                    [ 37%]
oauthenticator/tests/test_gitlab.py::test_group_whitelist PASSED                                                                                                                                                                           [ 40%]
oauthenticator/tests/test_gitlab.py::test_project_id_whitelist PASSED                                                                                                                                                                      [ 42%]
oauthenticator/tests/test_globus.py::test_globus PASSED                                                                                                                                                                                    [ 45%]
oauthenticator/tests/test_globus.py::test_globus_pre_spawn_start PASSED                                                                                                                                                                    [ 48%]
oauthenticator/tests/test_globus.py::test_allow_refresh_tokens PASSED                                                                                                                                                                      [ 51%]
oauthenticator/tests/test_globus.py::test_restricted_domain PASSED                                                                                                                                                                         [ 54%]
oauthenticator/tests/test_globus.py::test_namespaced_domain PASSED                                                                                                                                                                         [ 57%]
oauthenticator/tests/test_globus.py::test_token_exclusion PASSED                                                                                                                                                                           [ 60%]
oauthenticator/tests/test_globus.py::test_revoke_tokens PASSED                                                                                                                                                                             [ 62%]
oauthenticator/tests/test_globus.py::test_custom_logout PASSED                                                                                                                                                                             [ 65%]
oauthenticator/tests/test_globus.py::test_logout_revokes_tokens PASSED                                                                                                                                                                     [ 68%]
oauthenticator/tests/test_google.py::test_google PASSED                                                                                                                                                                                    [ 71%]
oauthenticator/tests/test_google.py::test_hosted_domain PASSED                                                                                                                                                                             [ 74%]
oauthenticator/tests/test_google.py::test_multiple_hosted_domain PASSED                                                                                                                                                                    [ 77%]
oauthenticator/tests/test_google.py::test_admin_google_groups PASSED                                                                                                                                                                       [ 80%]
oauthenticator/tests/test_google.py::test_whitelisted_google_groups PASSED                                                                                                                                                                 [ 82%]
oauthenticator/tests/test_mediawiki.py::test_mediawiki PASSED                                                                                                                                                                              [ 85%]
oauthenticator/tests/test_mediawiki.py::test_login_redirect PASSED                                                                                                                                                                         [ 88%]
oauthenticator/tests/test_oauth2.py::test_serialize_state PASSED                                                                                                                                                                           [ 91%]
oauthenticator/tests/test_okpy.py::test_okpy PASSED                                                                                                                                                                                        [ 94%]
oauthenticator/tests/test_okpy.py::test_no_code PASSED                                                                                                                                                                                     [ 97%]
oauthenticator/tests/test_openshift.py::test_openshift PASSED                                                                                                                                                                              [100%]

================================================================================================================ warnings summary ================================================================================================================
venv/lib/python3.8/site-packages/pytest_asyncio/plugin.py:39: 32 tests with warnings
  /Users/ricardo/src/oauthenticator/venv/lib/python3.8/site-packages/pytest_asyncio/plugin.py:39: PytestDeprecationWarning: direct construction of Function has been deprecated, please use Function.from_parent
    item = pytest.Function(name, parent=collector)

venv/lib/python3.8/site-packages/pytest_asyncio/plugin.py:45: 32 tests with warnings
  /Users/ricardo/src/oauthenticator/venv/lib/python3.8/site-packages/pytest_asyncio/plugin.py:45: PytestDeprecationWarning: direct construction of Function has been deprecated, please use Function.from_parent
    item = pytest.Function(name, parent=collector)  # To reload keywords.

-- Docs: https://docs.pytest.org/en/latest/warnings.html
======================================================================================================== 35 passed, 64 warnings in 0.45s =========================================================================================================
```

missingcharacter:
I'm just making my example work, I'm pretty sure there are better ways to do it, I'm not smart enough to figure these out yet.

NickolausDS:
The current changes here expose the extra params for any Oauthenticator using the default LoginHandler (Or a derivative). For example, any OAuthenticator would be able to configure custom params through the Jupyterhub Config file without overriding LoginHandler:
```
# Config enabling Google Refresh Tokens
from oauthenticator.google import GoogleOAuthenticator
c.JupyterHub.authenticator_class = GoogleOAuthenticator
c.LocalGoogleOAuthenticator.extra_authorize_params = {'access_type': 'offline'}
```
Or for CILogon
```
from oauthenticator.cilogon import CILogonOAuthenticator
c.JupyterHub.authenticator_class = CILogonOAuthenticator
c.CILogonOAuthenticator.extra_authorize_params = {'selected_idp': 'https://accounts.google.com'}
```
Or for Globus
```
# Enable refresh tokens.
from oauthenticator.globus import LocalGlobusOAuthenticator
c.JupyterHub.authenticator_class = LocalGlobusOAuthenticator
c.LocalGlobusOAuthenticator.extra_authorize_params = {'access_type': 'offline'}
```
It would be nice to have these params generally configurable for any OAuthentictor.

missingcharacter:
@NickolausDS I see that, what I see in my approach that I don't see in yours is that in my approach users could just update to a new version and won't need to change anything for their current setup to work, and if they decide to add more params they can just do that too.

NickolausDS:
For CiLogon, your changes would be a nice way to keep the explicit `idp` and `skin` settings.

`extra_params` is also used by the [Generic OAuthenticator](https://github.com/jupyterhub/oauthenticator/blob/e133e9172919dc527fba9221bc51d90802d10fa3/oauthenticator/generic.py#L29), using that would be a bit of a namespace collision. I think the Generic OAuthenticator would override the value in that context, but it would be confusing since they're used for different things (It looks like the Generic OAuthenticator uses `extra_params` when it `POST`s back the [Authorization Code](https://github.com/jupyterhub/oauthenticator/blob/e133e9172919dc527fba9221bc51d90802d10fa3/oauthenticator/generic.py#L85), not when setting query params on the initial request to the IdP). Might be good to name it something other than `extra_params` so it isn't confused with the usage here.

missingcharacter:
@NickolausDS I went ahead and did it

https://github.com/jupyterhub/oauthenticator/blob/10030d6f6769ff1b0018c070c3802145190fdf17/oauthenticator/cilogon.py#L38-L49

https://github.com/jupyterhub/oauthenticator/blob/10030d6f6769ff1b0018c070c3802145190fdf17/oauthenticator/oauth2.py#L236-L242

tests passed

I guess we should still wait to hear back from the maintainers on what approach they'd prefer

oauthenticator/oauth2.py
```diff
@@ -261,6 +267,16 @@ def _userdata_url_default(self):
         """,
     )
 
+    @default("extra_authorize_params")
+    def _extra_authorize_params(self):
+        return os.environ.get("OAUTH2_EXTRA_AUTHORIZE_PARAMS", {})
```
manics, line 272: The environment variable will be a string, so you'll somehow need to convert it to a `dict`. Might be easier to remove it?

oauthenticator/oauth2.py
```diff
@@ -72,6 +72,10 @@ def _OAUTH_ACCESS_TOKEN_URL(self):
     def _OAUTH_USERINFO_URL(self):
         return self.authenticator.userdata_url
 
+    @property
+    def _OAUTH_EXTRA_AUTHORIZE_PARAMS(self):
```
manics, line 76: Is this needed? The comment above says "these URLs are part of the OAuth2Mixin API get them from the Authenticator object", but this property isn't listed in https://www.tornadoweb.org/en/stable/auth.html#tornado.auth.OAuthMixin

NickolausDS: No, it isn't needed. I originally added it to be consistent with the other properties. Looking at this again, I think you're right and it's probably better to remove it.

oauthenticator/oauth2.py
NickolausDS: That's a good point, I'll take this out.

missingcharacter: @NickolausDS instead of taking it out, why not do something like this?

https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/google.py#L87-L95

NickolausDS: That's a good idea, but it will be a bit more difficult to parse a dict rather than a list. For a dict type, encoding as a JSON string would probably be easiest. However, JSON is syntax heavy and uncomfortable to cram into an environment string. 

Unless you think there's a need for it, I think the most common use-case would be providing these extra params as part of the jupyterhub.py config. If we did want to add an additional config option per-provider, such as `REFRESH_TOKENS=True` for `google.py`, that could be added as a separate property that updates `extra_authorize_params`. Each provider seems to be pretty unique for which extra parameters it supports (refresh tokens seem to be the most common, but not all providers support them), so this might be the best middle ground.

missingcharacter: I was thinking something like `:` separates key from value, and `;` separates each entry. But you are right this is not necessary at all.

manics: 👍 It should be easy enough to add this in future if there's demand, it's much hard to remove/change maintain something once it's added.

NickolausDS:
@manics I applied both your suggestions. Thank you for reviewing! Also thanks to @missingcharacter for suggestions and testing. 

I retested with Globus to make sure I could supply params in the config file. Let me know if you see anything else.

PR #339, opened 2020-03-09
Fix failing GenericOAuthenticator tests

GeorgianaElena:
Creating the `AsyncHTTPClient` with `force_instance=True` attribute in `GenericOAuthenticator` made the tests fail because the mocked http client was not being used anymore and instead an actual `AsyncHttpClient` instance was being created and used (ref: https://github.com/tornadoweb/tornado/blob/master/tornado/httpclient.py#L200).

This PR forces the `GenericOAuthenticator` tests use the mocked http client.

### DIFF
--- a/oauthenticator/generic.py
+++ b/oauthenticator/generic.py
@@ -68,10 +68,14 @@ class GenericOAuthenticator(OAuthenticator):
         help="Disable basic authentication for access token request",
     )
 
+    def http_client(self):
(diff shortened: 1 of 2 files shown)


consideRatio:
I still don't fully grasp this PR, but I learned enough to say it LGTM! I have never properly read up on Pytest mocking, and still haven't, but read enough to grasp what goes on with the patch() call thanks to this guide: https://realpython.com/python-mock-library/#patch

PR #343, opened 2020-03-18
Reorganizing extras_require

missingcharacter:
I thought this could tidy things up a bit

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -110,9 +110,25 @@ with open('requirements.txt') as f:
             continue
         install_requires.append(req)
 
-setup_args['extras_require'] = {
-    'googlegroups': ['google-api-python-client==1.7.11', 'google-auth-oauthlib==0.4.1']
-}
+
+setup_args['extras_require'] = install_extra_require = {}
+
+install_extra_require['googlegroups'] = []
+with open('googlegroups-requirements.txt') as f:
+    for line in f.readlines():
+        req = line.strip()
+        if not req or req.startswith(('-e', '#', '-r')):
+            continue
+        install_extra_require['googlegroups'].append(req)
+
+install_extra_require['globus'] = []
+with open('globus-requirements.txt') as f:
+    for line in f.readlines():
+        req = line.strip()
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -110,9 +110,25 @@ def run(self):
             continue
         install_requires.append(req)
 
-setup_args['extras_require'] = {
-    'googlegroups': ['google-api-python-client==1.7.11', 'google-auth-oauthlib==0.4.1']
-}
+
```
GeorgianaElena: @missingcharacter, what do you think about leaving `setup.py` unchanged and deal with `googlegroups` extra requirements similar with how the extra [globus requirements](https://github.com/jupyterhub/oauthenticator/blob/44121eab53ec1a6f6c957b64f438f3cd65377845/oauthenticator/globus.py#L24) are handled?
* Keep the googlegroups-requirements.txt
* Update the docs
* Update the comments of missing imports
* Add googlegroups-requirements.txt to test-requirements

missingcharacter: I think it is a better user experience if people can install dependencies using `pip install package[extras_require]`, I could update globus to also show a message indicating these dependencies can be installed using `pip install oauthenticator[globus]` and update docs I find in this repo that reference globus too.

If you still think strongly about not using `pip install oauthenticator[extras_require]` I'll make the change you requested.

GeorgianaElena: You're right, @missingcharacter, setting up extra requirements might be the more user friendly way of supporting these optional features.

As long as the same mechanism is used for all the optional features (globus and googlegroups), I'm fine with both solutions. :+1: 

If using `extras_require`, I'm wondering if we could even get rid of the `globus-requirements.txt` and `googlegroups-requirements.txt` files. We could then put the dependencies directly in `extras_require` list, as they aren't that many.
What do you think?

missingcharacter: what do you think of this @GeorgianaElena ?

missingcharacter:
build 572 passed https://travis-ci.org/github/jupyterhub/oauthenticator/builds/665894434?utm_medium=notification&utm_source=github_status

PR #352, opened 2020-05-14
Fix broken link for OpenShift OAuth service accounts

nscozzaro:
Previous link was broken,(https://docs.openshift.org/latest/architecture/additional_concepts/authentication.html#service-accounts-as-oauth-clients)

So I'm suggesting replacing with this link:
https://docs.openshift.com/container-platform/latest/authentication/using-service-accounts-as-oauth-client.html


minrk:
Thanks!

PR #357, opened 2020-06-18
[Globus] Support custom username handling

NickolausDS:
Move the code responsible for fetching the username to a separate get_username() method so it can be overridden.

An example is running a Globus Jupyterhub Server that can accept logins from one or more providers (Google, ORCID, Any university in CILogon, etc.), wants needs to ensure usernames are unique. That way, person@uchicago.edu and person@gmail.com are assigned different usernames. The jupyterhub config could look like this: 

```
import base64

class OmniProviderOAuthenticator(LocalGlobusOAuthenticator):
    def get_username(self, user_info):
        username, domain = user_info.get('preferred_username').split('@', 1)
        # 'sub' is a uniquely generated uuid by Globus for each user identity
        enc_sub = base64.b32encode(bytes(user_info['sub'], 'utf-8')).decode('utf-8')
        return f'{username}-{enc_sub}'[:32]

c.JupyterHub.authenticator_class = OmniProviderOAuthenticator
```

This is similar to the [GenericOAuthenticator](https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/generic.py#L140) username picker, except the key in Globus will never change, so a callable is always preferred.

### DIFF
--- a/oauthenticator/globus.py
+++ b/oauthenticator/globus.py
@@ -171,20 +171,7 @@ class GlobusOAuthenticator(OAuthenticator):
         user_headers['Authorization'] = 'Bearer {}'.format(token_json['access_token'])
         req = HTTPRequest(self.userdata_url, method='GET', headers=user_headers)
         user_resp = await http_client.fetch(req)
-        user_json = json.loads(user_resp.body.decode('utf8', 'replace'))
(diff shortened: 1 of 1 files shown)


NickolausDS:
Thank you!

PR #358, opened 2020-06-18
[Google] Allow for checking of google_groups for admin only

dwilliams782:
I was testing the Google Group integration because I wanted to use it to specify a google group for JupyterHub admins, however the current implementation expects that you specify a list of groups for regular user access too.

This PR enables the user to either omit or specify an empty dict for google_group_whitelist:

```
c.GoogleOAuthenticator.admin_google_groups = {'example.com': ['admingroup']} 

c.GoogleOAuthenticator.google_group_whitelist = {}
```

providing admin group functionality without requiring a list of groups to be present.


minrk:
Rebased and merged. Thanks for your patience!

PR #359, opened 2020-07-06
Support for Auth0's Custom Domain fetaure

prateek2408:
Currently there is only support to add subdomain in which we have to provide something like "hello.eu" and ".auth0.com" is appended in the domain string.

This case does not work if we try to provide a custom domain which can be created in Auth0.


prateek2408:
Closing this Request as there are some changes still pending.

PR #274, opened 2019-06-13
[GenericOAuthenticator] Add client_id and client_secret to params

kkapper:
This relates to, and should fix, https://github.com/jupyterhub/oauthenticator/issues/273.

PR #360, opened 2020-07-13
[Generic] Add support for client_secret_post method

enolfc:
Support the client_secret_post method as specificed at
https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication

Keeps the behaviour of the `basic_auth` config and adds a new config parameter `client_auth_method` to support `client_secret_post` and `client_secret_basic` (this is the currently existing one)

### DIFF
--- a/oauthenticator/generic.py
+++ b/oauthenticator/generic.py
@@ -68,6 +68,12 @@ class GenericOAuthenticator(OAuthenticator):
         help="Disable basic authentication for access token request",
     )
 
+    client_auth_method = Unicode(
+        os.environ.get('OAUTH2_AUTH_METHOD', 'client_secret_basic'),
+        config=True,
+        help="Authentication method to use. Supported methods: client_secret_basic, client_secret_post. Default: client_secret_basic",
+    )
+
     def http_client(self):
         return AsyncHTTPClient(force_instance=True, defaults=dict(validate_cert=self.tls_verify))
 
@@ -91,11 +97,13 @@ class GenericOAuthenticator(OAuthenticator):
 
         headers = {"Accept": "application/json", "User-Agent": "JupyterHub"}
 
-        if self.basic_auth:
+        if self.basic_auth and self.client_auth_method == "client_secret_basic":
             b64key = base64.b64encode(
                 bytes("{}:{}".format(self.client_id, self.client_secret), "utf8")
             )
             headers.update({"Authorization": "Basic {}".format(b64key.decode("utf8"))})
+        elif self.client_auth_method == "client_secret_post":
+            params.update(dict(client_id=self.client_id, client_secret=self.client_secret))
 
         req = HTTPRequest(
             url,


enolfc:
This is very similar to #274

oauthenticator/generic.py
```diff
@@ -91,11 +97,13 @@ async def authenticate(self, handler, data=None):
 
         headers = {"Accept": "application/json", "User-Agent": "JupyterHub"}
 
-        if self.basic_auth:
+        if self.basic_auth and self.client_auth_method == "client_secret_basic":
```
manics, line 100: According to https://tools.ietf.org/html/rfc6749#section-2.3.1
If a client is using `basic_auth` it should not be sending credentials in the body, so I think this should be just `if self.basic_auth:`

Alternatively we could deprecate `basic_auth` and replace it with `client_auth_method: 'basic' (default) | 'client_secret_post'`, or something like that?

@consideRatio @minrk any thoughts?

consideRatio: I think the `basic_auth` flag should be deprecated along with this, and that we transition to a system based on choosing `client_auth_method` where the allowed options are a subset of the OpenID specifications allowed options. Since they include more than things starting with `client_secret_` i suggest we don't abbreviate this to just `basic` but let it remain `client_secret_basic` to be more transparent and easier to connect to the specification we try to follow.

> If a client is using basic_auth it should not be sending credentials in the body, so I think this should be just if self.basic_auth:

With self.basic_auth deprecated in favor of self.client_auth_method, I figure `self.basic_auth` and `self.client_auth_method == "client_secret_basic"` can be asserted to be the same value in deprecation logic, and that we check only for `self.client_auth_method == "client_secret_basic"`

---

I think the hardest part for me working with this repo are the environment variables. I lack an overview on what environment variables do for libraries used. I'm currently afraid of removing it for example, because perhaps it influence something else, and I'm also afraid of introducing the new `OAUTH2_AUTH_METHOD` not knowing if it has meaning or not for another library.

```
    basic_auth = Bool(
        os.environ.get('OAUTH2_BASIC_AUTH', 'True').lower() in {'true', '1'},
        config=True,
        help="Disable basic authentication for access token request",
    )
```

manics: As far as I know several of the environment variables were required because they used to be the only way to set some properties, but @minrk did some work to refactor this so all properties can now be set as traitlets. I don't think environment variables are needed for new properties, and we could perhaps remove the existing env-vars in a future breaking release (but that's a separate discussion).

GeorgianaElena:
Thank you very much for this work @enolfc! This and the discussion in the meta issue has been really helpful in finding and fixing this bug ✨ 

I will close this PR now that the bug has been fixed as part of https://github.com/jupyterhub/oauthenticator/pull/568 and this refactor https://github.com/jupyterhub/oauthenticator/pull/526

PR #363, opened 2020-08-05
[OpenShifAuthenticator] Enable cert verification for self-signed certs and auto-load auth api URL

vpavlin:
The latest version of OpenShift OAuthenticator relies on user configuring `OPENSHIFT_URL` environment variable and then uses it for multiple kind of unrelated things, so it is not clear it `OPENSHIFT_URL` should be an API url, or OAuth API url.

Also, the default value `localhost:8443` would not work in OpenShift environment since in a container, localhost will not really do much.

This change sets more sensible default for `openshift_url` which actually represents internal URL for OpenShift API.

Next it automatically gathers the OpenShift OAuth URL, so that the user does not have to provide it externally via env vars.

Lastly it sets the OpenShift provided CA certs in a container by default, so that we don't have to turn off the verification on clusters using self-signed certs.

I had to use `requests` library to fetch the URLs for the OAuth issuer since the Async client did not work for me (event loop in an event loop error) - let me know if there is a better solution.

### DIFF
--- a/oauthenticator/openshift.py
+++ b/oauthenticator/openshift.py
@@ -7,12 +7,13 @@ Derived from the GitHub OAuth authenticator.
 
 import json
 import os
+import requests
 
 from tornado.auth import OAuth2Mixin
 from tornado import web
 
 from tornado.httputil import url_concat
-from tornado.httpclient import HTTPRequest, AsyncHTTPClient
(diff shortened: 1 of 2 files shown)


vpavlin:
Can anyone take a look, please?

minrk:
Thanks for the PR!

PR #364, opened 2020-08-12
lti 1.3 Authenticator based on illumidesk

eliaswimmer:
I have ported the code from https://github.com/IllumiDesk/illumidesk/tree/main/src/illumidesk/authenticators and integrated it into oauthenticator.

### DIFF
--- a/oauthenticator/lti13.py
+++ b/oauthenticator/lti13.py
@@ -1,0 +1,246 @@
+import os
+import logging
+import hashlib
+import re
+
+import json
+import jwt
+import time
+
+from josepy.jws import JWS
+from josepy.jws import Header
+
+from typing import Any, Dict, List, cast
+
+from traitlets.config import LoggingConfigurable
+from traitlets import Unicode, Bool, Union
+from .traitlets import Callable
+
+
+from .oauth2 import OAuthenticator
+from .oauth2 import guess_callback_uri
+from .oauth2 import OAuthLoginHandler
+from .oauth2 import OAuthCallbackHandler
+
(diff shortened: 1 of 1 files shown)


manics:
Hi! Thanks for your contribution.

We're not accepting new authenticator classes
https://github.com/jupyterhub/oauthenticator/blob/master/CONTRIBUTING.md

It's extremely difficult for us to maintain them since each authenticator requires manual testing with real authentication tokens, and this is too much of a burden for the maintainers. Instead we encourage everyone to use the [GenericAuthenticator](https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/generic.py) where possible, or to maintain your own package that depends on oauthenticator. We've previously done some work to make this easier.

eliaswimmer:
Hi Simon,

what if I will provide an integration test for my plugin?

best
Elias

manics:
An integration test helps prevent inadvertent breakages, but we'll still have the problem of long term maintenance:
- if someone reports a bug it's difficult for someone who's not familiar with the authenticator to fix it
- If someone opens a PR to update it or add a new feature it's difficult for us to review it
- It's not possible to release it on it's own, all authenticators on this repo have to be released together so it's not easy to release a "quick fix" if necessary

This means users of your authenticator will have a better experience if it *isn't* part of this repo. Please feel free to open a PR to add a link to it in the docs though. I hope this explains things a bit better!

eliaswimmer:
Thank you for your explanation. I will create my own repo and will open a new PR for a link in the docs.

Best wishes
Elias

PR #366, opened 2020-09-05
Rename OAuthenticator.whitelist to allow

GeorgianaElena:
- [x] deprecate traits, config
- [x] test deprecated names still work
- [x] test deprecated config
- [x] update docs

https://github.com/jupyterhub/team-compass/issues/299
https://github.com/jupyterhub/jupyterhub/pull/3090

### DIFF
--- a/oauthenticator/bitbucket.py
+++ b/oauthenticator/bitbucket.py
@@ -54,7 +54,6 @@ class BitbucketOAuthenticator(OAuthenticator):
         config=True, help="Automatically allow members of selected teams"
     )
 
-    bitbucket_allowed_teams = allowed_teams
 
     headers = {
         "Accept": "application/json",
--- a/oauthenticator/bitbucket.py
+++ b/oauthenticator/bitbucket.py
@@ -104,7 +104,7 @@ class BitbucketOAuthenticator(OAuthenticator):
 
         # Check if user is a member of any allowed teams.
         # This check is performed here, as the check requires `access_token`.
-        if self.bitbucket_allowed_teams:
+        if self.allowed_teams:
             user_in_team = await self._check_membership_allowed_teams(username, access_token)
             if not user_in_team:
                 self.log.warning("%s not in team allowed list of users", username)
--- a/oauthenticator/bitbucket.py
+++ b/oauthenticator/bitbucket.py
@@ -131,7 +131,7 @@ class BitbucketOAuthenticator(OAuthenticator):
 
             user_teams = set([entry["username"] for entry in resp_json["values"]])
             # check if any of the organizations seen thus far are in the allowed list
-            if len(self.bitbucket_allowed_teams & user_teams) > 0:
+            if len(self.allowed_teams & user_teams) > 0:
                 return True
         return False
 
--- a/oauthenticator/bitbucket.py
+++ b/oauthenticator/bitbucket.py
@@ -13,7 +13,7 @@ from tornado.httpclient import HTTPRequest, AsyncHTTPClient
 
 from jupyterhub.auth import LocalAuthenticator
 
-from traitlets import Set, default
+from traitlets import Set, default, observe
 
 from .oauth2 import OAuthLoginHandler, OAuthenticator
 
--- a/oauthenticator/bitbucket.py
+++ b/oauthenticator/bitbucket.py
@@ -28,6 +28,14 @@ def _api_headers(access_token):
 
 class BitbucketOAuthenticator(OAuthenticator):
 
+    _deprecated_aliases = {
+        "team_whitelist": ("allowed_teams", "0.12.0"),
+    }
+
+    @observe(*list(_deprecated_aliases))
+    def _deprecated_trait(self, change):
+        super()._deprecated_trait(change)
+
     login_service = "Bitbucket"
     client_id_env = 'BITBUCKET_CLIENT_ID'
     client_secret_env = 'BITBUCKET_CLIENT_SECRET'
@@ -40,11 +48,13 @@ class BitbucketOAuthenticator(OAuthenticator):
     def _token_url_default(self):
         return "https://bitbucket.org/site/oauth2/access_token"
 
-    team_whitelist = Set(
-        config=True, help="Automatically whitelist members of selected teams"
+    team_whitelist = Set(help="Deprecated, use `BitbucketOAuthenticator.allowed_teams`", config=True,)
+
+    allowed_teams = Set(
+        config=True, help="Automatically allow members of selected teams"
     )
 
-    bitbucket_team_whitelist = team_whitelist
+    bitbucket_allowed_teams = allowed_teams
 
     headers = {
         "Accept": "application/json",
@@ -93,12 +103,12 @@ class BitbucketOAuthenticator(OAuthenticator):
 
         username = resp_json["username"]
 
-        # Check if user is a member of any whitelisted teams.
+        # Check if user is a member of any allowed teams.
         # This check is performed here, as the check requires `access_token`.
-        if self.bitbucket_team_whitelist:
-            user_in_team = await self._check_team_whitelist(username, access_token)
+        if self.bitbucket_allowed_teams:
+            user_in_team = await self._check_membership_allowed_teams(username, access_token)
             if not user_in_team:
-                self.log.warning("%s not in team whitelist", username)
+                self.log.warning("%s not in team allowed list of users", username)
                 return None
 
         return {
@@ -106,7 +116,7 @@ class BitbucketOAuthenticator(OAuthenticator):
             'auth_state': {'access_token': access_token, 'bitbucket_user': resp_json},
         }
 
-    async def _check_team_whitelist(self, username, access_token):
+    async def _check_membership_allowed_teams(self, username, access_token):
         http_client = AsyncHTTPClient()
 
         headers = _api_headers(access_token)
@@ -121,8 +131,8 @@ class BitbucketOAuthenticator(OAuthenticator):
             next_page = resp_json.get('next', None)
 
             user_teams = set([entry["username"] for entry in resp_json["values"]])
-            # check if any of the organizations seen thus far are in whitelist
-            if len(self.bitbucket_team_whitelist & user_teams) > 0:
+            # check if any of the organizations seen thus far are in the allowed list
+            if len(self.bitbucket_allowed_teams & user_teams) > 0:
                 return True
         return False
 
--- a/oauthenticator/github.py
+++ b/oauthenticator/github.py
@@ -38,7 +38,7 @@ class GitHubOAuthenticator(OAuthenticator):
     # c.GitHubOAuthenticator.scope = ['read:org']
 
     _deprecated_aliases = {
-        "github_organization_whitelist": ("allowed_github_organizations", "0.12.0"),
+        "github_organization_whitelist": ("allowed_organizations", "0.12.0"),
     }
 
     @observe(*list(_deprecated_aliases))
@@ -116,7 +116,7 @@ class GitHubOAuthenticator(OAuthenticator):
     client_id_env = 'GITHUB_CLIENT_ID'
     client_secret_env = 'GITHUB_CLIENT_SECRET'
 
-    github_organization_whitelist = Set(help="Deprecated, use `GitHubOAuthenticator.allowed_github_organizations`", config=True,)
+    github_organization_whitelist = Set(help="Deprecated, use `GitHubOAuthenticator.allowed_organizations`", config=True,)
 
     allowed_organizations = Set(
         config=True, help="Automatically allow members of selected organizations"
--- a/oauthenticator/github.py
+++ b/oauthenticator/github.py
@@ -180,8 +180,8 @@ class GitHubOAuthenticator(OAuthenticator):
             return None
         # Check if user is a member of any allowed organizations.
         # This check is performed here, as it requires `access_token`.
-        if self.allowed_github_organizations:
-            for org in self.allowed_github_organizations:
+        if self.allowed_organizations:
+            for org in self.allowed_organizations:
                 user_in_org = await self._check_membership_allowed_organizations(
                     org, username, access_token
                 )
--- a/oauthenticator/github.py
+++ b/oauthenticator/github.py
@@ -17,7 +17,7 @@ from tornado.httpclient import HTTPRequest, AsyncHTTPClient, HTTPError
 
 from jupyterhub.auth import LocalAuthenticator
 
-from traitlets import List, Set, Unicode, default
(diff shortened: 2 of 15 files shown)


oauthenticator/github.py
```diff
@@ -168,17 +178,17 @@ async def authenticate(self, handler, data=None):
...
             return None
-        # Check if user is a member of any whitelisted organizations.
+        # Check if user is a member of any allowed organizations.
         # This check is performed here, as it requires `access_token`.
-        if self.github_organization_whitelist:
-            for org in self.github_organization_whitelist:
-                user_in_org = await self._check_organization_whitelist(
+        if self.allowed_github_organizations:
+            for org in self.allowed_github_organizations:
+                user_in_org = await self._check_memebership_allowed_organizations(
```
manics, line 185:
```suggestion
                user_in_org = await self._check_membership_allowed_organizations(
```

oauthenticator/github.py
```diff
@@ -197,7 +207,7 @@ async def authenticate(self, handler, data=None):
 
         return userdict
 
-    async def _check_organization_whitelist(self, org, username, access_token):
+    async def _check_memebership_allowed_organizations(self, org, username, access_token):
```
manics, line 210:
```suggestion
    async def _check_membership_allowed_organizations(self, org, username, access_token):
```

oauthenticator/gitlab.py
```diff
@@ -157,19 +171,19 @@ async def authenticate(self, handler, data=None):
...
-            user_in_group = await self._check_group_whitelist(user_id, access_token)
+            user_in_group = await self._check_membership_allowed_groups(user_id, access_token)
 
-        # We skip project_id check if user is in whitelisted group.
-        if self.gitlab_project_id_whitelist and not user_in_group:
+        # We skip project_id check if user is in allowed group.
+        if self.allowed_gitlab_project_ids and not user_in_group:
             is_project_id_specified = True
-            user_in_project = await self._check_project_id_whitelist(
+            user_in_project = await self._check_memebership_allowed_project_ids(
```
manics, line 186:
```suggestion
            user_in_project = await self._check_membership_allowed_project_ids(
```

oauthenticator/gitlab.py
```diff
@@ -219,11 +233,11 @@ async def _check_group_whitelist(self, user_id, access_token):
                 return True  # user _is_ in group
         return False
 
-    async def _check_project_id_whitelist(self, user_id, access_token):
+    async def _check_memebership_allowed_project_ids(self, user_id, access_token):
```
manics, line 236:
```suggestion
    async def _check_membership_allowed_project_ids(self, user_id, access_token):
```

manics (commented):
Thanks for opening this. How thorough do you think the unit tests are? Are they enough to verify the authenticators all work as before, or does it require configuring each of the authenticators with real credentials and testing manually?

oauthenticator/tests/test_google.py
```diff
@@ -72,37 +75,55 @@ async def test_admin_google_groups(google_client):
...
+    handler = google_client.handler_for_user(user_model('fakealloweduser@email.com'))
+    allowed_user_info = await authenticator.authenticate(handler, google_groups=['anotherone', 'fakegroup'])
+    allowed_user_groups = allowed_user_info['auth_state']['google_user']['google_groups']
+    admin_field = allowed_user_info.get('admin')
+    assert 'fakegroup' in allowed_user_groups
     assert admin_field is None
-    handler = google_client.handler_for_user(user_model('fakenonwhitelisted@email.com'))
-    whitelisted_user_groups = await authenticator.authenticate(handler, google_groups=['anotherone', 'fakenonwhitelistedgroup'])
-    assert whitelisted_user_groups is None
+    handler = google_client.handler_for_user(user_model('faallowedisted@email.com'))
```
consideRatio, line 109: A code suggestion based on mirroring the logic deleted rather than understanding it properly. I assume this string doesn't matter though.

```suggestion
    handler = google_client.handler_for_user(user_model('fakenonalloweduser@email.com'))
```

oauthenticator/tests/test_google.py
```diff
@@ -72,37 +75,55 @@ async def test_admin_google_groups(google_client):
...
+    assert caplog.record_tuples == [
+        (
+            log.name,
+            logging.WARNING,
+            'GoogleOAuthenticator.google_group_whitelist is deprecated in GoogleOAuthenticator 0.12.0, use '
+            'GoogleOAuthenticator.allowed_google_groups instead',
+        )
+    ]
+
+    assert authenticator.allowed_google_groups == {'email.com': ['group']}
```
consideRatio, line 129:
```suggestion
    assert authenticator.allowed_google_groups == {'email.com': ['group']}
 
```

consideRatio (commented):
Awesome work @GeorgianaElena! :tada: :heart:

This looks great to me! It seems like this will cover various deprecated configuration options specific to OAuthenticator derived classes, but also updates documentation of the deprecation made in the base Authenticator class.

The logic is defined in `oauth2.py` where the OAuthenticator base class is, but reused in the derived classes for their various deprecated configurations are. Nice call!

From what I can tell, all of these deprecated configurations are tested to both emit a warning if they are used and to still function as the new configuration!

## Not in this PR, but in https://github.com/jupyterhub/jupyterhub/pull/3090
`Authenticator.whitelist` -> `Authenticator.allowed_users`

## In this PR
- `BitbucketOAuthenticator.team_whitelist` -> `BitbucketOAuthenticator.bitbucket_allowed_teams`
- `CILogonOAuthenticator.idp_whitelist` -> `CILogonOAuthenticator.allowed_idps`
- `GitHubOAuthenticator.github_organization_whitelist` -> `GitHubOAuthenticator.allowed_github_organizations`
- `GitLabOAuthenticator.gitlab_group_whitelist` -> `GitLabOAuthenticator.allowed_gitlab_groups`
- `GitLabOAuthenticator.gitlab_project_id_whitelist` -> `GitLabOAuthenticator.allowed_gitlab_project_ids`

## Discussion

I think the documentation is now referring to the new JupyterHub configuration which isn't yet released. What do you think we should do about this? Should we add a note saying that this assumes JupyterHub `1.2`? I'm not sure.

oauthenticator/tests/test_google.py
```diff
@@ -72,37 +75,55 @@ async def test_admin_google_groups(google_client):
...
```
GeorgianaElena, line 129:
```suggestion
    assert authenticator.allowed_google_groups == {'email.com': ['group']}

```

GeorgianaElena:
> Thanks for opening this. How thorough do you think the unit tests are? Are they enough to verify the authenticators all work as before, or does it require configuring each of the authenticators with real credentials and testing manually?

To be honest, I didn't think about this and kind of assumed that the tests passing was enough confirmation that I didn't broke any functionality, without paying attention on how thorough they are :thinking: 
Looking at them though, it looks like the oauth clients are all mocked and the `authenticate` method` is called. Maybe it's worth testing the authentication manually too. At least for some of them.
I will do that and come with updates. Thanks @manics

GeorgianaElena:
> I think the documentation is now referring to the new JupyterHub configuration which isn't yet released. What do you think we should do about this? Should we add a note saying that this assumes JupyterHub 1.2? I'm not sure.

Maybe add something like: "This assumes JupyterHub 1.2 For JupyterHub < 1.2, check this PR (or the changelog when we'll have one) about latest deprecations"?

What do you think about adding this note to every doc section that reference the JH 1.2 `allowed_users`? Should we add it in a "central" place instead?

oauthenticator/bitbucket.py
```diff
@@ -40,11 +48,13 @@ def _authorize_url_default(self):
...
-    team_whitelist = Set(
-        config=True, help="Automatically whitelist members of selected teams"
+    team_whitelist = Set(help="Deprecated, use `BitbucketOAuthenticator.allowed_teams`", config=True,)
+
+    allowed_teams = Set(
+        config=True, help="Automatically allow members of selected teams"
     )
 
-    bitbucket_team_whitelist = team_whitelist
+    bitbucket_allowed_teams = allowed_teams
```
minrk, line 57: We don't need to keep the deprecated `bitbucket_` prefix alias for the new trait

```suggestion
```

oauthenticator/bitbucket.py
```diff
@@ -93,20 +103,20 @@ async def authenticate(self, handler, data=None):
 
         username = resp_json["username"]
 
-        # Check if user is a member of any whitelisted teams.
+        # Check if user is a member of any allowed teams.
         # This check is performed here, as the check requires `access_token`.
-        if self.bitbucket_team_whitelist:
-            user_in_team = await self._check_team_whitelist(username, access_token)
+        if self.bitbucket_allowed_teams:
```
minrk, line 108:
```suggestion
        if self.allowed_teams:
```

oauthenticator/bitbucket.py
```diff
@@ -121,8 +131,8 @@ async def _check_team_whitelist(self, username, access_token):
             next_page = resp_json.get('next', None)
 
             user_teams = set([entry["username"] for entry in resp_json["values"]])
-            # check if any of the organizations seen thus far are in whitelist
-            if len(self.bitbucket_team_whitelist & user_teams) > 0:
+            # check if any of the organizations seen thus far are in the allowed list
+            if len(self.bitbucket_allowed_teams & user_teams) > 0:
```
minrk, line 135:
```suggestion
            if len(self.allowed_teams & user_teams) > 0:
```

oauthenticator/github.py
```diff
@@ -108,8 +116,10 @@ def _github_client_secret_changed(self, name, old, new):
     client_id_env = 'GITHUB_CLIENT_ID'
     client_secret_env = 'GITHUB_CLIENT_SECRET'
 
-    github_organization_whitelist = Set(
-        config=True, help="Automatically whitelist members of selected organizations"
+    github_organization_whitelist = Set(help="Deprecated, use `GitHubOAuthenticator.allowed_github_organizations`", config=True,)
+
+    allowed_github_organizations = Set(
```
minrk, line 121: I've been trying to keep reduntant `github`, `bitbucket` names out of the GitHub and Bitbucket traits:

```suggestion
    allowed_organizations = Set(
```

oauthenticator/github.py
```diff
@@ -168,17 +178,17 @@ async def authenticate(self, handler, data=None):
...
         if not username:
             return None
-        # Check if user is a member of any whitelisted organizations.
+        # Check if user is a member of any allowed organizations.
         # This check is performed here, as it requires `access_token`.
-        if self.github_organization_whitelist:
-            for org in self.github_organization_whitelist:
-                user_in_org = await self._check_organization_whitelist(
+        if self.allowed_github_organizations:
+            for org in self.allowed_github_organizations:
```
minrk, line 184:
```suggestion
        if self.allowed_organizations:
            for org in self.allowed_organizations:
```

oauthenticator/gitlab.py
```diff
@@ -37,6 +37,15 @@ class GitLabOAuthenticator(OAuthenticator):
     # set scopes via config, e.g.
     # c.GitLabOAuthenticator.scope = ['read_user']
 
+    _deprecated_aliases = {
+        "gitlab_group_whitelist": ("allowed_gitlab_groups", "0.12.0"),
+        "gitlab_project_id_whitelist": ("allowed_gitlab_project_ids", "0.12.0")
```
minrk, line 42:
```suggestion
        "gitlab_group_whitelist": ("allowed_gitlab_groups", "0.12.0"),
        "gitlab_project_id_whitelist": ("allowed_project_ids", "0.12.0")
```

oauthenticator/gitlab.py
```diff
@@ -93,12 +102,17 @@ def _authorize_url_default(self):
...
+    gitlab_group_whitelist = Set(help="Deprecated, use `GitLabOAuthenticator.allowed_gitlab_groups`", config=True,)
+
+    allowed_gitlab_groups = Set(
+        config=True, help="Automatically allow members of selected groups"
     )
-    gitlab_project_id_whitelist = Set(
+
+    gitlab_project_id_whitelist = Set(help="Deprecated, use `GitLabOAuthenticator.allowed_gitlab_project_ids`", config=True,)
+
+    allowed_gitlab_project_ids = Set(
```
minrk, line 113:
```suggestion
    gitlab_project_id_whitelist = Set(help="Deprecated, use `GitLabOAuthenticator.allowed_project_ids`", config=True,)

    allowed_project_ids = Set(
```

oauthenticator/gitlab.py
```diff
@@ -157,19 +171,19 @@ async def authenticate(self, handler, data=None):
...
-        if self.gitlab_group_whitelist:
+        if self.allowed_gitlab_groups:
             is_group_specified = True
-            user_in_group = await self._check_group_whitelist(user_id, access_token)
+            user_in_group = await self._check_membership_allowed_groups(user_id, access_token)
 
-        # We skip project_id check if user is in whitelisted group.
-        if self.gitlab_project_id_whitelist and not user_in_group:
+        # We skip project_id check if user is in allowed group.
+        if self.allowed_gitlab_project_ids and not user_in_group:
```
minrk, line 184:
```suggestion
        if self.allowed_project_ids and not user_in_group:
```

oauthenticator/gitlab.py
```diff
@@ -219,11 +233,11 @@ async def _check_group_whitelist(self, user_id, access_token):
...
         return False
 
-    async def _check_project_id_whitelist(self, user_id, access_token):
+    async def _check_membership_allowed_project_ids(self, user_id, access_token):
         http_client = AsyncHTTPClient()
         headers = _api_headers(access_token)
-        # Check if user has developer access to any project in the whitelist
-        for project in self.gitlab_project_id_whitelist:
+        # Check if user has developer access to any project in the allowed list
+        for project in self.allowed_gitlab_project_ids:
```
minrk, line 240:
```suggestion
        for project in self.allowed_project_ids:
```

minrk (commented):
Great! Thanks for tackling this.

I left a few comments where I'd like to avoid using redundant trait names where unambiguous, e.g. `GitHubOAuthenticator.allowed_github_organizations` -> `GitHubOAuthenticator.allowed_organizations`, but otherwise 👍!

minrk:
This PR should be safe, because it's mostly orthogonal to the changes in JupyterHub. OAuthenticator implements its own handling of the traits defined in OAuthenticator, while JupyterHub implements the handling of `whitelist->allowed_users`. The way to verify this is to check for references to `self.whitelist` or `self.group_whitelist` in this repo. There aren't any, so this should be independent of the jupyterhub changes, and fully compatible with jupyterhub 1.1 and 1.2.

minrk:
Excellent!

Issue #335, opened 2020-02-28
v.11 returns v1 token while v.10 returns v2


scivm:
I tried upgrading from  0.9.0-beta.3 to 0.9.0-beta.4 but had an issue that with the AD token.  In .4 I get a v1 token while in .3 I am getting a v2 token.   

The oauthenticator code is calling the older v1 endpoint in both the old .10 and new .11 azuread oauthenicator.

old v1 endpoint:
```
https://login.microsoftonline.com/{0}/oauth2/authorize
https://login.microsoftonline.com/{0}/oauth2/token
```

For some reason, the .3 with oauthenticator .10 returns a v2 token while .4 with oauthenticator .11 returns a v1 token.

The new v2 urls are:
```
https://login.microsoftonline.com/{0}/oauth2/v2.0/authorize
https://login.microsoftonline.com/{0}/oauth2/v2.0/token
```

```
auth:
    state:
       enabled: true
       cryptoKey: "xxxxx"
    admin:
      users:
        - shamberger michael (ext)
    type: custom
    custom:
      className: "CustomAzureAdOAuthenticator"
      config:
        client_id: "xxxx"
        client_secret: "xxxxx"
        oauth_callback_url: "https://xxxxx.cloudapp.azure.com/hub/oauth_callback"
        tenant_id: "xxxxx"
```

```
    authpasstoken: |
      from oauthenticator.azuread import AzureAdOAuthenticator
      from tornado import gen
      class CustomAzureAdOAuthenticator(AzureAdOAuthenticator):
        @gen.coroutine
        def pre_spawn_start(self, user, spawner):
          auth_state = yield user.get_auth_state()
          if not auth_state:
            # user has no auth state
            return
          # define some environment variables from auth_state
          spawner.environment['AD_TOKEN'] = auth_state['access_token']
      c.JupyterHub.authenticator_class = CustomAzureAdOAuthenticator
```

**0.9.0-beta.3**
```
[I 2020-02-28 13:02:17.669 JupyterHub pages:347] shamberger michael (ext) is pending spawn
[I 2020-02-28 13:02:17.670 JupyterHub log:174] 200 GET /hub/spawn-pending/shamberger%20michael%20%28ext%29 (shamberger michael (ext)@10.0.1.4) 4.24ms
[I 2020-02-28 13:02:17.671 JupyterHub spawner:1769] PVC claim-shamberger-20michael-20-28ext-29 already exists, so did not create new pvc.
[...]

PR #373, opened 2020-10-12
[AzureAD] Add (optional) additional request to get username attribute

iwilltry42:
Hi there,
this PR adds the ability to do an additional second request against AzureAD (more specifically, the Microsoft Graph API) to fetch the username attribute defined by `username_claim`, if it isn't present in the default OAuth reply.

Example use case that we hit: 
- We're syncing an on-premises AD with AzureAD and heavily rely on the (old) short usernames, which are in the `sAMAccountName` field in AD and in the `onPremisesSamAccountName` in AzureAD.
- The field `onPremisesSamAccountName` is not present in the small default reply of the Login requests (/authorize and /token on login.microsoftonline.com).
- This results in a 500 error on JupyterHub (`KeyError` in the OAuthenticator)

I guess there's still room for improvement (or you prefer to rather drop this in a GenericOAuthenticator), but I'd like to get some feedback on my first iteration :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,5 @@ jupyterhub_cookie_secret
 
 .vscode/
 .idea/
+
+Pipfile*
--- a/examples/azuread/sample_jupyter_config.py
+++ b/examples/azuread/sample_jupyter_config.py
@@ -1,5 +1,7 @@
 import os
+
 from oauthenticator.azuread import AzureAdOAuthenticator
+
 c.JupyterHub.authenticator_class = AzureAdOAuthenticator
 
 c.Application.log_level = 'DEBUG'
@@ -14,3 +16,8 @@ c.AzureAdOAuthenticator.client_secret = '{AAD-APP-CLIENT-SECRET}'
 # uncomment the line below to use 'unique_name' rather than the default 'name'. Consult the Azure
 # documentation for other field names.
 #c.AzureAdOAuthenticator.username_claim = 'unique_name'
+#
+# If the alternate field defined by username_claim is not part of the default OAuth request,
+# we can do an additional lookup request to the Microsoft Graph API to fetch it. You can enable it
+# by uncommenting the settings below (if needed)
+#c.AzureAdOAuthenticator.username_claim_request = True
--- a/oauthenticator/azuread.py
+++ b/oauthenticator/azuread.py
@@ -2,17 +2,20 @@
 Custom Authenticator to use Azure AD with JupyterHub
 """
 
+import json
 import os
 import urllib
 from distutils.version import LooseVersion as V
 
 import jwt
 from jupyterhub.auth import LocalAuthenticator
-from tornado.httpclient import HTTPRequest
+from tornado.httpclient import AsyncHTTPClient, HTTPRequest
+from tornado.log import app_log
+
 from traitlets import Unicode, default
+from traitlets.traitlets import Bool
 
 from .oauth2 import OAuthenticator
-
 
 # pyjwt 2.0 has changed its signature,
 # but mwoauth pins to pyjwt 1.x
@@ -32,11 +35,17 @@ class AzureAdOAuthenticator(OAuthenticator):
     def _tenant_id_default(self):
         return os.environ.get('AAD_TENANT_ID', '')
 
-    username_claim = Unicode(config=True)
+    username_claim = Unicode(config=True, help="User's attribute to return as username")
 
     @default('username_claim')
     def _username_claim_default(self):
         return 'name'
+
+    username_claim_request = Bool(config=True, help="Allow an additional user attribute request if username_claim is not in the initial response")
+
+    @default('username_claim_request')
+    def _username_claim_request_default(self):
+        return False
 
     @default("authorize_url")
     def _authorize_url_default(self):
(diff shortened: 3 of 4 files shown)


manics:
Could other users of this authenticator please help review this? In particular:
- does the change make sense?
- is it sufficiently general enough that it should be included in this repository?
- does it cover all reasonable/related use cases?
- is it possible to break any existing uses?

jabbera:
I don't think this is necessary. You can tell Azure AD to include the samAccountName in the token per this documentation: 
https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-optional-claims

iwilltry42:
@jabbera , oh I didn't know about this, will give it a try and close the PR, if it works :+1:

manics:
@jabbera Thanks for the info! Would you or @iwilltry42 mind updating the documentation to help others?

dhirschfeld:
> You can tell Azure AD to include the samAccountName in the token per this documentation:

That only seems to refer to groups? I'm also trying to figure out the right `username_claim` to use the `sAMAccountName` but have so far been unsuccessful :/

dhirschfeld:
@jabbera is correct that this can be achieved without having to make a call to the graph api by configuring the JupyterHub Service Principal; however it's not straight-forward.

AFAICS there is no way to request the `sAMAccountName` in the portal as it's not part of the standard claims. You can however configure the JupyterHub Service Principal to return non-standard claims such as  `sAMAccountName`.

First you need to install the `AzureADPreview` PowerShell module (which only works in old PowerShell, not PowerShell Core):
```powershell
Install-Module -Name AzureADPreview
Import-Module AzureADPreview
Connect-AzureAD -Confirm
```
...then you create an `AzureADPolicy` to return the `sAMAccountName`:
```powershell
$policyTemplate = @"
{
    "ClaimsMappingPolicy": {
        "Version": 1,
        "IncludeBasicClaimSet": "true",
        "ClaimsSchema": [
            {
                "Source": "user",
                "ID": "onpremisessamaccountname",
                "JwtClaimType": "samid"
            }
        ]
    }
}
"@
$policy = New-AzureADPolicy -Type "ClaimsMappingPolicy" -Definition ($policyTemplate) -DisplayName "sAMAccountName"
```
Finally, you assign that policy to your JupyterHub Service Principal:
```powershell
$servicePrincipal = Get-AzureADServicePrincipal -SearchString JupyterHub
Add-AzureADServicePrincipalPolicy -Id $servicePrincipal.ObjectId -RefObjectId $policy.Id
```

The final thing you need to do is to set `"acceptMappedClaims": true` in the application manifest:

image

```python
 In[4]: claims.keys()
Out[4]: dict_keys(['aud', 'iss', 'iat', 'nbf', 'exp', 'aio', 'amr', 'email', 'family_name', 'given_name', 'groups', 'ipaddr', 'name', 'oid', 'onprem_sid', 'preferred_username', 'rh', 'sub', 'tid', 'unique_name', 'upn', 'uti', 'ver', 'samid'])

 In[5]: claims['samid']
Out[5]: 'dhirschfeld'
```

----
***References:***
* https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-claims-mapping
* https://goodworkaround.com/2019/11/15/adding-custom-attributes-to-the-azure-ad-openid-connect-id-token/

jabbera:
@dhirschfeld how did you get this to come back in the access token? In postman I have to request offline access to get an ID token.

jabbera:
@dhirschfeld never mind. I had to switch over to the v2 endpoints. That should really be the default. I killed so much time trying to figure this out.

dhirschfeld:
> *I had to switch over to the v2 endpoints.*

Over in https://github.com/jupyterhub/oauthenticator/issues/335#issuecomment-814891321 I did some experimentation with manually patching to use v2.0 urls however I found that it appears you can use the v1.0 endpoints and request v2.0 tokens by setting `"accessTokenAcceptedVersion: 2"` in the manifest.

image

iwilltry42:
Hi @dhirschfeld & @jabbera , thanks for your input!
That way would be way simpler and we could stick with upstream oauthenticator :)
However, I just fail to get this to work... even with all your changes mentioned in https://github.com/jupyterhub/oauthenticator/pull/373#issuecomment-815463677 and https://github.com/jupyterhub/oauthenticator/pull/373#issuecomment-816240991 the token returned does not contain the `samid`.
One thing to note here though is that we had to switch to using the `access_token` from `v2.0` instead of the `id_token` as before, because that one gave me errors.

jabbera:
@iwilltry42 the custom claims only come back in the ID token apparently. I had to use the v2 endpoint to get it. Work it out in postman first.

dhirschfeld:
> I had to use the v2 endpoint to get it.

I'm pretty sure I'm running a patched container using the v2 endpoints.

iwilltry42:
Hi @jabbera & @dhirschfeld , thanks for your input!
However, we have everything setup on AzureAD site as mentioned inhttps://github.com/jupyterhub/oauthenticator/pull/373#issuecomment-815463677 and https://github.com/jupyterhub/oauthenticator/pull/373#issuecomment-816240991, but still can't get the optional claim.
When using the default https://login.microsoftonline.com/{0}/oauth2/token, the optional claim (`samid`) is not included in the returned token.
When using the v2.0 endpoint (https://login.microsoftonline.com/{0}/oauth2/v2.0/token), the response only includes an `access_token`, but not the `id_token` (and the `access_token` also does not include the extra claim) :thinking:

jabbera:
@iwilltry42 

Unsure if it will help but here is a summarized version of the config settings I'm using from my config values:

```
      client_id: "SET FROM PIPELINE"
      tenant_id: "SET FROM PIPELINE"
      username_claim: samAccountName
      oauth_callback_url: "SET FROM PIPELINE"
      authorize_url: https://login.microsoftonline.com/SET FROM PIPELINE/oauth2/v2.0/authorize
      token_url: https://login.microsoftonline.com/SET FROM PIPELINE/oauth2/v2.0/token
      scope:
        - User.Read
        - profile
        - openid
        - email
    JupyterHub:
      admin_access: true
      authenticator_class: azuread
```

consideRatio:
@iwilltry42 hi and thanks for your work! I've not been so active maintaining this repo but am looking to help out. If you want to make a push to get this merged, I'll iterate quickly with you going onwards.

Action points:
- Can you rebase this on current master branch? By doing so you can solve merge conflicts and we can avoid getting the annotations from code coverage that can't be deleted but have since last been disabled to not disrupt like this.
- Can you update the PR description to reflect the current state of things or declare that it is up to date?

iwilltry42:
Hi @consideRatio , thanks for chiming in and sorry for not getting back to this lately.
It seems that according to the other comments, this feature is not needed after all (though I couldn't test it so far unfortunately). Thanks for the input @dhirschfeld & @jabbera :pray: 
So I'll go ahead and close the PR :+1: (If anyone else thinks this may be useful in the future, I can invest some time into it again)

consideRatio:
Ah, thank you for the followup @iwilltry42!! Btw when I see your name I also become thankful about k3d and k3s :) Thank you for your contributions to open source!

iwilltry42:
@consideRatio thanks for the kind words 🙂
I can only give that back and I'm especially thankful for your work in the Jupyter ecosystem, which we use quite heavily in my team at trivago 🙂

Issue #353, opened 2020-05-15
[Generic] Refactoring the authenticate function to make it easier to replace parts of it


mhaley-miovision:
### Proposed change

Hello, I have recently setup Jupyterlab using the generic oauth authenticator and AWS Cognito.  Everything works exactly as expected, and it was quite easy to get going, which was really nice.  My next step was to get the idtoken that cognito returns that you need to request temporary AWS credentials exposed inside the user's workspace some how.  The problem is that both the old [aws cognito authenticator](https://github.com/jupyterhub/oauthenticator/blob/0.10.0/oauthenticator/awscognito.py) and the [generic authenticator](https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/generic.py#L150) (which I was using) do not return the idtoken as authentication state.  To get around this I basically had to copy the authenticate method from the generic authenticator in a subclass just to get one extra value out of the response.  Once I was able to get the actual idtoken into the user's auth_state I overrode the pre-spawn and a post-spawn methods to create a k8s secret with that value to feed it into the container for use inside a user's environment.  What are people's thoughts for extracting additional information from the token response of the oauth2 workflow?  This seems like alot of work and effort to get one extra value.

Other Relates Issues:
https://github.com/jupyterhub/oauthenticator/issues/327

### Alternative options

For a bit I considered creating an extension in Jupyterhub (so inside the user's workspace) to run the authentication flow and setup all the user's personal secrets, but this route would be more annoying for the user and end up being harder to maintain moving forward.

### Who would use this feature?

Anyone looking to use AWS cognito with jupyterlab on AWS and wants to provide user specific AWS credentials inside of notebooks.

### (Optional): Suggest a solution

I see these two options:
1. Bring back the AWSCognito specific authenticator that saves the extra cognito specific information, such as the idtoken, inside the user's auth_state so it is easily accessible.
2. Split up the single authentication method that does multiple steps into smaller functions and provide additional hooks into the authentication flow so that people can inject themselves more easily into the flow with simple configuration.  This would make subclassing the generic authenticator easier, or not even needed in most cases.

manics:
The main reason we're not adding more authenticators to this repository is maintainability. The problem with OAuth is it's almost impossible to automate the testing. This means it's a large manual burden on the current maintainers, and it's made even harder by needing to acquire accounts and tokens for testing.

[...]

PR #374, opened 2020-10-14
[Generic] Break apart authenticate method for better composability

mhaley-miovision:
This is not tested, it is just to give an idea of how to break this method up.  See the below issue for more context:
https://github.com/jupyterhub/oauthenticator/issues/353#issuecomment-708386168


manics:
Rebased as https://github.com/jupyterhub/oauthenticator/pull/389

PR #376, opened 2020-10-25
[Generic] Remove userdata_method configuration supposedly not relevant

consideRatio:
A final piece discussed to be deprecated in #197 which I think should be closed no matter what though.

The idea is that the `userdata_method` is always going to be `GET`, but I'm not 100% that is true. I'm not sure how to get knowledge if this is a relevant configuration option any more though, it is like needing to proove there is no black sheeps, its hard.


minrk:
Rebased and merged

PR #378, opened 2020-10-26
change to BaseHandler._set_cookie to do set cookie

### LINKED ISSUE #277: use baseHandler to set cookie
Currently oauth.py directly calls the tornado set_secure_cookie when setting the state cookie here:
https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/oauth2.py#L67

This bypasses a bunch of logic in the JupyterHub BaseHandler to handle things like custom cookie options and handling some other edge cases. 
See:
https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/handlers/base.py#L488

The example I had in mind was setting "secure: True" via cookie_options to force secure cookies in an SSL terminated environment, but there are plenty of other use cases.

It might make sense to just call BaseHandler._set_cookie instead. If this sounds like the right thing to do, I can submit a PR.

Wh1isper:
fix #277
if you can't update, there is a way to fix it in config file 
```
from oauthenticator.oauth2 import OAuthLoginHandler, STATE_COOKIE_NAME
class SecureOAuthLoginHandler(OAuthLoginHandler):
      def set_state_cookie(self, state):
          self._set_cookie(STATE_COOKIE_NAME, state, expires_days=1, httponly=True)
class SecureOAuthenticator(GenericOAuthenticator):
      login_handler = SecureOAuthLoginHandler
c.JupyterHub.authenticator_class = SecureOAuthenticator
```

### DIFF
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -73,7 +73,7 @@ class OAuthLoginHandler(OAuth2Mixin, BaseHandler):
         return self.authenticator.userdata_url
 
     def set_state_cookie(self, state):
-        self.set_secure_cookie(STATE_COOKIE_NAME, state, expires_days=1, httponly=True)
+        self._set_cookie(STATE_COOKIE_NAME, state, expires_days=1, httponly=True)
 
     _state = None
 


minrk:
My only qualm about this is the use of a private internal API of JupyterHub, so it could break without notice. But the absence of support for setting cookies with cookie options as desired in #277.

So I'm happy to merge this, but it would be also great if we can open an issue on jupyterhub/jupyterhub to request a documented API for setting cookies.

consideRatio:
> it would be also great if we can open an issue on jupyterhub/jupyterhub to request a documented API for setting cookies.

@minrk I created https://github.com/jupyterhub/jupyterhub/issues/3224

PR #383, opened 2020-11-11
migrate from travis to github actions

### LINKED ISSUE #381: Migrate Travis CI to GitHub actions

minrk:
also removes circle build of docs, since we have rtd set up for pull requests

closes #381


minrk:
Results can be seen at https://github.com/minrk/oauthenticator/actions since we can't see action results on PRs until the first action PR is merged.

This includes a generic action for publishing releases from tags that should work for ~any Python package. It always runs to make sure building releases isn't broken, but only publishes on tags

.github/workflows/test.yml
```diff
@@ -0,0 +1,97 @@
...
+
+    runs-on: ubuntu-20.04
+    timeout-minutes: 10
+
+    strategy:
+      # Keep running even if one variation of the job fail
+      fail-fast: false
+      matrix:
+        python:
+          - "3.5"
```
consideRatio, line 59: I see that we are not testing py35 currently, but have not updated setup.py yet to require py36, so I figure this is fine to add but I think we should require py36 from now on.

consideRatio:
LGTM! Thanks @minrk!

.github/workflows/test.yml
consideRatio: I opened #384

oauthenticator/tests/test_mediawiki.py
```diff
@@ -27,7 +27,7 @@ def post_token(request, context):
                         'iss': 'https://meta.wikimedia.org',
                         'iat': time.time(),
                         'nonce': request_nonce,
-                    }, 'client_secret')
+                    }, 'client_secret').encode("utf8")
```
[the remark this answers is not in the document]
minrk: This should be unrelated to the LANG env, which Python does not generally take into account for encoding except when producing terminal output. The error here was a TypeError because str is not allowed at all, only None or Bytes, so which encoding is used is not relevant. I think this was just a change in the requests_mock or requests version being strict about types.

PR #389, opened 2020-11-21
[Generic] Refactoring the authenticate function to make it easier to replace parts of it (updated #353)

### LINKED ISSUE #353: [Generic] Refactoring the authenticate function to make it easier to replace parts of it

manics:
@mhaley-miovision Thanks for opening https://github.com/jupyterhub/oauthenticator/issues/353
I think splitting up `authenticate()` into smaller methods makes sense.

I've rebased your PR, and fixed up some errors so now the tests are passing.

Closes #353

### DIFF
--- a/oauthenticator/generic.py
+++ b/oauthenticator/generic.py
@@ -81,7 +81,7 @@ class GenericOAuthenticator(OAuthenticator):
             headers.update({"Authorization": "Basic {}".format(b64key.decode("utf8"))})
         return headers
 
-    def _get_token(self, http_client, headers, params):
+    async def _get_token(self, http_client, headers, params):
         if self.token_url:
             url = self.token_url
         else:
@@ -97,7 +97,7 @@ class GenericOAuthenticator(OAuthenticator):
         resp = await http_client.fetch(req)
         return json.loads(resp.body.decode('utf8', 'replace'))
 
-    def _get_user_data(self, http_client, token_response):
+    async def _get_user_data(self, http_client, token_response):
         access_token = token_response['access_token']
         token_type = token_response['token_type']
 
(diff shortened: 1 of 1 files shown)


consideRatio:
I think this looks great! I love it! Thank you for your work @mhaley-miovision and @manics !

PR #393, opened 2020-12-04
Return refresh and id tokens from authenticate

biomath-vlad:
Custom client libraries subclassed from this authenticator will benefit from receiving not only the access_token but also refresh_token and id_token.  These token are usually provided by callback from Auth0 service.


minrk:
Seems sensible, thanks!

PR #395, opened 2020-12-04
added allowed_groups and admin_groups to generic.py

mcmartins:
Similarly to what is done for `gitlab` and `google` authenticators, this pull request introduces `allowed_groups` and `admin_groups` into the `generic` authenticator.

In `Keycloak` (this mention is solely because `Keycloak` is the IDM tool we use) one can map user roles or policies from a source, e.g. Active Directory, configure the client scopes to include this information within the id token and access token as claims. This allows applications to obtain information that can be used to authorize access to functionality within that application.

This functionality is supported by all major IDM tools, e.g. Auth0, Okta, ForgeRock, etc, and is quite common. I do believe this is useful functionality to be supported by this authenticator.

I tried to keep things generic, wrote some tests and some documentation on how to configure the `GenericOAuthenticator`. Plus everything works as before, if no `allowed_groups` are specified.

Thanks,
Manuel

### DIFF
--- a/examples/generic/README.md
+++ b/examples/generic/README.md
@@ -1,0 +1,68 @@
+# OAuthenticator
+
+Example of `GenericOAuthenticator` using an IDM - identity and access management - solution (keycloak).
+
+## Keycloak
+
+`Keycloak` is an open source IDM solution. 
+In `Keycloak` one can map user roles or policies from a source, 
+e.g. Active Directory, configure the client scopes to include this information
+within the id token and access token as claims.
+
+This allows applications to obtain information that can be used to authorize
+access to functionality within that application.
+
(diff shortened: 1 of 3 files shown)


DavidFair:
Is there any plans to review / merge this PR? I'm trying to avoid writing the equivalent of this in a custom authenticator class for a hub instance.

I'm happy to help by testing against a bespoke OAuth provider so we know it works generically if that would help the review / merge process?

minrk:
This is terrific and well documented and tested. Thank you!

consideRatio:
Wieeee thank you @mcmartins for the thorough work on this, @DavidFair for your attention to it and @minrk for the review! :heart: :tada:

PR #401, opened 2020-12-23
Support username_claim in Google OAuth

dtaniwaki:
I'd like to use unique ID instead of email for Google OAuth because it contains `@` and `.` which can be non-unique even w/ https://github.com/jupyterhub/jupyterhub/pull/3316.

Here's the payload of the Google OAuth token: https://developers.google.com/identity/protocols/oauth2/openid-connect


minrk:
Thanks!

PR #402, opened 2021-01-07
PyJWT 2.0 compliant

rragundez:
PyJWT deprecated the argument `verify` and made it part of options as `verify_signature`. This PR solves that bug, basically the same as this one: https://github.com/Clinical-Genomics/cgbeacon2/issues/116

you will get:
jwt.exceptions.DecodeError: It is required that you pass in a value for the "algorithms" argument when calling decode().


yuvipanda:
Thank you very much, @rragundez!

rragundez:
No worries @yuvipanda , when can we expect a release with the patch? then I can remove the workaround in our JH + AD deployment.

PR #404, opened 2021-01-24
Clear cookie on logout

dtaniwaki:
We should clear the `oauthenticator-state` cookie just in case for security.


minrk:
Sure! This should very rarely have any effect because the oauth state cookie is already cleared on successful login, so this should only happen in the event of logout after failed login...

Still good to clean up everything just in case, though.

Issue #385, opened 2020-11-13
Implement the curl_httpclient


joshlrogers:
### Proposed change

Within the authenticate method there is a TODO to configure the curl_httpclient.  This is a proposal to implement this.

### Who would use this feature?

All users would as it appears to be more reliable.  For instance, we are no longer able to authenticate against Auth0 starting as of yesterday without having modified the auth0.py file to use the curl_httpclient.  We are assuming something has changed with how they handle HTTP requests on their end, but after having tried multiple mitigation measures the only successful one was to modify the code to implement the curl_httpclient.

### (Optional): Suggest a solution

To configure the curl_httpclient for Tornado.  I would submit a PR request myself to do this, at least against the auth0 authenticator, but I am not a Python dev and am not quite sure how the package dependency of pycurl is inserted into the flow.  I think requirements.txt but not sure of the implications there.

manics:
The approach taken in JupyterHub is to use the curl client if it's installed:
https://github.com/jupyterhub/jupyterhub/blob/e7bc6c2ba999a43bc2465b9a128262d29cbc615c/jupyterhub/app.py#L2287-L2296

Do you think you could do the same in oauthenticator?

joshlrogers:
@manics I do not remember whether I did or not, I'll change the code soon and report back whether it continues to work.

joshlrogers:
@manics It does appear that simply installing pycurl works.  I removed the altered line and restarted jupyterhub and it continued working.

meeseeksmachine:
This issue has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/failing-to-authenticate-against-auth0-with-stream-closed-exception/6723/3

PR #410, opened 2021-02-08
[OpenShift] add allowed_groups and admin_groups

wseaton:
What this PR does is add both `allowed_groups` and `admin_groups` configuration options similar to that of the other oauthenticators.

Since OpenShift now allows you to sync groups from other IdP such as LDAP it is useful to be able to leverage group based authentication, but have the hub still point to the built-in OAuth provider. 

Example: In a multi-namespace deployment it might be helpful to feature-gate a hub based on the type of hardware attached to it (ie. the GPU cluster should only be available to those in the `datascience` cluster group).

@vpavlin do you know if this should be feature-gated to a particular OCP version? I tested it personally on OCP 4.6 and don't have access to any other clusters.

### DIFF
--- a/oauthenticator/openshift.py
+++ b/oauthenticator/openshift.py
@@ -162,6 +162,7 @@ class OpenShiftOAuthenticator(OAuthenticator):
         is authenticated based on groups, an admin, or both.
         """
         user_groups = set(user_info['auth_state']['openshift_user']['groups'])
+        username = user_info['name']
 
         if self.admin_groups:
             is_admin = self.user_in_groups(user_groups, self.admin_groups)
@@ -175,7 +176,7 @@ class OpenShiftOAuthenticator(OAuthenticator):
             return user_info
         else:
             msg = f"username:{username} User not in any of the allowed/admin groups"
-            self.log.warning(msg.format(username=user_info['name']))
+            self.log.warning(msg)
             return None
 
 
--- a/oauthenticator/openshift.py
+++ b/oauthenticator/openshift.py
@@ -28,18 +28,21 @@ class OpenShiftOAuthenticator(OAuthenticator):
     scope = ['user:info']
 
     openshift_url = Unicode(
-        os.environ.get('OPENSHIFT_URL') or 'https://openshift.default.svc.cluster.local', config=True
+        os.environ.get('OPENSHIFT_URL')
+        or 'https://openshift.default.svc.cluster.local',
+        config=True,
     )
 
     validate_cert = Bool(
         True, config=True, help="Set to False to disable certificate validation"
     )
 
-    ca_certs = Unicode(
-        config=True
-    )
+    ca_certs = Unicode(config=True)
 
-    allowed_groups = List(config=True, help="List of OpenShift groups that should be allowed to access the hub.")
+    allowed_groups = List(
+        config=True,
+        help="List of OpenShift groups that should be allowed to access the hub.",
+    )
 
     @default("ca_certs")
     def _ca_certs_default(self):
@@ -61,7 +64,9 @@ class OpenShiftOAuthenticator(OAuthenticator):
         return resp_json.get('issuer')
 
     openshift_rest_api_url = Unicode(
-        os.environ.get('OPENSHIFT_REST_API_URL') or 'https://openshift.default.svc.cluster.local', config=True
+        os.environ.get('OPENSHIFT_REST_API_URL')
+        or 'https://openshift.default.svc.cluster.local',
+        config=True,
     )
 
     @default("openshift_rest_api_url")
@@ -138,13 +143,11 @@ class OpenShiftOAuthenticator(OAuthenticator):
             'auth_state': {'access_token': access_token, 'openshift_user': ocp_user},
         }
 
-
         if self.allowed_groups:
             if not any(set(ocp_user['groups']).intersection(set(self.allowed_groups))):
                 msg = "username:{username} User not in any of the allowed groups"
                 self.log.warning(msg.format(username=username))
                 return None
-
 
         return auth_data
 
--- a/oauthenticator/openshift.py
+++ b/oauthenticator/openshift.py
@@ -138,18 +138,16 @@ class OpenShiftOAuthenticator(OAuthenticator):
 
         username = ocp_user['metadata']['name']
 
-        auth_data = {
-            'name': username,
-            'auth_state': {'access_token': access_token, 'openshift_user': ocp_user},
-        }
-
         if self.allowed_groups:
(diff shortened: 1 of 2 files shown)


wseaton:
I am also working on `admin_groups` as well, pending a refactor. Will update the status when ready to review

oauthenticator/openshift.py
```diff
@@ -28,15 +28,25 @@ class OpenShiftOAuthenticator(OAuthenticator):
     scope = ['user:info']
```
vpavlin, line 28: I believe that for this to work properly, we'd need to have `user:full` by default, because line 166 will fail if not

wseaton: @vpavlin are you sure? I am able to get back the groups a user is in even with the `user:info` scope in my test cluster and can persist them to `auth_state`. I thought that `user:full` was only for reserved for R/W

vpavlin: Yeah, sorry, my bad, I guess I misunderstood the scope originally. `info` looks good

vpavlin:
I tested this in OpenShift 4.6 and it works as intended. Thanks @wseaton!

@minrk Can you please take a look?

oauthenticator/openshift.py
```diff
@@ -7,18 +7,23 @@
...
 from tornado.httputil import url_concat
-from tornado.httpclient import HTTPRequest, AsyncHTTPClient, HTTPClient
-from traitlets import Bool, Unicode, Set, default
 
-from jupyterhub.auth import LocalAuthenticator
+from oauthenticator.oauth2 import OAuthenticator
+from traitlets import Bool, Set, Unicode, default
 
-from .oauth2 import OAuthenticator
+try:
```
wseaton, line 22: @vpavlin in my testing, at least on our UBI based images using the `pycurl` backend fixes any issues we have been having with self-signed certs. Honestly this makes sense considering I couldn't reproduce the issue with standard curl. From what I can tel this isn't pulled in as a dependency automatically by tornado so we'll have to add a step to our build to compile it w/ `nss` support.

vpavlin: Yeah, that makes sense. The change looks good to me and I'll make sure the right packages are installed in our images to support pycurl

oauthenticator/openshift.py
```diff
@@ -127,13 +149,42 @@ async def authenticate(self, handler, data=None):
...
+        """
+        user_groups = set(user_info['auth_state']['openshift_user']['groups'])
+
+        if self.admin_groups:
+            is_admin = self.user_in_groups(user_groups, self.admin_groups)
+
+        user_in_allowed_group = self.user_in_groups(user_groups, self.allowed_groups)
+
+        if self.admin_groups and (is_admin or user_in_allowed_group):
+            user_info['admin'] = is_admin
```
vpavlin, line 179: @wseaton I did a bit more testing and it seams that the `admin` "flag" is not working properly with this flow:

1. group `odh-admins` is set in `admin_groups`
1. user `vpavlin` is in `odh-admins`
1. `is_admin` is set for `vpavlin`
1 `vpavlin` is removed from `odh-admins`
1. log out and log back in
1. `vpavlin` is *still* `admin` (I think the `is_admin` is set in JH DB at that point)
1. go to `Admin` section and remove `admin` privilege for `vpavlin`
2. `vpavlin` is *no longer* `admin`

I think we might need to set `user_info['admin']` always to make sure that it properly respects the groups configuration

oauthenticator/openshift.py
```diff
@@ -7,18 +7,23 @@
...
-from traitlets import Bool, Unicode, default
 
-from jupyterhub.auth import LocalAuthenticator
+from oauthenticator.oauth2 import OAuthenticator
+from traitlets import Bool, Set, Unicode, default
 
-from .oauth2 import OAuthenticator
+try:
+    import pycurl
+    AsyncHTTPClient.configure("tornado.curl_httpclient.CurlAsyncHTTPClient")
```
manics, line 24: This is already done in JupyterHub:
https://github.com/jupyterhub/jupyterhub/blob/d289cd1e025dfc35eec1d42df55cc2d274453b48/jupyterhub/app.py#L2321-L2330
and it should apply globally.

See the discussion in https://github.com/jupyterhub/oauthenticator/issues/385

I might be worth making this clearer in the docs though.

wseaton: @manics no worries, I can remove this block. Got caught up in testing for our hotfix and didn't test w/ `pycurl` installed w/o it.

wseaton: @manics I'll remove this, and yes a docs update would be great. I was misled a bit by the TODO inline comment so we may want to strip those as well.

oauthenticator/openshift.py
vpavlin: Ok, I made a mistake in my flow up there - what I actually did was not to remove a user, but remove the `admin_group` which basically made oauthenticator skip the whole `admin` section.

When I did the test correctly, it worked as expected.

vpavlin (approved):

oauthenticator/openshift.py
manics: Could you create a new issue on this repo so it doesn't get forgotten :smile: ?

oauthenticator/openshift.py
```diff
@@ -127,13 +142,42 @@ async def authenticate(self, handler, data=None):
...
+
+        user_in_allowed_group = self.user_in_groups(user_groups, self.allowed_groups)
+
+        if self.admin_groups and (is_admin or user_in_allowed_group):
+            user_info['admin'] = is_admin
+            return user_info
+        elif user_in_allowed_group:
+            return user_info
+        else:
+            msg = "username:{username} User not in any of the allowed/admin groups"
```
consideRatio, line 177: Supposed to be an fstring here right?

```suggestion
            msg = f"username:{username} User not in any of the allowed/admin groups"
```

consideRatio (commented):
I love that tests are added! I'm not able to test this myself and I've never worked with openshift, but this seems reasonable to me.

oauthenticator/openshift.py
wseaton: @consideRatio yes, good catch! fixed.

wseaton:
@consideRatio do you need anything more from me on this PR? We are currently using an internal fork of these changes but it would be nice to make it into the next release :slightly_smiling_face:

consideRatio:
@wseaton this is a change only to the openshift authenticator and you have verified it to work as intended, that is excellent!

Is this a pure opt-in enhancement or will it cause changes to existing function that users need to know about when upgrading? Perhaps you can update the first post of this PR to answer that so writing a changelog becomes easy!

Thank you for your work on this!

PR #422, opened 2021-03-31
Note that whitelist should be used if not in 1.2

mafloh:
As commented in https://github.com/jupyterhub/oauthenticator/pull/366#pullrequestreview-483095919 it should be noted that whitelist for groups and projects works in gitlab if still on version 1.2.

I spend some hours because I did not realise the doc refers to a version which is not the standard version.


consideRatio:
Thank you @mafloh!!! :heart: :tada:

Issue #313, opened 2019-11-26
When to define a custom authenticator?


consideRatio:
There will be almost infinite amounts of authenticators in this repo if we make a custom class for all of them. With this in mind, when should we make a new module in this repo with a class for a custom authenticator, and when shouldn't we? Getting a clear answer on that, and then documenting it would be nice.

manics:
Could we encourage all new authenticators to be in a separate package with `oauthenticator` as a dependency? It means ensuring breaking changes aren't made to the core classes, and perhaps improving the docs?

minrk:
I think this is a really hard decision to make. If I had my choice, we would probably only have major repo hosts (GitHub, Bitbucket, GitLab) and Google here. But kicking any out would mean creating and maintaining separate packages for the ones we have, which is *more* work than keeping them here. At this point, it's hard to draw a line anywhere other than who got here first.

I think the only feasible course is @manics idea that we stop here and say "oauthenticator is not accepting new providers, here are the docs for writing your own." That also means the work of making a stable public API for writing new OAuthenticator subclasses in external packages, which we don't have right now.

minrk:
opened #316 to start this. It's very rough for now, but gets things started with the right places to write these docs.

PR #425, opened 2021-04-13
Add FeiShu Authenticator

harrywang:
FeiShu (https://www.feishu.cn/en/) is one of the most popular team collaboration tools developed by ByteDance (the company behind Tiktok). This PR enables OAuth authentication via FeiShu. We have tested this in our internal JupyterHub deployment at Tezign.com.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ The docs also provide example setups for different OAuth2 identity providers:
 
 * [General Setup](https://oauthenticator.readthedocs.io/en/latest/getting-started.html#general-setup)
 * [Azure AD](https://oauthenticator.readthedocs.io/en/latest/getting-started.html#azure-ad-setup)
+* [FeiShu](https://oauthenticator.readthedocs.io/en/latest/getting-started.html#feishu-setup)
 * [GitHub](https://oauthenticator.readthedocs.io/en/latest/getting-started.html#github-setup)
 * [GitLab](https://oauthenticator.readthedocs.io/en/latest/getting-started.html#gitlab-setup)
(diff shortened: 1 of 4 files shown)


manics:
Hi! Thanks for opening a PR.
I'm afraid we're not accepting new authenticators into this repository due to the testing and maintenance overhead, you can see some of the discussions in https://github.com/jupyterhub/oauthenticator/issues/313.

What we'd like everyone to do is to package new authenticators in their own python package, with oauthenticator as a dependency. This has the added advantage that it'll be easier for you to fix issues since you have full control over the repository, and you'll also be able to add full integration tests requiring secrets.

Please feel free to add a link in our docs though once it's published. Hope this makes sense!

harrywang:
@manics this makes sense - Do you have an example of such a package? I searched around and could not find one - I want to follow an example to create the FeiShu package if possible. Thanks.

manics:
@harrywang Here's a couple of examples:
- https://github.com/facebookincubator/jupyterhub_fb_authenticator
- https://github.com/swan-cern/jupyterhub-extensions/tree/c250978c7f2666c0caca7861d98c0747574f218f/KeyCloakAuthenticator

If you have any more questions would you mind asking on the community forum https://discourse.jupyter.org/ ? That way the rest of the community will benefit from the discussion, and it's also a chance for you to advertise your work :smile:

harrywang:
@manics Thanks a lot. We will work on this :).

harrywang:
@manics we have tested and published the FeiShu Authenticator and issued a PR #427 

Thanks!!

PR #420, opened 2021-03-25
[azuread] pyjwt 1+2 compatibility, azuread test coverage

minrk:
Since mwoauth pins pyjwt 1, we need to keep support for pyjwt 1 and 2.

Adds coverage for AAD, which would have caught the regression in #402

closes #419

PR #431, opened 2021-05-06
ci: unpin pyjwt in test-requirements.txt

### LINKED ISSUE #430: Support for PyJWT > 2
### Bug description
PyJWT is constraint to [`>=1.7,<2.0`](https://github.com/jupyterhub/oauthenticator/blob/master/test-requirements.txt#L5)

#### Expected behaviour
Allow later PyJWT release

#### Actual behaviour
This constraint makes the package building process fails on NixOS. NixOS has PyJWT 2.1.0. I guess other distributions are affected too in the meantime as PyJWT 2.0.0 was almost 6 months ago.

Thanks

consideRatio:
I think there is support for pyjwt in OAUthenticator since of #420, so we can unpin the test-requirements.txt and close #430 which mentions the test-requirements.txt file.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -2,7 +2,7 @@
 .[googlegroups]
 codecov
 flake8
-pyjwt>=1.7,<2.0
+pyjwt
 mwoauth >= 0.3.7
 pytest >= 2.8
 pytest-cov


consideRatio:
Since this is a CI PR only modifying test-requirements.txt and the CI pipeline making use of it worked fine, I'll go ahead and self-merge this quickly.

PR #434, opened 2021-05-14
pre-commit configured and executed

consideRatio:
General maintenance to follow the common practice of having a pre-commit configuration setup for the JupyterHub org repo.


.flake8
```diff
@@ -1,6 +1,16 @@
...
+# W: style warnings
+# C: complexity
+# F401: module imported but unused
+# F403: import *
+# F811: redefinition of unused `name` from line `N`
+# F841: local variable assigned but never used
+# E402: module level import not at top of file
+# I100: Import statements are in the wrong order
+# I101: Imported names are in the wrong order. Should be
+ignore = E, C, W, F401, F403, F811, F841, E402, I100, I101, D400
```
manics, line 13: Was this change necessary to deal with the black autoformatter? It seems like an unexpectedly long list of exclusions.

In general I think we should be as strict as possible except where it conflicts with the black formatter.

manics: Actually maybe this is fine,is it to with removing `select = F`?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,52 @@
...
+#
+# Common tasks
+#
+# - Run on all files:   pre-commit run --all-files
+# - Register git hooks: pre-commit install --install-hooks
+#
+repos:
+  # Autoformat: Python code
+  - repo: https://github.com/ambv/black
+    rev: stable
```
manics, line 14: This causes an installation warning:
`[WARNING] The 'rev' field of repo 'https://github.com/ambv/black' appears to be a mutable reference (moving tag / branch).  Mutable references are never updated after first install and are not supported.  See https://pre-commit.com/#using-the-latest-version-for-a-repository for more details.  Hint: `pre-commit autoupdate` often fixes this.`

manics, line 14:
```suggestion
  - repo: https://github.com/psf/black
    rev: 21.5b1
```

manics:
A few minor comments, but as you can probably guess I'm in favour of pre-commit :smile:

.flake8
consideRatio: I don't know, i just copy pasted to mimic jupyterhub repo.

I recall Min being opinionated towards being relaxed unless it can cause problems in the flake8 checks. I'm personally inclined towards not needing to f

.pre-commit-config.yaml
consideRatio: 👍

.flake8
manics: `ignore = E, C, W, F401, F841` seems to pass flake8

Of the remainder:
```
# F403: import *
```
Best to avoid in new code as it pollutes the namespace
```
# F811: redefinition of unused `name` from line `N`
```
Potential bug in new code, you've assigned a value to a variable and then assigned another value to that variable without using the first value
```
# E402: module level import not at top of file
# I100: Import statements are in the wrong order
# I101: Imported names are in the wrong order. Should be
```
These are presumably taken care of by `id: reorder-python-imports`?

manics: https://github.com/consideRatio/oauthenticator/pull/1 cleans up the unused imports

consideRatio: Thanks! Makes sense to me!!

manics: It can be a lot of work to make an old repo compliant with flake8, I suspect this is one reason for lots of exclusions. In general once the code is compliant it's not too difficult to keep it that way, I've had several bugs that were detected by flake8 before I'd even run my tests.

Arguably for a repo like oauthenticator where we don't run a lot of the code ourselves we should be even stricter as we're often merging solely based on code review, but that's a separate discussion :smile: . Shall we merge this PR?

consideRatio:
❤️ thank you soo much @manics for your thorough care and effort, i love it!

Issue #436, opened 2021-05-21
[Auth0] Custom logout URLs


GeorgianaElena:
### Proposed change
Enable `Auth0OAuthenticator` to handle custom logout URLs.

This was noticed when using an [Auth0 database connection](https://auth0.com/docs/connections/database) to authenticate users using an email/username and password (ref: https://github.com/2i2c-org/pilot-hubs/issues/422).  

### Who would use this feature?

`Auth0OAuthenticator`-enabled hub users. 

### (Optional): Suggest a solution

Users need to be sent to the [Auth0 logout endpoint](https://auth0.com/docs/api/authentication#logout) in order to be logged-out out at 
[Auth0 Session Layer](https://auth0.com/docs/logout), otherwise only the JupyterHub session is closed. 

The `GlobusOAuthenticator` already implements  a [custom logout handler](https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/globus.py#L21) to enable custom logout URLs.

PR #437, opened 2021-05-26
Support custom logout url (logout_redirect_url)

GeorgianaElena:
Fixes https://github.com/jupyterhub/oauthenticator/issues/436

This should allow setting `OAuthenticator.logout_redirect_url` to be a logout endpoint, such as the [Auth0 logout endpoint](https://auth0.com/docs/api/authentication#logout)  to log users out of the Auth0 session and potentially out of their identity providers as well if the `federated` query string parameter is passed.

Edit by Erik: updated to be non Auth0 specific

### DIFF
--- a/oauthenticator/auth0.py
+++ b/oauthenticator/auth0.py
@@ -35,7 +35,6 @@ from jupyterhub.auth import LocalAuthenticator
 from tornado.httpclient import HTTPRequest
 from traitlets import default
 from traitlets import Unicode
-from traitlets import Bool
 
 from .oauth2 import OAuthenticator
 from .oauth2 import OAuthLogoutHandler
--- a/oauthenticator/auth0.py
+++ b/oauthenticator/auth0.py
@@ -76,7 +76,7 @@ class Auth0OAuthenticator(OAuthenticator):
 
     @default("logout_redirect_url")
     def _logout_redirect_url_default(self):
-        return os.getenv('LOGOUT_REDIRECT_URL', '')
+        return 'https://%s.auth0.com/v2/logout' % self.auth0_subdomain
 
     @default("authorize_url")
     def _authorize_url_default(self):
--- a/oauthenticator/auth0.py
+++ b/oauthenticator/auth0.py
@@ -58,12 +58,6 @@ class Auth0OAuthenticator(OAuthenticator):
 
     auth0_subdomain = Unicode(config=True)
 
-    return_to_login = Bool(
-        False,
-        help="""Whether or not to return to the hub main page after auth0 logout endpoint redirect.""",
-        config=True,
-    )
-
     logout_redirect_url = Unicode(help="""URL for logging out of Auth0""", config=True)
 
     @default("auth0_subdomain")
--- a/oauthenticator/auth0.py
+++ b/oauthenticator/auth0.py
@@ -57,8 +57,6 @@ class Auth0OAuthenticator(OAuthenticator):
 
     auth0_subdomain = Unicode(config=True)
 
-    logout_redirect_url = Unicode(help="""URL for logging out of Auth0""", config=True)
-
     @default("auth0_subdomain")
     def _auth0_subdomain_default(self):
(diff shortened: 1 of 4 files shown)


oauthenticator/auth0.py
```diff
@@ -54,6 +68,10 @@ def _auth0_subdomain_default(self):
                 % self.__class__.__name__
             )
 
+    @default("logout_redirect_url")
+    def _logout_redirect_url_default(self):
+        return os.getenv('LOGOUT_REDIRECT_URL', '')
```
minrk, line 73: Should the default be `"https://%s.auth0.com/v2/logout" % self.auth0_subdomain` like the authorize url? Is there a reason to *not* use this?

Let's also use `OAUTH_LOGOUT_REDIRECT_URL` to be consistent with other env names.

oauthenticator/auth0.py
```diff
@@ -37,14 +37,28 @@
...
 
 class Auth0OAuthenticator(OAuthenticator):
 
     login_service = "Auth0"
 
+    logout_handler = Auth0LogoutHandler
+
     auth0_subdomain = Unicode(config=True)
 
+    logout_redirect_url = Unicode(help="""URL for logging out of Auth0""", config=True)
```
minrk, line 60: Should this behavior be on the base class instead of just Auth0? Then the only thing in Auth0 is the default value for the url?

minrk (commented):
Great! Just a question about defaults and base classes, but 👍

oauthenticator/auth0.py
GeorgianaElena: > Should the default be "https://%s.auth0.com/v2/logout" % self.auth0_subdomain like the authorize url? Is there a reason to not use this?

I think it should. I just wasn't sure if we wanted the logout endpoint to be enabled by default.

oauthenticator/auth0.py
GeorgianaElena: That would make more sense! I just moved it! Thank you!

oauthenticator/auth0.py
```diff
@@ -37,12 +37,24 @@
 from traitlets import Unicode
 
 from .oauth2 import OAuthenticator
+from .oauth2 import OAuthLogoutHandler
+
+
+class Auth0LogoutHandler(OAuthLogoutHandler):
+    async def render_logout_page(self):
+        if self.authenticator.logout_redirect_url:
+            self.redirect(self.authenticator.logout_redirect_url)
```
minrk, line 46: Now that the `logout_redirect_url` config is on the base class, this redirect behavior should probably be on the base OAuthLogoutHandler class to match, and Auth0 doesn't need a custom `logout_handler` class anymore.

GeorgianaElena: Thanks, @minrk! I just moved the redirect behavior and rebased the PR

oauthenticator/oauth2.py
```diff
@@ -239,6 +239,13 @@ class OAuthLogoutHandler(LogoutHandler):
     async def handle_logout(self):
         self.clear_cookie(STATE_COOKIE_NAME)
 
+    async def render_logout_page(self):
+        if self.authenticator.logout_redirect_url:
+            self.redirect(self.authenticator.logout_redirect_url)
+            return
+
+        super().render_logout_page()
+
```
minrk, line 248: To ensure proper extension of possibly async method, we must *either*:

- use `async def` and  await the super result, or
- use `def` and directly return the super result so it can be awaited

```suggestion
   def render_logout_page(self):
        if self.authenticator.logout_redirect_url:
            self.redirect(self.authenticator.logout_redirect_url)
            return

        return super().render_logout_page()

```

minrk: One last thing, then 👍 to merge.

consideRatio:
Wieee thanks for your work on this @GeorgianaElena and reviewing @minrk!

minrk:
Great!

Issue #266, opened 2019-05-22
[Auth0] Using Auth0 usernames instead of email addresses


JuanCab:
I noticed that `https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/auth0.py` appears to hardcode use of email address as the username when using Auth0.  Is there a way to allow Auth0 USERNAME as the JupyterHub username?  It would make my life a lot easier to not have usernames locked to email addresses.

minrk:
Definitely possible, just need to add a switch. It *seems* like this should have been the default, but it's hard to change defaults without causing pain for existing users. Check out [username_key](https://github.com/jupyterhub/oauthenticator/blob/04d664e559275c0c705f1882ce1307cc47227594/oauthenticator/generic.py#L56) in generic.py for a configurable source of username from the OAuth response.

JuanCab:
I am not an OAuth expert, just a professor trying to piece together a solution for a class server.  As far as I can interpret, username_key is a traitlet (which I partly understand the concept of) which can contain a string, the default being 'username'.  Is this supposed to allow me to change what part of the response JSON the generic `oauthenticator` interprets as the username? 

I believe (if I understand the documentation well enough) the Auth0 OAuth authenticator responds with `preferred_username` as the username, so do I just have to set the `username_key` traitlet value in the Jupyter config and then I will get the proper username instead of email address as the username?

JuanCab:
I should note I am dealing with this in the context go TLJH and not a full JupyterHub installation.

JuanCab:
OK, I attempted to set up the generic Oauth authenticator, using the following settings (these are from TLJH config):
```
auth:
  type: oauthenticator.generic.GenericOAuthenticator
  GenericOAuthenticator:
    client_id: [REMOVED]
    client_secret: [REMOVED]
    oauth_callback_url: https://jupyter.mnstate.edu/hub/oauth_callback
    token_url: https://msumjupyter.auth0.com/oauth/token
```
and this just seems to generate a `ERR_TOO_MANY_REDIRECTS` in Chrome. :(  

I also, as a brute force approach, tried to edit my auth0.py file and replaced `'name': resp_json["email”]` with `'name': resp_json["username”]` but that triggered a 500 Server error, the TLJH logs indicating that this is because 
```
File "/opt/tljh/hub/lib/python3.6/site-packages/oauthenticator/auth0.py", line 99, in authenticate
 'name': resp_json["username"],
 KeyError: 'username'
[...]

PR #439, opened 2021-05-27
Allow changing auth0 username key

GeorgianaElena:
Fixes https://github.com/jupyterhub/oauthenticator/issues/266

This should allow choosing what key in the authenticated user's profile to use as hub username.


minrk:
Great!

PR #440, opened 2021-05-29
[Globus] Add config username_from_email

rpwagner:
By default, the Globus OAuthenticator takes the JupyterHub `username`  from the OIDC `preferred_username` claim. In many cases, this is the same as the emailaddress. However, some identity providers use an opaque string, e.g.,`046f34a240f0615e01420b3ff4350922@ucsd.edu`. This added the configuration `username_from_email = True` to get it from the user's email address. Setting this will automatically adds `email` to the list of scopes. When `identity_provider` is set, the email address domain must still match the identity provider domain.

Documentation and tests are included in this pull request.

### DIFF
--- a/docs/source/getting-started.rst
+++ b/docs/source/getting-started.rst
@@ -399,6 +399,18 @@ to login, and will continue to be tied to that identity after changing this
 setting. Create a new Globus App with your preferred 'Required Identity Provider'
 to avoid this problem.
 
+Username from Email Address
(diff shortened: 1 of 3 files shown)


rpwagner:
@NickolausDS a change to get the username from a user's email address, rather than the preferred username. Helpful for our campus that uses an opaque string.

consideRatio (approved):
Oh this PR was such a breeze to review! I thought the code looked good and with the addition of docs and tests this looks amazing! Thanks!

PR #441, opened 2021-05-29
[Globus] Add config to manage: allowed, admin, and blocked users through Globus groups

rpwagner:
This adds functionality to manage allowed, admin, and blocked users through [Globus
Groups](https://docs.globus.org/how-to/managing-groups/) similar to what's in the Google and GitLab OAuthenticators. Updated documentation included describes the expected behavior and test are include.

@NickolausDS please take a look.


NickolausDS:
The member and admin group features make sense and would be great additions. 

However, I think `blocked_globus_groups` could be removed. This should be covered by membership in the `allowed_globus_groups` groups set -- if a user doesn't have an active membership in any of those groups they will be blocked by default. Additionally, it may be a little strange to manage members of a group who are explicitly denied access. Globus Groups work on an invitation model, and so some admins may get confused and think they need to 'invite' banned users to the group in order to ban them.

rpwagner:
I see your point about the invite model not making sense for blocked users. So even with the new Groups API, [adding members](https://docs.globus.org/api/groups/#edit_members_on_group) still requires the user to accept an invitation?

NickolausDS:
While that is possible, the inclusive nature of groups makes them a bad way to track blocked users.

rpwagner:
Ah, I see your point. In particular, members can remove themselves from Globus Groups. So there's no way in Globus Groups to maintain a list of members that is appropriate, like being private and unchangeable by users. I'll remove the `blocked_globus_groups` feature and update the documentation with a note on how to manager blocked users either through the existing config, or by setting an allowed group to invitation only, so users can't add themselves. Does that make sense?

NickolausDS:
The new changes look much better! 

For banning users, admins should still be able to use `allowed_globus_groups` and simply suspend a user account they no longer want to allow access. I didn't check the code to verify, but as long as its checking for active membership this should work just fine. 

One edge case to this model is that it may be possible for a banned user to regain access through a secondary group. For instance, if an admin set two groups Group A and Group B, and banned a user from group A, said user may be able to join group B to regain access while still being suspended from group A. It might be nice to log a warning just so the Admin is aware if this happens.

NickolausDS:
I don't have time at the moment, but tomorrow I'll try to run the code to test.

oauthenticator/globus.py
```diff
@@ -199,13 +225,51 @@ async def authenticate(self, handler, data=None):
...
             },
         }
+        use_globus_groups = False
+        user_allowed = False
+        if self.allowed_globus_groups or self.admin_globus_groups:
+            # If any of these configurations are set, user must be in the allowed or admin Globus Group
+            use_globus_groups = True
+            user_group_ids = set()
+            # Groups user is an admin or manager of
+            user_admin_groups = set()
```
NickolausDS, line 243: This variable `user_admin_groups` isn't used and can be removed.

NickolausDS:
Ok, I tested with a few different criteria, and everything looks good to me. I left one small code comment on an unused variable, but otherwise the code is functional. The things I tested include:

* User not in any allowed Globus groups is unable to login with 403
* User in an allowed_globus_group is able to login
* User who is suspended in an allowed_globus_groups is unable to login with 403
* User who is suspended in an allowed_globus_group but also member of another allowed_globus_group is able to login
* User who is suspended in multiple globus groups isn't able to login with 403
* User in an admin group is able to login with admin privileges. 

Overall, I think the 'suspension' feature of Globus groups works really well for rejecting specific users, and I think it contains the desired behavior of the old `blocked_globus_groups` feature. It's also really easy to read on the Globus Groups page who is and isn't allowed access. 

Changes look good to me, thanks @rpwagner!

rpwagner:
Thanks for checking it out, @NickolausDS! I've removed the unused variable. Glad it looks good. I think it will also be really useful for group using Globus Auth.

docs/source/getting-started.rst
```diff
@@ -455,6 +455,39 @@ disable transfers, modify ``c.GlobusOAuthenticator.scope`` instead of
...
+managed by others. For example, researchers could manage groups of
+collaborators. Each of these settings can contain multiple Globus
+Groups.
+
+.. code:: python
+
+   # Groups of allowed users
+   c.GlobusOAuthenticator.allowed_globus_groups = set
+   authenticator.allowed_globus_groups = set({'d11abe71-5132-4c04-a4ad-50926885dc8c',
+                                                                           '21c6bc5d-fc12-4f60-b999-76766cd596c2'})
```
consideRatio, line 475:
```suggestion
   authenticator.allowed_globus_groups =  {
       'd11abe71-5132-4c04-a4ad-50926885dc8c',
       '21c6bc5d-fc12-4f60-b999-76766cd596c2',
   }
```

docs/source/getting-started.rst
```diff
@@ -455,6 +455,39 @@ disable transfers, modify ``c.GlobusOAuthenticator.scope`` instead of
...
+   # Admin users
+   authenticator.admin_globus_groups = set({'3f1f85c4-f084-4173-9efb-7c7e0b44291a'})
```
consideRatio, line 477:
```suggestion
   authenticator.admin_globus_groups = {'3f1f85c4-f084-4173-9efb-7c7e0b44291a'}
```

docs/source/getting-started.rst
```diff
@@ -455,6 +455,39 @@ disable transfers, modify ``c.GlobusOAuthenticator.scope`` instead of
 ``c.GlobusOAuthenticator.exclude`` to avoid procuring unnecessary
 tokens.
 
+Group Management
+~~~~~~~~~~~~~~~
```
consideRatio, line 459:
```suggestion
~~~~~~~~~~~~~~~~
```

consideRatio (approved):
Another excellent PR to review! Docs + tests and already pre-reviewed by @NickolausDS that made excellent notes as well!

:+1: :tada: :sunflower: :heart:

PR #442, opened 2021-06-05
[GitHub] Set JH user's email with non-public email if needed and granted scope to do so

### LINKED ISSUE #438: GitHub OAuth process not retrieving email address.
### Bug description
we have modified the github authenticator to pass on login, token, and email to the user environment. however, while login and token are being passed, the email address is not. the reading scope is set to read user info including email. we are hoping to set this so we can add that for git operations in the lab.

https://github.com/dandi/dandi-hub/blob/dandi/dandi-info/config.yaml.j2#L49

#### Expected behaviour
email address should be set

#### Actual behaviour
email address is empty

### How to reproduce
other than signing up for our service, i don't have a simple demo that reduces the issue to a localhost setup with github oauth, but in principle one could use the authenticator code above and run a dev environment.

### Your personal set up
a jupyterhub deployed on aws.

if it's helpful this is the logs from the auth flow. i'm able to login, just that the variable representing email address is empty.

```
/opt/conda/lib/python3.8/site-packages/jupyter_server/base/handlers.py:131: RuntimeWarning: Expected to see HubAuthenticatedHandler in <class 'jupyterlab_server.handlers.LabHandler'>.mro()
  return self.login_handler.get_user(self)
[I 2021-05-14 15:31:04.645 SingleUserNotebookApp log:189] 302 GET /user/satra/lab? -> /hub/api/oauth2/authorize?client_id=jupyterhub-user-satra&redirect_uri=%2Fuser%2Fsatra%2Foauth_callback&response_type=code&state=[secret] (@::ffff:172.20.58.53) 2.85ms
[I 2021-05-14 15:31:04.929 SingleUserNotebookApp auth:992] Logged-in user {'kind': 'user', 'name': 'satra', 'admin': True, 'groups': [], 'server': '/user/satra/', 'pending': None, 'created': '2021-05-14T01:52:17.596441Z', 'last_activity': '2021-05-14T15:31:04.921258Z', 'servers': None}
[I 2021-05-14 15:31:04.930 SingleUserNotebookApp log:189] 302 GET /user/satra/oauth_callback?code=[secret]&state=[secret] -> /user/satra/lab? (@::ffff:172.20.58.53) 53.09ms
[I 2021-05-14 15:31:05.049 SingleUserNotebookApp log:189] 200 GET /user/satra/lab? (satra@::ffff:172.20.58.53) 12.73ms
/opt/conda/lib/python3.8/site-packages/jupyter_server/base/handlers.py:131: RuntimeWarning: Expected to see HubAuthenticatedHandler in <class 'jupyter_server.base.handlers.FileFindHandler'>.mro()
  return self.login_handler.get_user(self)
[I 2021-05-14 15:31:08.485 SingleUserNotebookApp log:189] 200 GET /user/satra/api/kernelspecs?1621006268342 (satra@::ffff:172.20.58.53) 38.21ms
/opt/conda/lib/python3.8/site-packages/jupyter_server/base/handlers.py:131: RuntimeWarning: Expected to see HubAuthenticatedHandler in <class 'jupyterlab_server.settings_handler.SettingsHandler'>.mro()
  return self.login_handler.get_user(self)
/opt/conda/lib/python3.8/site-packages/jupyter_server/base/handlers.py:131: RuntimeWarning: Expected to see HubAuthenticatedHandler in <class 'jupyterlab_server.workspaces_handler.WorkspacesHandler'>.mro()
  return self.login_handler.get_user(self)
```

satra:
This allows emails to be read and set in the state dictionary when scope is authorized.

closes #438

### Related references
- https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps#available-scopes
- https://docs.github.com/en/rest/guides/basics-of-authentication#checking-granted-scopes
- https://docs.github.com/en/rest/reference/users#list-email-addresses-for-the-authenticated-user
- https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#response
  The `scope` field in the JSON response when acquiring the access token is not always present it seems.

### DIFF
--- a/oauthenticator/github.py
+++ b/oauthenticator/github.py
@@ -7,6 +7,7 @@ import warnings
 
 from jupyterhub.auth import LocalAuthenticator
 from tornado import web
+from tornado.httpclient import HTTPClientError
 from tornado.httpclient import HTTPRequest
 from tornado.httputil import url_concat
 from traitlets import default
@@ -193,17 +194,20 @@ class GitHubOAuthenticator(OAuthenticator):
         auth_state['github_user'] = resp_json
         # To retrieve emails an extra call has to be made to a secondary endpoint
         # using the access token. see https://github.com/jupyterhub/oauthenticator/issues/438
-        if 'user:email' in self.scope:
-            req = HTTPRequest(
-                      self.github_api + "/user/emails",
-                      method="GET",
-                      headers=_api_headers(access_token),
-                      validate_cert=self.validate_server_cert,
-                 )
+        req = HTTPRequest(
+            self.github_api + "/user/emails",
+            method="GET",
+            headers=_api_headers(access_token),
+            validate_cert=self.validate_server_cert,
+        )
+        try:
             resp_json = await self.fetch(req, "fetching user emails")
+        except HTTPClientError:
+            pass
+        else:
             for val in resp_json:
-              if val["primary"]:
-                  auth_state['github_user']['email'] = val['email']
+                if val["primary"]:
+                    auth_state['github_user']['email'] = val['email']
 
         return userdict
 
--- a/oauthenticator/github.py
+++ b/oauthenticator/github.py
@@ -195,8 +195,9 @@ class GitHubOAuthenticator(OAuthenticator):
         # store the whole user model in auth_state.github_user
         auth_state['github_user'] = resp_json
         # If a public email is not available, an extra API call has to be made
-        # to a secondary endpoint using the access token.
-        # see https://github.com/jupyterhub/oauthenticator/issues/438
+        # to a /user/emails using the access token to retrieve emails. The
+        # scopes relevant for this are checked based on this documentation:
+        # https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps#normalized-scopes
         if not auth_state['github_user']['email'] and (
             'user' in scopes or 'user:email' in scopes
         ):
--- a/oauthenticator/github.py
+++ b/oauthenticator/github.py
@@ -191,8 +191,19 @@ class GitHubOAuthenticator(OAuthenticator):
         auth_state['access_token'] = access_token
         # store the whole user model in auth_state.github_user
         auth_state['github_user'] = resp_json
-        # A public email will return in the initial query (assuming default scope).
-        # Private will not.
+        # To retrieve emails an extra call has to be made to a secondary endpoint
+        # using the access token. see https://github.com/jupyterhub/oauthenticator/issues/438
+        if 'user:email' in self.scope:
+            req = HTTPRequest(
+                      self.github_api + "/user/emails",
+                      method="GET",
+                      headers=_api_headers(access_token),
+                      validate_cert=self.validate_server_cert,
+                 )
+            resp_json = await self.fetch(req, "fetching user emails")
+            for val in resp_json:
+              if val["primary"]:
+                  auth_state['github_user']['email'] = val['email']
 
         return userdict
 
--- a/oauthenticator/github.py
+++ b/oauthenticator/github.py
@@ -155,6 +155,8 @@ class GitHubOAuthenticator(OAuthenticator):
         else:
             raise web.HTTPError(500, "Bad response: {}".format(resp_json))
 
+        scopes = resp_json['scope'].split(',')
+
         # Determine who the logged-in user is
         req = HTTPRequest(
             self.github_api + "/user",
@@ -195,7 +197,9 @@ class GitHubOAuthenticator(OAuthenticator):
         # If a public email is not available, an extra API call has to be made
         # to a secondary endpoint using the access token.
         # see https://github.com/jupyterhub/oauthenticator/issues/438
-        if not auth_state['github_user']['email']:
+        if not auth_state['github_user']['email'] and (
+            'user' in scopes or 'user:email' in scopes
+        ):
             req = HTTPRequest(
                 self.github_api + "/user/emails",
                 method="GET",
--- a/oauthenticator/github.py
+++ b/oauthenticator/github.py
@@ -196,7 +196,10 @@ class GitHubOAuthenticator(OAuthenticator):
         # If a public email is not available, an extra API call has to be made
         # to a /user/emails using the access token to retrieve emails. The
         # scopes relevant for this are checked based on this documentation:
-        # https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps#normalized-scopes
+        # - about scopes: https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps#available-scopes
+        # - about /user/emails: https://docs.github.com/en/rest/reference/users#list-email-addresses-for-the-authenticated-user
+        #
+        # Note that the read:user scope does not imply the user:emails scope!
         if not auth_state['github_user']['email'] and (
             'user' in granted_scopes or 'user:email' in granted_scopes
         ):
@@ -210,6 +213,7 @@ class GitHubOAuthenticator(OAuthenticator):
             for val in resp_json:
                 if val["primary"]:
                     auth_state['github_user']['email'] = val['email']
+                    break
 
         return userdict
 
--- a/oauthenticator/github.py
+++ b/oauthenticator/github.py
@@ -192,22 +192,24 @@ class GitHubOAuthenticator(OAuthenticator):
         auth_state['access_token'] = access_token
         # store the whole user model in auth_state.github_user
         auth_state['github_user'] = resp_json
-        # To retrieve emails an extra call has to be made to a secondary endpoint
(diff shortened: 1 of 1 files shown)


oauthenticator/github.py
```diff
@@ -191,8 +191,19 @@ async def authenticate(self, handler, data=None):
         auth_state['access_token'] = access_token
         # store the whole user model in auth_state.github_user
         auth_state['github_user'] = resp_json
-        # A public email will return in the initial query (assuming default scope).
-        # Private will not.
+        # To retrieve emails an extra call has to be made to a secondary endpoint
+        # using the access token. see https://github.com/jupyterhub/oauthenticator/issues/438
+        if 'user:email' in self.scope:
```
satra, line 196: i'm not sure if this is the only scope to check or support for emails.

consideRatio: `email` is access to the public email address
`user:email` is access to the private email addresses (note plural)
`user` or `read:user` would imply `user:email` and perhaps some more details, so you can check for `read:user` and `user` alongside `user:email` to decide if this check should be made.

Details here: https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps#available-scopes

consideRatio (changes requested):
Thank you @satra for working on this! :heart: :tada:

### Suggested changes
- [ ] Include `user` and `read:user` alongisde `user:email` in the check to decide if collecting additional information from the user endpoint is relevant.
- [ ] I'd like the comment on line 194-195 to reflect the difference between the single public email, and the plural private emails
- [ ] I'd like to know this is at least manually tested against a user with multiple private emails registered, and that we understand the return value in those circumstances. Based on the code it seems like we get a primary email, then I think the action point I ask for is to clarify that while `user:email` provides info about all emails, we only use the primary email.
- [ ] Update PR description to include mentioning of https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps#available-scopes as a reference about this

### Open questions

`self.scope` is checked, but I think this is what the user has requested, not what was granted. Is there a point where we update `self.scope` (the server's spawner object's scope field) or where we keep track of the granted scopes vs the requested scopes?

There is no point for us to make a request if the granted scope lacks permissions, so can we acquire the granted scopes from the response when we get the token to call the /user endpoint perhaps?

https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps#requested-scopes-and-granted-scopes

satra:
@consideRatio - changing the logic to first check if public email is set, and relying on HTTP error rather than scope to set the email. i could not figure out a way to determine what scopes were granted, and it seemed simpler to rely on the endpoint.

satra:
nevermind, found how to get scopes. so the current state should cover your comments.

oauthenticator/github.py
```diff
@@ -195,7 +197,9 @@ async def authenticate(self, handler, data=None):
         # If a public email is not available, an extra API call has to be made
         # to a secondary endpoint using the access token.
         # see https://github.com/jupyterhub/oauthenticator/issues/438
-        if not auth_state['github_user']['email']:
+        if not auth_state['github_user']['email'] and (
+            'user' in scopes or 'user:email' in scopes
+        ):
```
consideRatio, line 203: `"user" in ["read:user"]` evaluates to False, so this should also check for `read:user`

oauthenticator/github.py
```diff
@@ -155,6 +155,8 @@ async def authenticate(self, handler, data=None):
         else:
             raise web.HTTPError(500, "Bad response: {}".format(resp_json))
 
+        scopes = resp_json['scope'].split(',')
```
consideRatio, line 158: This is granted scopes right? Let's be explicit by that and name it so

oauthenticator/github.py
satra: the normalization function would return `user` according to their example i think.

oauthenticator/github.py
```diff
@@ -191,8 +194,27 @@ async def authenticate(self, handler, data=None):
...
+            req = HTTPRequest(
+                self.github_api + "/user/emails",
+                method="GET",
+                headers=_api_headers(access_token),
+                validate_cert=self.validate_server_cert,
+            )
+            try:
+                resp_json = await self.fetch(req, "fetching user emails")
+            except HTTPClientError:
+                pass
```
consideRatio, line 213: Perhaps we should let this error properly, as we actually expect this to work reliably - right?

oauthenticator/github.py
consideRatio: `user` is read and write, so if you have requested `user` and `read:user` then `read:user` is normalized away. But, if you have only requested `read:user`, then you won't have `user`.

satra: correct and then the only way you can read email is if you requested and received `user:email`. this branch is only active if the public email based on `read:user` (the normal request earlier) is not defined.

oauthenticator/github.py
satra: the tests seem to not like `scope` as an attribute of the returned dictionary.  while this https://docs.github.com/en/rest/guides/basics-of-authentication#checking-granted-scopes would suggest otherwise.

oauthenticator/github.py
consideRatio: Hmmmm? Are you saying only user:email is relevant and neither user or read:user makr sense?

satra: `read:user` is required as a basic scope to read the user profile and that is done earlier. however, this may not set email because it is not public.

to retrieve the emails whoever is running the authenticator would need to add `user:email` or `user` the broader one to scope. thus this piece of the code uses only those two to check. i'm saying `read:user` is irrelevant for this part.

consideRatio: I understood `user:read` would not imply that you would be allowed to _set_ user email, but I expected it to allow you to read private user email. But, okay that isn't the case and then `user:email` is what is relevant.

This is the correct logic then

oauthenticator/github.py
```diff
@@ -191,8 +193,23 @@ async def authenticate(self, handler, data=None):
...
+            req = HTTPRequest(
+                self.github_api + "/user/emails",
+                method="GET",
+                headers=_api_headers(access_token),
+                validate_cert=self.validate_server_cert,
+            )
+            resp_json = await self.fetch(req, "fetching user emails")
+            for val in resp_json:
+                if val["primary"]:
+                    auth_state['github_user']['email'] = val['email']
```
consideRatio, line 212: This logic seems weird to me. Is resp_json an list, and the list elements always contain a key called primary that sometimes is a falsy value, and we set the email based on the last list element having a non-falsy "primary" key value?

consideRatio: Okay it makes sense, we can add a `break` statement as well after finding the primary email for readability

```json
[
  {
    "email": "email1@example.com",
    "primary": true,
    "verified": true,
    "visibility": "public"
  },
  {
    "email": "email2@example.com",
    "primary": false,
    "verified": true,
    "visibility": null
  }
]
```

oauthenticator/github.py
```diff
@@ -191,8 +193,23 @@ async def authenticate(self, handler, data=None):
         auth_state['access_token'] = access_token
         # store the whole user model in auth_state.github_user
         auth_state['github_user'] = resp_json
-        # A public email will return in the initial query (assuming default scope).
-        # Private will not.
+        # If a public email is not available, an extra API call has to be made
+        # to a /user/emails using the access token to retrieve emails. The
+        # scopes relevant for this are checked based on this documentation:
+        # https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps#normalized-scopes
```
consideRatio, line 199:
```suggestion
        # https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps#available-scopes
        # Note that the read:user scope does not imply the user:emails scope!
```

oauthenticator/github.py
consideRatio:
```suggestion
            for val in resp_json:
                if val["primary"]:
                    auth_state['github_user']['email'] = val['email']
                    break
```

oauthenticator/github.py
```diff
@@ -154,6 +154,8 @@ async def authenticate(self, handler, data=None):
         else:
             raise web.HTTPError(500, "Bad response: {}".format(resp_json))
 
+        granted_scopes = resp_json['scope'].split(',')
```
consideRatio, line 157: This is probably fine no matter what. I'm not confident we should count on this field being part of the response.

```suggestion
        granted_scopes = []
        if resp_json.get("scope"):
            granted_scopes = resp_json["scope"].split(",")
```

consideRatio: Test failure relates to mocked responses I think.

satra: yes, i could not find any github docs on the returned response. only pointers that suggest "scope" is in there.

consideRatio (approved):
This looks great! Thanks for your thorough work @satra! This will be included in 14.1.0 release coming this week I hope.

PR #443, opened 2021-06-23
[GitLab] Fix missing use validate_server_cert config for some web requests

wOvAN:
fixes missing validate_cert param in HTTPRequest  function calls

### DIFF
--- a/oauthenticator/gitlab.py
+++ b/oauthenticator/gitlab.py
@@ -218,7 +218,12 @@ class GitLabOAuthenticator(OAuthenticator):
                 self.member_api_variant,
                 user_id,
             )
-            req = HTTPRequest(url, method="GET", headers=headers, validate_cert=self.validate_server_cert)
+            req = HTTPRequest(
+                url,
+                method="GET",
+                headers=headers,
+                validate_cert=self.validate_server_cert,
+            )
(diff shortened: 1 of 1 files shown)


oauthenticator/gitlab.py
```diff
@@ -218,7 +218,7 @@ async def _check_membership_allowed_groups(self, user_id, access_token):
                 self.member_api_variant,
                 user_id,
             )
-            req = HTTPRequest(url, method="GET", headers=headers)
+            req = HTTPRequest(url, method="GET", headers=headers, validate_cert=self.validate_server_cert)
```
consideRatio, line 221:
```suggestion
            req = HTTPRequest(
                url,
                method="GET",
                headers=headers,
                validate_cert=self.validate_server_cert,
            )
```

oauthenticator/gitlab.py
```diff
@@ -234,7 +234,7 @@ async def _check_membership_allowed_project_ids(self, user_id, access_token):
...
```
consideRatio, line 237:
```suggestion
            req = HTTPRequest(
                url,
                method="GET",
                headers=headers,
                validate_cert=self.validate_server_cert,
            )
```

consideRatio (approved):
Thanks TLGM, there are just two formatting fixes needed to comply with our autoformatter.

Issue #445, opened 2021-07-15
AzureAD: app roles for something similar to admin_groups and allowed_groups


weisdd:
### Proposed change

Currently, AzureAD has a very basic implementation, which doesn't allow to work with users based on their group membership for admin access or for prohibiting from entering JupyterHub.

IMO, app roles would be the simplest implementation. They can be added to a group (a user must be a direct member of the group), so we would see those roles in an ID token and can later decide whether a user can be considered an admin or prohibited from entering JupyterHub at all. We could see something similar (only for role mapping) in [Grafana's AzureAD integration](https://grafana.com/docs/grafana/latest/auth/azuread/).

### Alternative options

Based on what I learned about AzureAD in a couple of days (hadn't really worked with it before), linking anything to groups will require us to give additional privileges to an app and to send subsequent requests to another endpoint, which seems to be too complex.

### Who would use this feature?

Anyone who relies on Azure AD.

### (Optional): Suggest a solution

I'll open a PR with a draft implementation.

Bfoster-melrok:
any updates on this? I'd really like to be able to use this feature.

weisdd:
@Bfoster-melrok PR still stays without any comments from JupyterHub maintainers, so no further progress.

malaman:
Hi from 2022 :) 
any plans to implement the requested feature ?

christian-sapconet:
Hi from 2023 :)

I would also like to know if there are any plans to implement this, or if there is a workaround to get the same functuionality?

Teodoriku:
Hi from 2024, that is of interest still

PR #446, opened 2021-07-15
[WIP] App roles for AzureAD

weisdd:
Signed-off-by: Igor Beliakov <demtis.register@gmail.com>

This PR is a suggested solution for https://github.com/jupyterhub/oauthenticator/issues/445 (opened by me).
Since it's the first time I try to do something with Oauth2 in Python, I'd be grateful if you bear with me for all potential defects my implementation might have.

So, I suggest to rely on App roles to give admin privileges to a user and to decide whether he/she is even allowed to enter JupyterHub.

It requires the following configuration to be done on the AzureAD side:

Azure Active Directory -> App registrations -> [Name] -> Manifest:
```json
	"appRoles": [
		{
			"allowedMemberTypes": [
				"User"
			],
			"description": "JupyterHub regular users",
			"displayName": "user",
			"id": "[RANDOM-UUID]",
			"isEnabled": true,
			"lang": null,
			"origin": "Application",
			"value": "user"
		},
		{
			"allowedMemberTypes": [
				"User"
			],
			"description": "JupyterHub admin users",
			"displayName": "admin",
			"id": "[RANDOM-UUID]",
			"isEnabled": true,
			"lang": null,
			"origin": "Application",
			"value": "admin"
		}
	],
```

**Note:** UUIDs can be generated with the help of `uuidgen`.

The second step would be:
Enterprise applications -> [NAME] -> Users and Groups -> [assign some of the above-mentioned roles to user groups]

Then, you'll need to deploy jupyterhub (`zero-to-jupyterhub`) with the following values (requires to install the modified version of `azuread.py`):

```yaml
hub:
  config:
    JupyterHub:
      admin_access: true
      authenticator_class: azuread
    Authenticator:
      admin_azure_app_roles:
        - admin
      allowed_azure_app_roles:
        - admin
        - user
```

After that, any user with the `admin` role assigned would get admin permissions and only those who have either `user` or `admin` role would be allowed to enter JupyterHub.

Additional notes:
* I'm not aware of how the whole test setup is supposed to be built, so I just replaced this code in a live container and manually started the jupyterhub process;
* I'm pretty sure I don't have exactly the same settings set for Black, so your tests are likely to fail (it'd be great if you can point me at some docs on how to make it all conformant to what you expect to see).
* It's just a draft implementation (that works in my setup), so you're more than welcomed to point at any changes in naming or main logic that need to be made.

### DIFF
--- a/oauthenticator/azuread.py
+++ b/oauthenticator/azuread.py
@@ -9,6 +9,7 @@ import jwt
 from jupyterhub.auth import LocalAuthenticator
 from tornado.httpclient import HTTPRequest
 from traitlets import default
+from traitlets import List
 from traitlets import Unicode
 
 from .oauth2 import OAuthenticator
@@ -17,6 +18,10 @@ from .oauth2 import OAuthenticator
 # pyjwt 2.0 has changed its signature,
 # but mwoauth pins to pyjwt 1.x
 PYJWT_2 = V(jwt.__version__) >= V("2.0")
+
+
+def check_user_has_role(member_roles, allowed_roles):
+    return any(role in member_roles for role in allowed_roles)
 
 
 class AzureAdOAuthenticator(OAuthenticator):
@@ -37,6 +42,18 @@ class AzureAdOAuthenticator(OAuthenticator):
     @default('username_claim')
     def _username_claim_default(self):
         return 'name'
+
+    admin_azure_app_roles = List(
+        Unicode(),
+        config=True,
+        help="App roles that should give Jupyterhub admin privileges to a user",
+    )
+
+    allowed_azure_app_roles = List(
+        Unicode(),
+        config=True,
+        help="Automatically allow users with selected app roles",
+    )
 
     @default("authorize_url")
     def _authorize_url_default(self):
@@ -80,9 +97,7 @@ class AzureAdOAuthenticator(OAuthenticator):
 
         if PYJWT_2:
             decoded = jwt.decode(
-                id_token,
-                options={"verify_signature": False},
-                audience=self.client_id,
+                id_token, options={"verify_signature": False}, audience=self.client_id,
             )
         else:
             # pyjwt 1.x
@@ -93,6 +108,16 @@ class AzureAdOAuthenticator(OAuthenticator):
         auth_state['access_token'] = access_token
         # results in a decoded JWT for the user data
         auth_state['user'] = decoded
+
+        roles = decoded.get("roles", [])
+
+        if self.admin_azure_app_roles:
+            if check_user_has_role(roles, self.admin_azure_app_roles):
+                userdict["admin"] = True
+
+        if self.allowed_azure_app_roles:
+            if not check_user_has_role(roles, self.allowed_azure_app_roles):
+                return None
 
         return userdict
 


weisdd:
@consideRatio Could you, please, let me know who can review the PR? :) Thanks!

oauthenticator/azuread.py
```diff
@@ -94,6 +111,16 @@ async def authenticate(self, handler, data=None):
         # results in a decoded JWT for the user data
         auth_state['user'] = decoded
 
+        roles = decoded.get("roles", [])
```
consideRatio, line 114: Just for readability.

```suggestion
        roles = auth_state['user'].get("roles", [])
```

consideRatio:
I updated the z2jh Helm chart config example with some notes

```yaml
hub:
  config:
    JupyterHub:
      admin_access: true
      authenticator_class: azuread
    # NOTE: This must be a class that has these traitlets defined, which isn't the base Authenticator class
    AzureADAuthenticator:
      # NOTE: I think we should name this similar to `admin_users`, suggestion below
      app_roles_with_admin_users:
        - admin
      # NOTE: I think we should name this similar to `allowed_users`, suggestion below
      # NOTE: I think we should make the admin app roles be assumed to be valid users as well
      app_roles_with_allowed_users:
        - user
```

oauthenticator/azuread.py
```diff
@@ -38,6 +43,18 @@ def _tenant_id_default(self):
...
+        config=True,
+        help="App roles that should give Jupyterhub admin privileges to a user",
+    )
+
+    allowed_azure_app_roles = List(
+        Unicode(),
+        config=True,
+        help="Automatically allow users with selected app roles",
+    )
+
```
consideRatio, line 57: I think we should make these to be `Set` instead of `List`. Practically, if anyone passes a List, for example via a YAML configuration that is read and passed to the configuration - then it will be automatically casted to a Set so thats fine.

consideRatio: I'd also like to have it clarified how to list the app roles. Is it their display name rather than id? That seems a bit odd to me but that is what matches your example configuration, but perhaps it is required to be unique as well or assumed to not risk causing problems.

consideRatio: Perhaps a link or something as a reference to learn more about app roles this is relevant as well? I'm not sure, but in general I'm thinking the help text seem a bit too short and could be extended with another sentence or two.

consideRatio:
@weisdd thanks for the ping, sorry for the stale feedback you have received. I think this PR makes sense and isn't breaking in any way and is also an opt-in mechanism that is safe to use without breaking something else.

I consider this PR to be quite merge-ready with my comments are addressed. Oh btw, I'd also like a consideration if a test could be added to verify some logic of this or if its too complicated or similar.

Btw @weisdd and @thomafred, you have both created PRs (this, and #448) to AzureAD - perhaps you can review each others PR as well to help with the maintenance burden?

thomafred:
This is quite neat, and I like how simple it is. It would certainly help our usecase.

A few questions:
- Will this work with B2C?
- Does the tenant require an active subscription?

weisdd:
> What is an app role? Is it a list of users? Is it a property a user can have? Not sure on my naming suggestions because I don't know anything about app roles.

@consideRatio Thanks for the review :)
App roles are described [here](https://docs.microsoft.com/en-us/azure/active-directory/develop/howto-add-app-roles-in-azure-ad-apps). It's very similar to [client roles in keycloak](https://www.keycloak.org/docs/latest/server_admin/#client-roles). And, btw, Grafana relies on them as well ([docs](https://grafana.com/docs/grafana/latest/auth/azuread/)).

When a user authenticates via an identity provider (IDP), the IDP may add a claim (=attribute) to identity token and/or access token sent to the client. In Azure AD, identity token is used for that.
The claim contains an array with role names (not IDs). Then, it's up to the application to decide whether it wants to do something with those roles. An app is free to ignore the claim if it doesn't care.
Every client (=application) can have its own isolated set of roles. So, even if you're reusing the same names across multiple apps, it's absolutely fine as long as those apps have their own client (and, thus, credentials).
Roles can be assigned either directly to a user or to a group the user belongs to.
The beauty of this approach is the fact that you don't have to give any permissions to the client that interacts with IDP.

oauthenticator/azuread.py
weisdd: @consideRatio 
> I think we should make these to be Set instead of List. Practically, if anyone passes a List, for example via a YAML configuration that is read and passed to the configuration - then it will be automatically casted to a Set so thats fine.

In other implementations for IDPs, I saw both `Set` and `List` being used, so wasn't sure which of those was the best fit. I'll change that to `Set` as per your recommendation.

> I'd also like to have it clarified how to list the app roles. Is it their display name rather than id? That seems a bit odd to me but that is what matches your example configuration, but perhaps it is required to be unique as well or assumed to not risk causing problems.

It's rather `value` that gets sent in the token claim. It's allowed to have the same `displayName`, but not the `value`. For the latter, Azure AD will throw an error like this:
`Failed to update jupyterhub application. Error detail: It contains duplicate value. Please Provide unique value.`
IDs are used by Azure AD internally and they're not exposed in the respective token claim.

> Perhaps a link or something as a reference to learn more about app roles this is relevant as well? I'm not sure, but in general I'm thinking the help text seem a bit too short and could be extended with another sentence or two.

Sure, I'll think about what I can add here.

oauthenticator/azuread.py
weisdd: Addressed in https://github.com/jupyterhub/oauthenticator/pull/446/commits/6dc41a6df1e479a7a7048b7ddc49630319da40ea

oauthenticator/azuread.py
weisdd: Moved to `Set` and improved help messages in https://github.com/jupyterhub/oauthenticator/pull/446/commits/6dc41a6df1e479a7a7048b7ddc49630319da40ea

weisdd:
> I updated the z2jh Helm chart config example with some notes
> 
> ```yaml
> hub:
>   config:
>     JupyterHub:
>       admin_access: true
>       authenticator_class: azuread
>     # NOTE: This must be a class that has these traitlets defined, which isn't the base Authenticator class
>     AzureAdOAuthenticator:
>       # NOTE: I think we should name this similar to `admin_users`, suggestion below
>       #       Hmm... Not sure on the name but I think we should include "admin_users" in it
>       app_roles_with_admin_users:
>         - admin
>       # NOTE: I think we should name this similar to `allowed_users`, suggestion below
>       # NOTE: I think we should make the admin app roles be assumed to be valid users as well
>       #       Hmm... Not sure on the name but I think we should include "allowed_users" in it
>       app_roles_with_allowed_users:
>         - user
> ```

@consideRatio What do you think about this naming?

```yaml
hub:
  config:
    JupyterHub:
      admin_access: true
      authenticator_class: azuread
    AzureAdOAuthenticator:
      admin_users_app_roles:
        - admin
      allowed_users_app_roles:
        - user
      revoke_stale_admin_privileges: true
```

I decided to add `revoke_stale_admin_privileges` (it's set to `False` by default) along the way ([code]( 
https://github.com/jupyterhub/oauthenticator/pull/446/files#diff-07a3788519a29315ed927e748c5afbd019255426618e35501f054ceb168d036cR125)). It should cover the following case:
JupyterHub remembers that a user is admin meaning even if `userdict["admin"]` is absent during login, the admin privileges are not revoked. That means that if we change user's app role in future (e.g. `admin` -> `user`), the user will still remain admin unless someone manually revokes privileges.
The flag is aimed to address that case.

```python
        # Admin privileges will be revoked if a user is not listed in admin_users
        # and then added back if the user has one of admin_users_app_roles.
        if self.revoke_stale_admin_privileges:
            if userdict["name"].lower() not in self.admin_users:
                userdict["admin"] = False

        if self.admin_users_app_roles:
            if check_user_has_role(roles, self.admin_users_app_roles):
                userdict["admin"] = True
                return userdict
```
A couple of notes:
* As I understand, JupyterHub always converts names to lowercase, that's why I used `userdict["name"].lower() not in self.admin_users:` an alternative would probably be:
```python
        if self.revoke_stale_admin_privileges:
            admin_users = [admin.lower() for admin in self.admin_users]
            if userdict["name"].lower() not in admin_users:
                userdict["admin"] = False
```
* I thought that making the flag optional can give users a chance to assign privileges manually if they want to.

> I consider this PR to be quite merge-ready with my comments are addressed. Oh btw, I'd also like a consideration if a test could be added to verify some logic of this or if its too complicated or similar.

Maybe we can only check the logic for assigning admin privileges, though the test is likely to be over-complicated unless the code is moved to a separate function. The latter will make the code look differently from other providers. So, not sure about that.

weisdd:
> This is quite neat, and I like how simple it is. It would certainly help our usecase.
> 
> A few questions:
> 
> * Will this work with B2C?
> * Does the tenant require an active subscription?

@thomafred Unfortunately, I'm don't know Azure deep enough to answer any of those questions (I mostly used Keycloak in the past). I can only say that as long as you're using Azure AD and can create app roles and assign users/usergroups to them, it all should work just fine. :)

manics:
> As I understand, JupyterHub always converts names to lowercase

That's the default but it's not mandatory, it's handled by an overridable function:
https://github.com/jupyterhub/jupyterhub/blob/59b25813705ef884e54661daf9c10bb48fa1fb54/jupyterhub/auth.py#L397-L407

consideRatio:
Awwww haha I was so happy that this was potentially ready for merge but now there is another feature to consider and such that may not fit under the title of this PR.

@weisdd would you mind waiting with adding the revoke admin status logic into a separate PR? I think it makes a lot of sense but I think it would be hard to help readers of a changelog or similar to manage understand the changes of the next release if it bundles into this PR for example.

weisdd:
> Awwww haha I was so happy that this was potentially ready for merge but now there is another feature to consider and such that may not fit under the title of this PR.
> 
> @weisdd would you mind waiting with adding the revoke admin status logic into a separate PR? I think it makes a lot of sense but I think it would be hard to help readers of a changelog or similar to manage understand the changes of the next release if it bundles into this PR for example.

@consideRatio Ah, I had thought it would be a good idea to have it included in this PR, though you're right, better to move discussion on it to a separate PR then. I've just removed the feature :)
Please, let me know if anything else should be adjusted here :)

weisdd:
> That's the default but it's not mandatory, it's handled by an overridable function:
> https://github.com/jupyterhub/jupyterhub/blob/59b25813705ef884e54661daf9c10bb48fa1fb54/jupyterhub/auth.py#L397-L407

@manics thanks for the info!

oauthenticator/azuread.py
```diff
@@ -19,6 +20,10 @@
 PYJWT_2 = V(jwt.__version__) >= V("2.0")
 
 
+def check_user_has_role(member_roles, allowed_roles):
```
consideRatio, line 23: Hmmm, this function name:

1. Let's prefix it with `_`
2. Let's reconsider naming of this function and its arguments.
   It receives a user's app roles and checks for any of the allowed roles, but from the name of the function as it is, I would expect a user and a role to be passed.
   So perhaps something more like `_check_if_any_user_role_is_an_allowed_role`?

Perhaps the readability is best improved by letting this one line function be inlined where its used rather than updating its name?

weisdd: @consideRatio I was just trying to create something similar to what I saw in google provider. There you have something more wordy than my one-liner:

```python
def check_user_in_groups(member_groups, allowed_groups):
    # Check if user is a member of any group in the allowed groups
    if any(g in member_groups for g in allowed_groups):
        return True  # user _is_ in group
    else:
        return False
```

I think maybe we should just get rid of `check_user_has_role()` at all.
Since we moved from `List` to `Set` in `AzureAdOAuthenticator`, how do you feel about something like this?

```python
        roles = set(auth_state['user'].get("roles", []))

        if self.admin_users_app_roles:
            if roles.intersection(self.admin_users_app_roles):
                userdict["admin"] = True
                return userdict

        if self.allowed_users_app_roles:
            if not roles.intersection(self.allowed_users_app_roles):
                return None
```

I know that `intersection` can be replaced with `&`, though I rarely get to see sets, so, IMO, `intersection` improves readability. I hope I'm not the only one who prefers it. :)

consideRatio: I fully agree on using intersection rather than `&` for readability :D I didn't know about `&` btw :)

consideRatio: > @consideRatio I was just trying to create something similar to what I saw in google provider.

Ah yeah, I'm not a fan of that implementation now that I know about it ;D The function name was not obvious, so an inline comment needed to correct for it etc. Sorry for having you need to experience review friction even though similar code have been accepted in the past.

+1 For not having such function at all and just letting the code speak for itself like you suggested above!

consideRatio (approved):
LGTM! Thanks for your patience with the review iterations @weisdd!

oauthenticator/azuread.py
weisdd: @consideRatio No worries :) It's great that we now have a better piece of code that can later be set as a good example for someone else. The review process was worth it for sure. Thanks for helping me on this!

weisdd:
@mafloh Could you, please, take a look at the PR? :) Thanks!

devhell:
> mafloh

Is it possible you meant @manics? :)

weisdd:
@devhell you're right, thanks :)

@manics could you, plz, take a look at the PR?
Thanks!

Santhin:
Any updates?

christian-sapconet:
@manics Would it possible to get this merged?

PR #448, opened 2021-07-23
[AzureAD] Add an implementation for Authenticator.authenticator_managed_groups=true

thomafred:
See [this](https://github.com/jupyterhub/jupyterhub/pull/3548) related PR in jupyterhub/jupyterhub-repo.

Added support for external user group management to Azure AD authenticator.

Usage example:

```python
"""sample jupyterhub config file for testing

configures jupyterhub with dummyauthenticator and simplespawner
to enable testing without administrative privileges.
"""

c = get_config()  # noqa
c.Application.log_level = 'DEBUG'

from oauthenticator.azuread import AzureAdOAuthenticator
import os

c.JupyterHub.authenticator_class = AzureAdOAuthenticator

c.AzureAdOAuthenticator.client_id = os.getenv("AAD_CLIENT_ID")
c.AzureAdOAuthenticator.client_secret = os.getenv("AAD_CLIENT_SECRET")
c.AzureAdOAuthenticator.oauth_callback_url = os.getenv("AAD_CALLBACK_URL")
c.AzureAdOAuthenticator.tenant_id = os.getenv("AAD_TENANT_ID")
c.AzureAdOAuthenticator.username_claim = "email"
c.AzureAdOAuthenticator.authorize_url = os.getenv("AAD_AUTHORIZE_URL")
c.AzureAdOAuthenticator.token_url = os.getenv("AAD_TOKEN_URL")
c.Authenticator.authenticator_managed_groups = True
c.Authenticator.refresh_pre_spawn = True

from jupyterhub.spawner import SimpleLocalProcessSpawner

c.JupyterHub.spawner_class = SimpleLocalProcessSpawner
```

### DIFF
--- a/oauthenticator/azuread.py
+++ b/oauthenticator/azuread.py
@@ -33,13 +33,13 @@ class AzureAdOAuthenticator(OAuthenticator):
     def _tenant_id_default(self):
         return os.environ.get('AAD_TENANT_ID', '')
 
-    username_claim = Unicode(config=True)
+    username_claim = Unicode(config=True, help="Name of claim containing username")
 
     @default('username_claim')
     def _username_claim_default(self):
         return 'name'
 
-    user_groups_claim = Unicode(config=True)
+    user_groups_claim = Unicode(config=True, help="Name of claim containing user group memberships")
 
     @default('user_groups_claim')
     def _user_groups_claim_default(self):
--- a/oauthenticator/azuread.py
+++ b/oauthenticator/azuread.py
@@ -10,7 +10,6 @@ from jupyterhub.auth import LocalAuthenticator
 from tornado.httpclient import HTTPRequest
 from traitlets import default
 from traitlets import Unicode
-from traitlets import Bool
 
 from .oauth2 import OAuthenticator
 
@@ -39,7 +38,9 @@ class AzureAdOAuthenticator(OAuthenticator):
     def _username_claim_default(self):
         return 'name'
 
-    user_groups_claim = Unicode(config=True, help="Name of claim containing user group memberships")
+    user_groups_claim = Unicode(
+        config=True, help="Name of claim containing user group memberships"
+    )
 
     @default('user_groups_claim')
     def _user_groups_claim_default(self):
@@ -106,7 +107,6 @@ class AzureAdOAuthenticator(OAuthenticator):
     def load_user_groups(self, auth_state):
         auth_user_state = auth_state["auth_state"]["user"]
         return auth_user_state[self.user_groups_claim]
-
 
 
 class LocalAzureAdOAuthenticator(LocalAuthenticator, AzureAdOAuthenticator):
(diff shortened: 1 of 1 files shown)


oauthenticator/azuread.py
```diff
@@ -38,6 +39,12 @@ def _tenant_id_default(self):
     def _username_claim_default(self):
         return 'name'
 
+    user_groups_claim = Unicode(config=True)
```
consideRatio, line 42: I'd like there to be a help string for this describing the configuration better in our documentation.

oauthenticator/azuread.py
```diff
@@ -96,6 +103,11 @@ async def authenticate(self, handler, data=None):
 
         return userdict
 
+    def load_user_groups(self, auth_state):
+        auth_user_state = auth_state["auth_state"]["user"]
+        return auth_user_state[self.user_groups_claim]
```
consideRatio, line 108: Can we assume `auth_user_state` always include a claim named `groups`? Does it imply that you have requested the scope `groups` as well or not?

consideRatio: Is the `load_user_groups` function an Authenticator base class function that is overridden?

consideRatio: Hmmm, what is calling this function? I find no reference to it.

consideRatio: Ah, this PR relies on https://github.com/jupyterhub/jupyterhub/pull/3548.

thomafred: @consideRatio - Unsure whether `auth_user_state` will return with a `groups`-claim.

The functionality in this PR requires `auth_users_state` to return some claim with a list of groups if `c.Authenticator.authenticator_managed_groups = True`. The claim name itself is configurable.

oauthenticator/azuread.py
thomafred: Done.

thomafred:
This is quite neat, and I like how simple it is. It would certainly help our usecase.

A few questions:
- Will this work with B2C?
- Does the tenant require an active subscription?

oauthenticator/azuread.py
sseppola: The `groups`-claim is non standard for OIDC, but AAD makes it available as an optional claim
https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-optional-claims. 

In our case we extend our AAD's policy to make API calls to fetch and return group membership, as shown in this example: https://github.com/azure-ad-b2c/samples/tree/master/policies/rest-api-idp

oauthenticator/azuread.py
```diff
@@ -32,12 +32,20 @@ class AzureAdOAuthenticator(OAuthenticator):
...
     def _username_claim_default(self):
         return 'name'
 
+    user_groups_claim = Unicode(
+        config=True, help="Name of claim containing user group memberships"
+    )
+
+    @default('user_groups_claim')
+    def _user_groups_claim_default(self):
+        return 'groups'
```
consideRatio, line 47: If you have a default that is fixed like this, you can have it as the first argument passed to the `Unicode` constructor above instead.

thomafred:
Abandoned in favor of this PR: https://github.com/jupyterhub/oauthenticator/pull/573

Issue #222, opened 2018-10-20
Pass along `expires_in` for GenericOAuthenticator


brasie:
I am wondering, would you consider passing along the `expires_in` value returned from the OAuth2 token response in the GenericOAuthenticator class' `authenticate` method?

It is a "recommended" OAuth2 token response field (https://tools.ietf.org/html/rfc6749#section-5.1) and I'd find it useful for determining when a token would need to be refreshed.

I think it'd simply be a matter of pulling it out of the token response here: https://github.com/jupyterhub/oauthenticator/blob/766b2afffee49493da83908cbc549054d52b15c7/oauthenticator/generic.py#L120-L123 e.g. by adding

    expires_in = resp_json.get('expires_in', None)

...and sticking it in the final `auth_state` like the other token response fields.

consideRatio:
I think this makes a lot of sense to do whenever we can, happy to review a PR about this.

mcmartins:
Hey! 

I can help and work on this one as this will be useful for us...

I think a simple change to @brasie 's proposal is needed. The `expires_in` within the response is used to store _'The lifetime in seconds of the access token'_, so it's only useful if we have the issued time, so I would include an `expiry_time` (timestamp) field instead, which is the current time plus the `expires_in` amount of seconds:
```
        ...
        now = time.time()
        expiry_time = now + resp_json.get('expires_in', None)
        ...
```
This way we can include a simple utility function to check if the `access_token` needs refreshing:
```
    @staticmethod
    def is_auth_token_expired(auth_state):
        now = time.time()
        expiry_time = auth_state.get('expiry_time', 0)
        if now < expiry_time:
            return False
        return True
```

Note: there will be always a short (milliseconds maybe) gap between the real expiry and the field introduced.

PR #456, opened 2021-09-14
[Generic] Pass along expires_in #222

mcmartins:
This pull request implements #222

I believe some documentation should be added, not really sure where though.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
 
   # Autoformat: markdown, yaml
   - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: v3.0.0-alpha.3
+    rev: v3.0.0-alpha.4
     hooks:
       - id: prettier
 


consideRatio:
Thank you!

My understanding is that we are now passing on `expiry_time` as authentication state, to me that sounds very reasonable and the implementation looks good. I wonder a bit about the naming of `expiry_time`, it is not clear from that name what the variable represents. Is it correct that it is a fraction of seconds until it expire?

I would love to have us align with a common practice for both the name of this kind of variable, and for the format of the data within it. Are you aware of common practices with regards to this @mcmartins?

mcmartins:
I agree, _expiry_time_ is not a good name for this field. 

So, this field contains the expiration time after which the token is not accepted for processing and is expressed in seconds since the epoch as a floating point number.

Maybe we could rename it to _expires_at_? Feels a bit more natural as the response token as already a _expires_in_ field.

As far as I know there's not a common field name for it on any of the drafts. In JWT headers is called "exp" which stands to "Expiration Time".

I'm happy to change it for any agreed name ;)

mcmartins:
Changed the property to _expires_at_ to make it more consistent with the kind of variable, as per the discussion mentioned on: https://oauth.net/faq/

PR #457, opened 2021-09-17
Fix My Service authenticator class names in documentation

sgaist:
The example in the documentation currently uses and mentions GitHub in the `OAuthenticator` subclasses  which makes it inconsistent with the rest of the file.

This patches updates all the references from GitHub to My Service.


consideRatio:
Thank you @sgaist !!!

Issue #136, opened 2017-10-13
Would it be possible to have multiple authentication methods simultaneously?


stefanonicotri:
Hi,
is it possible to setup multiple authentication methods simultaneously?
For example, I would like my JupyterHub instance to have Google, GitHub and Local authentication method on the same login page: is it possible?
Many thanks

minrk:
This should be possible, but it is not ready to go at the moment. A `MultiAuthenticator` could have various sub-authenticators and dispatch to them. The trick would be hooking up the various URLs needed.

stefanonicotri:
Thank you very much for your suggestion :)

zonca:
I would like to have this feature as well, even just for the authentication classes in `oauthenticator`...I don't have time right now to work on this myself, but I could be a beta-tester! ;-)

jm2004:
Same here. Our project also needs more than one way to do the authentication in different situations.
I am now trying to implement my own. But it definitely great if we have this MultiAuthenticator.

Thanks!

cmseal:
I'd like to second this... We use FreeIPA for our team, and LDAP works fine there, but our parent company uses Google Apps.  Being able to auth with either of these would be awesome.  I'm looking at how it might work, but has anyone else made any progress with this? Cheers!

betatim:
Might be worth checking out https://www.keycloak.org/ as well as investigating a `MultiAuthenticator`.

ablekh:
In addition to Keycloak, there is also Gluu (http://gluu.org). Also, _"a quick and ready-to-use solution off the top of my head seems to be using JupyterHub Globus authenticator, which offers Globus, OpenID, Google, SAML/CILogon and E-mail authentication in a single package. Unfortunately, it is far from ideal due to missing GitHub/GitLab integration and adding Globus infrastructure dependency"_ (from my post on Gitter last night).

@betatim For `MultiAuthenticator`, are you talking about this code: https://gist.github.com/danizen/78111676530738fcbca8d8ad87c56690?

cjreyn:
+1 for this

ablekh:
@zhiyuli Nice work, though a more detailed README would be appreciated. However, it seems that your approach only covers OAuth-based authenticators, whereas the original code (that I have linked above) is more generic, allowing for any authenticators (assuming it's relevantly updated). Am I correct on this?

zhiyuli:
[...]

PR #459, opened 2021-09-23
Implement MultiOAuthenticator

sgaist:
This authenticator allows to use several different services to authenticate to one instance of JupyterHub.

The code is based on the suggested implementation that can be found on #136 and more specifically https://github.com/jupyterhub/oauthenticator/issues/136#issuecomment-749931724.

Due to the merge of https://github.com/jupyterhub/jupyterhub/pull/3315, the support for showing multiple links is already supported.

This is the first step as discussed in this [JupyterHub discourse thread](https://discourse.jupyter.org/t/make-jupyterhub-authentication-pluggable/10122/2?u=sgaist).

### DIFF
--- a/oauthenticator/multioauthenticator.py
+++ b/oauthenticator/multioauthenticator.py
@@ -1,0 +1,95 @@
+"""
+Custom Authenticator to use multiple OAuth providers with JupyterHub
+
+Example of configuration:
+
+    from oauthenticator.github import GitHubOAuthenticator
+    from oauthenticator.google import GoogleOAuthenticator
+
+    c.MultiOAuthenticator.authenticators = [
+        (GitHubOAuthenticator, '/github', {
+            'client_id': 'xxxx',
+            'client_secret': 'xxxx',
+            'oauth_callback_url': 'http://example.com/hub/github/oauth_callback'
+        }),
+        (GoogleOAuthenticator, '/google', {
+            'client_id': 'xxxx',
+            'client_secret': 'xxxx',
+            'oauth_callback_url': 'http://example.com/hub/google/oauth_callback'
+        })
+    ]
+
+    c.JupyterHub.authenticator_class = 'oauthenticator.multioauthenticator.MultiOAuthenticator'
+
+The same Authenticator class can be used several to support different providers.
+
+"""
+from jupyterhub.auth import Authenticator
+from jupyterhub.utils import url_path_join
+from traitlets import List
+
+
+class MultiOAuthenticator(Authenticator):
+    """Wrapper class that allows to use more than one authentication provider
+    for JupyterHub"""
+
+    authenticators = List(help="The subauthenticators to use", config=True)
+
+    def __init__(self, *arg, **kwargs):
+        super().__init__(*arg, **kwargs)
+        self._authenticators = []
+        for (
+            authenticator_klass,
+            url_scope,
+            authenticator_configuration,
+        ) in self.authenticators:
+            configuration = self.trait_values()
+            # Remove this one as it will overwrite the value if the authenticator_klass
+            # makes it configurable and the default value is used (take a look at
+            # GoogleOAuthenticator for example).
+            configuration.pop("login_service")
+            configuration.update(authenticator_configuration)
+            self._authenticators.append(
+                {
+                    "instance": authenticator_klass(**configuration),
+                    "url_scope": url_scope,
+                }
+            )
+
+    def get_custom_html(self, base_url):
+        """Re-implementation generating one login button per configured authenticator"""
+
+        html = []
+        for authenticator in self._authenticators:
+            login_service = authenticator["instance"].login_service
+            url = url_path_join(base_url, authenticator["url_scope"], "oauth_login")
+
+            html.append(
+                f"""
+                <div class="service-login">
+                  <a role="button" class='btn btn-jupyter btn-lg' href='{url}'>
+                    Sign in with {login_service}
+                  </a>
+                </div>
+                """
+            )
+        return "\n".join(html)
+
+    def get_handlers(self, app):
+        """Re-implementation that will return the handlers for all configured
+        authenticators"""
+
+        routes = []
+        for _authenticator in self._authenticators:
+            for path, handler in _authenticator["instance"].get_handlers(app):
+
+                class WrapperHandler(handler):
+                    """'Real' handler configured for each authenticator. This allows
+                    to reuse the same authenticator class configured for different
+                    services (for example GitLab.com, gitlab.example.com)
+                    """
+
+                    authenticator = _authenticator["instance"]
+
+                routes.append((f'{_authenticator["url_scope"]}{path}', WrapperHandler))
+        return routes
--- a/oauthenticator/tests/test_multioauthenticator.py
+++ b/oauthenticator/tests/test_multioauthenticator.py
@@ -1,0 +1,92 @@
+"""Test module for the MultiOAuthenticator class"""
+from pytest import fixture
+
+from ..github import GitHubOAuthenticator
+from ..gitlab import GitLabOAuthenticator
+from ..google import GoogleOAuthenticator
+from ..multioauthenticator import MultiOAuthenticator
+
+
+@fixture
+def different_authenticators():
+    return [
+        (
+            GitLabOAuthenticator,
+            "/gitlab",
+            {
+                "client_id": "xxxx",
+                "client_secret": "xxxx",
+                "oauth_callback_url": "http://example.com/hub/gitlab/oauth_callback",
+            },
+        ),
+        (
+            GitHubOAuthenticator,
+            "/github",
+            {
+                "client_id": "xxxx",
+                "client_secret": "xxxx",
+                "oauth_callback_url": "http://example.com/hub/github/oauth_callback",
+            },
+        ),
+    ]
+
+
+@fixture
+def same_authenticators():
+    return [
+        (
+            GoogleOAuthenticator,
+            "/mygoogle",
+            {
+                "login_service": "My Google",
+                "client_id": "yyyyy",
+                "client_secret": "yyyyy",
+                "oauth_callback_url": "http://example.com/hub/mygoogle/oauth_callback",
+            },
+        ),
+        (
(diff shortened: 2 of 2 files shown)


manics:
Does `MultiOAuthenticator` only work with `OAuthenticator` derived authenticators, or would it work with others?

sgaist:
It's a good point you have here. I haven't thought about them.

As it is `MultiOAuthenticator` handles any authenticator that implements the `Authenticator` interface. It currently does not make use of any implementation specific details of `OAuthenticator`.

That said, there are some issues to address for the "local" authenticators:

1) The local authenticators have no "name" to show on the button so we could make use of the `login_service` configuration to give them one to generate the button in the custom html. But the `login.html` template makes use of `login_service` to determine whether it should render a button or the login form so when we get there it will render a single button that proposes to login with the authenticator selected and then back to the list of buttons when clicked.
2) The next issue is the login url used for the post action which comes from the settings rather than using the one from the authenticator.
3) Following point 2, the `authenticator_login_url` generated for the template will be wrong as the authenticator has no idea of the `url_scope` and thus will return an invalid value. 

For number 2, a change in the `login.html` template to use `authenticator_login_url` seems to be enough.

For number 3, I have tested an approach using a mixin class to overwrite `login_url`, `logout_url` and `get_handlers`. It is used to dynamically create a subclasses that is configured with the scope and appends it to the `base_url`. This simplifies the rest of the code a bit as well.

For number 1, it's a bit more tricky. One possible solution could be to handle a MultiAuthenticator specific "service name" configuration parameter that could be used in place of `login_service` when generating the buttons. This would allow to not modify the base class. We could also document the requirement to create a simple subclass adding that name.

Should I push the proof of concept on top of this merge request so you can compare the two implementation ?

sgaist:
Small ping :-)

abdelq:
Another ping. This is useful.

Ph0tonic:
Hello,
Any updates or plan to integrate this feature ?
\cc @consideRatio and @GeorgianaElena

Ph0tonic:
Hi, after looking at the proposition, I wonder whether this class should be incorporated into `Jupyterhub`. As it supports any Authenticator class, it might serve other authenticators defined in other libraries.

\cc @sgaist and @manics

Ph0tonic:
Hi,
After looking at this proposition, I wonder whether this code should be incorporated into `jupyterhub`. As it might be used by other authenticators not provided in this library.
\cc @sgaist and @manics

sgaist:
@Ph0tonic, while it was originally geared towards OAuth providers, it something that could indeed live in JupyterHub since it also supports other providers as it is not tied to the implementations of OAuth authenticators.

Maybe @minrk might want to also weigh in.

oauthenticator/multiauthenticator.py
```diff
@@ -0,0 +1,124 @@
...
+"""
+from jupyterhub.auth import Authenticator
+from jupyterhub.utils import url_path_join
+from traitlets import List
+
+
+class URLScopeMixin(object):
+    """Mixin class that adds the"""
+
+    scope = ""
```
Ph0tonic, line 36: Should be renamed to avoid colision with `scope` variable defined in `oauth2`, like this, it breaks with gitlab.
```suggestion
    url_scope = ""
```

oauthenticator/multiauthenticator.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def login_url(self, base_url):
+        return super().login_url(url_path_join(base_url, self.scope))
```
Ph0tonic, line 39:
```suggestion
        return super().login_url(url_path_join(base_url, self.url_scope))
```

oauthenticator/multiauthenticator.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def logout_url(self, base_url):
+        return super().logout_url(url_path_join(base_url, self.scope))
```
Ph0tonic, line 42:
```suggestion
        return super().logout_url(url_path_join(base_url, self.url_scope))
```

oauthenticator/multiauthenticator.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def get_handlers(self, app):
+        handlers = super().get_handlers(app)
+        return [
+            (url_path_join(self.scope, path), handler) for path, handler in handlers
```
Ph0tonic, line 47:
```suggestion
            (url_path_join(self.url_scope, path), handler) for path, handler in handlers
```

oauthenticator/multiauthenticator.py
```diff
@@ -0,0 +1,124 @@
...
+    for JupyterHub"""
+
+    authenticators = List(help="The subauthenticators to use", config=True)
+
+    def __init__(self, *arg, **kwargs):
+        super().__init__(*arg, **kwargs)
+        self._authenticators = []
+        for (
+            authenticator_klass,
+            url_scope,
```
Ph0tonic, line 62:
```suggestion
            url_scope_authenticator,
```

oauthenticator/multiauthenticator.py
```diff
@@ -0,0 +1,124 @@
...
+            authenticator_configuration,
+        ) in self.authenticators:
+            configuration = self.trait_values()
+            # Remove this one as it will overwrite the value if the authenticator_klass
+            # makes it configurable and the default value is used (take a look at
+            # GoogleOAuthenticator for example).
+            configuration.pop("login_service")
+
+            class WrapperAuthenticator(URLScopeMixin, authenticator_klass):
+                scope = url_scope
```
Ph0tonic, line 72:
```suggestion
                url_scope = url_scope_authenticator
```

oauthenticator/multiauthenticator.py
```diff
@@ -0,0 +1,124 @@
...
+
+            service_name = authenticator_configuration.pop("service_name", None)
+            configuration.update(authenticator_configuration)
+
+            authenticator = WrapperAuthenticator(**configuration)
+
```
Ph0tonic, line 78: When providing additional config values for example (`allowed_users`) :
```
(GitHubOAuthenticator, '/github', {
  'service_name': 'github',
  'client_id': '...',
  'client_secret': '...',
  'allowed_users': {'rene.dumont'},
}),
```

Those values are not set in the created Authenticator. I am not sure for which reason. I found a way to fix this issue, see below but I am sure there is a better way linked with the way `configuration` is built.

```suggestion

            traits = authenticator.traits()
            for key, value in authenticator_configuration.items():
                trait = traits.get(key, None)
                if hasattr(authenticator, key) and trait and trait.this_class == Authenticator and trait.metadata.get('config', False):
                    setattr(authenticator, key, value)

```

Ph0tonic (changes requested):
Hello,

Yes, indeed, let's see what maintainers think of it.

I tested your code locally with a double configuration `Gitlab` and `Github` and found a few issues.

Happy to help and provide more information if needed.

oauthenticator/multiauthenticator.py
```diff
@@ -0,0 +1,124 @@
...
+        ) in self.authenticators:
+            configuration = self.trait_values()
+            # Remove this one as it will overwrite the value if the authenticator_klass
+            # makes it configurable and the default value is used (take a look at
+            # GoogleOAuthenticator for example).
+            configuration.pop("login_service")
+
+            class WrapperAuthenticator(URLScopeMixin, authenticator_klass):
+                scope = url_scope
+
```
Ph0tonic, line 73: The username conflict might also be managed in the wrapper by adding a prefix to the username.
I managed to make it work with the following code:

```python
username_prefix = service_name+'_'

async def authenticate(self, handler, data=None, **kwargs):
    response = await super().authenticate(handler, data, **kwargs)
    if response is None:
        return None
    elif type(response) == str:
        return self.username_prefix+response
    else:
        response['name'] = self.username_prefix+response['name']
        return response

# def normalize_username(self, username):
#     print("normalize username :",username, super().normalize_username(username))
#     return self.username_prefix+super().normalize_username(username)

def check_allowed(self, username, authentication=None):
    print("check allowed provided :",username)
    return super().check_allowed(username.removeprefix(self.username_prefix), authentication)

def check_blocked_users(self, username, authentication=None):
    print("check blocked provided :",username)
    return super().check_allowed(username.removeprefix(self.username_prefix), authentication)
```

As you can see, I initially tried to add the prefix by overriding `normalize_username` but it appears to be called inside `oauth2`. One drawback is the display of this username's prefix in the upper corner of jupyterhub.

Ph0tonic:
Hi, any update on this PR ? Should we make a PR in jupyterhub ? We could manage conflicting users in a different way if it was managed in jupyterhub, by configuring an authenticator id and by storing it with the username in the db.

manics:
My vote is to put this in a new repository/package that has oauthenticator as a dependency, collaboratively work on it there, test it in production, and then review whether some or all of it should be brought into JupyterHub.

sgaist:
Hi guys,

Sorry for the late reply. The plan sounds good to me. 

@manics would that repo be under the jupyterhub organisation or my own space ?

I have done some preparation work taking my fork of oauthenticator apart to keep only the MultiAuthenticator class to follow the same project structure. The only part that might have some issue is the GitHub pipeline as I haven't used them yet actively.

Ph0tonic:
> Hi guys,
> 
> Sorry for the late reply. The plan sounds good to me.
> 
> @manics would that repo be under the jupyterhub organisation or my own space ?
> 
> I have done some preparation work taking my fork of oauthenticator apart to keep only the MultiAuthenticator class to follow the same project structure. The only part that might have some issue is the GitHub pipeline as I haven't used them yet actively.

I am available to help and test, let me know once you have a new repo 👍

manics:
@sgaist Thanks for working on this! Starting with your own personal repo will be the easiest way to start, and we can transfer it later if we want. Personally I think something `jupyterhub-contrib` would be a good home but that's still being discussed https://github.com/jupyterhub/team-compass/issues/519

sgaist:
@manics You're welcome !

`jupyterhub-contrib` sounds good. I was thinking about something similar so it's great to see the idea was already floating around.

I'll create the repo during the week.

Issue #432, opened 2021-05-06
Test failures


fabaff:
### Bug description
Not all tests are passing with  Python 3.8.9 and `pytest-6.2.3`. This happens during the build process for NixOS packages.

```bash
============================= test session starts ==============================
platform linux -- Python 3.8.9, pytest-6.2.3, py-1.10.0, pluggy-0.13.1
rootdir: /build/oauthenticator-14.0.0
plugins: asyncio-0.14.0, requests-mock-1.8.0
collected 58 items                                                             

oauthenticator/tests/test_auth0.py .                                     [  1%]
oauthenticator/tests/test_azuread.py .FFFF                               [ 10%]
oauthenticator/tests/test_bitbucket.py ...                               [ 15%]
oauthenticator/tests/test_cilogon.py .....                               [ 24%]
oauthenticator/tests/test_generic.py ........                            [ 37%]
oauthenticator/tests/test_github.py ...                                  [ 43%]
oauthenticator/tests/test_gitlab.py ....                                 [ 50%]
oauthenticator/tests/test_globus.py .........                            [ 65%]
oauthenticator/tests/test_google.py ........                             [ 79%]
oauthenticator/tests/test_mediawiki.py F.                                [ 82%]
oauthenticator/tests/test_oauth2.py ..                                   [ 86%]
oauthenticator/tests/test_okpy.py ..                                     [ 89%]
oauthenticator/tests/test_openshift.py ......                            [100%]

=================================== FAILURES ===================================
```

<details>

```bash
______________________________ test_azuread[None] ______________________________

username_claim = None
azure_client = <oauthenticator.tests.mocks.MockAsyncHTTPClient object at 0x7ffff4c40fd0>

    @pytest.mark.parametrize(
        'username_claim',
        [
            None,
            'name',
            'oid',
            'preferred_username',
        ],
    )
    async def test_azuread(username_claim, azure_client):
        authenticator = AzureAdOAuthenticator(
            tenant_id=str(uuid.uuid1()),
            client_id=str(uuid.uuid1()),
            client_secret=str(uuid.uuid1()),
        )
        if username_claim:
            authenticator.username_claim = username_claim

        handler = azure_client.handler_for_user(
>           user_model(
                tenant_id=authenticator.tenant_id,
                client_id=authenticator.client_id,
                name="somebody",
            )
        )

oauthenticator/tests/test_azuread.py:82:
[...]

PR #461, opened 2021-10-04
Support pyjwt >= 2 in tests

diego-plan9:
Allow the tests to be aware of the `pyjwt` version installed, converting the result of `jwt.encode()` from bytes to string and vice-versa accordingly as the return type changes between pyjwt versions.

### Details

It seems that `pyjwt` changed the return type between major versions, and running the tests with `pyjwt>=2.0` seems to result in #432 (I have been able to replicate the same errors). Even if `mwoauth` still pins `1.7.x` (although [might be able to use 2.x](https://github.com/mediawiki-utilities/python-mwoauth/pull/43) ) and as a result 2.x is never really installed in CI or a regular install, this PR tries to follow-up on the (guessed) rationale of #402 and #420 and allow running the test suite to support pyjwt 2.x.

### DIFF
--- a/oauthenticator/tests/test_azuread.py
+++ b/oauthenticator/tests/test_azuread.py
@@ -9,6 +9,7 @@ import jwt
 import pytest
 
 from ..azuread import AzureAdOAuthenticator
+from ..azuread import PYJWT_2
 from .mocks import setup_oauth_mock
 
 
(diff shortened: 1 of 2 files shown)


consideRatio:
@rragundez as the author of #402 with relevant background knowledge, would you be willing to help out with a review on this PR?

Thank you @diego-plan9 and @rragundez for your contributions! :heart: :tada:

minrk (approved):

Issue #398, opened 2020-12-14
[OAuth or Generic] Implement refresh_user


minrk:
### Proposed change

Implement JupyterHub's [refresh_user](https://jupyterhub.readthedocs.io/en/stable/api/auth.html#jupyterhub.auth.Authenticator.refresh_user) method to refresh /re-validate auth state. The upstream method was added specifically to allow Authenticators like oauthenticator to ensure auth state is up-to-date with upstream prior to spawn to avoid stale auth info.

### Who would use this feature?

Deployments where user environments are passed access tokens from auth state, especially those where access tokens expire.

### (Optional): Suggest a solution

1. Ideally, we would refactor the access_token -> auth model into a common method(s) in all OAuthenticators, e.g. `token_to_user(token)`. I'm not sure how generic this is going to be.
2. validate current token to make sure it's still valid (maybe new base `check_token` method)?
3. refresh with refresh_token, if available
4. return None if it's not valid and not refreshable (forces re-login)
5. return updated auth model, if valid in case of changes, e.g. group/admin changes
6. enable refresh_pre_spawn by default if self.enable_auth_state.

jabbera:
@minrk I'm thinking this might be useful to me and I might implement it, but it's not clear if it will solve my problem. Maybe you can clear it up?

I user Azure AD auth paired with kubespawner. I pass my access token to my notebooks hosts and the access token goes bad after 10 hours. If I request offline_access in my scope I can get a refresh token. However the refresh_user method is called in hubs context I believe? How can this ip me update my access token in my notebook?

minrk:
This won't, or at least isn't enough on its own, to solve the problem in the container.

If you pass the refresh token to the container, you can do the refresh from there, e.g. in a server extension or kernel-level code, depending on how you want to use it. I believe @betatim has had a working version of refreshing tokens for some provider.

GeorgianaElena:
Hey @Wykiki and @jthiltges! Thank you so much for your PRs https://github.com/jupyterhub/oauthenticator/pull/490 and https://github.com/jupyterhub/oauthenticator/pull/475 that implement the `refresh_user`.

In a recent maintenance effort, the repository has been refactored (https://github.com/jupyterhub/oauthenticator/pull/526) and the oauthenticators now use a common authentication model that more or less match @minrk's first point in the [suggestion solution from the top comment](https://github.com/jupyterhub/oauthenticator/issues/398#issue-766219499).

[...]

PR #475, opened 2021-12-28
[Generic] Implement refresh_user to periodically refresh OAuth tokens

jthiltges:
An attempt at OAuth token refresh, following the pattern in PR https://github.com/jupyterhub/oauthenticator/pull/287, without relying on tokens to be JWTs. This should address at least some of issue https://github.com/jupyterhub/oauthenticator/issues/398.

Requirements for token refresh:

- OAuth server response includes `expires_in` and `refresh_token`
- State is persisted on hub with `c.Authenticator.enable_auth_state = True` (to store the `refresh_token`)

### DIFF
--- a/oauthenticator/generic.py
+++ b/oauthenticator/generic.py
@@ -3,6 +3,7 @@ Custom Authenticator to use generic OAuth2 with JupyterHub
 """
 import base64
 import os
+import time
 from urllib.parse import urlencode
 
 from jupyterhub.auth import LocalAuthenticator
@@ -143,11 +144,18 @@ class GenericOAuthenticator(OAuthenticator):
         if isinstance(scope, str):
             scope = scope.split(' ')
 
+        try:
+            expires_in = int(token_response.get('expires_in'))
+            expires_at = time.time() + expires_in # seconds
+        except (KeyError, TypeError):
+            expires_at = None
+
         return {
             'access_token': access_token,
             'refresh_token': refresh_token,
             'oauth_user': user_data_response,
             'scope': scope,
+            'expires_at': expires_at,
         }
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


kkaraivanov1:
As in the next release of jupyterhub there will be authenticator user group management https://github.com/jupyterhub/jupyterhub/pull/3548 what happens if we have manage_groups=true  and the user's group membership changes after auth and before refresh? Your refresh code is not updating user_info['groups'] as it seem.

jthiltges:
@kkaraivanov1 Thanks for looking at the PR. I agree that group data should be refreshed as well.

Is group membership info already being stored on authentication?

kkaraivanov1:
The way I understand the intentions behind the PR I mention is  if c.Authenticator.manage_groups is set AND if the GenericOAuthenticator adds the groups to user_info['groups']  they will be populated automatically. 
So to answer your question - it could be 
Perhaps you can hide that behind a check like `if hasattr(self,manage_groups):  .......`. Please take my thoughts with big grain of salt, I'm not a dev and my python is rudimentary.

jthiltges:
That makes sense, and thank you for the clarification.

Adding support for `user_info['groups']` is probably best addressed in a separate PR. I'll be happy to revise this PR to follow the pattern that the project maintainers prefer.

oauthenticator/generic.py
```diff
@@ -216,6 +224,46 @@ async def authenticate(self, handler, data=None):
...
+
+        # Attempt renewal
+        params = dict(
+            client_id=self.client_id,
+            client_secret=self.client_secret,
+            refresh_token=refresh_token,
+            grant_type='refresh_token',
+        )
+
+        headers = self._get_headers()
```
kkaraivanov1, line 256: this one is wrong. It has to be 
`headers = { "Content-Type": "application/x-www-form-urlencoded" }`

PR #482, opened 2022-01-30
changes to update the requirements.txt file.

prakashcc:
This request is to update the requirements.txt file. I was making some changes on my side so I faced this issue and so raised my request to fix it.

Thank you.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,11 @@
 jupyterhub>=1.2
+entrypoints~=0.3
+Jinja2~=3.0.3
+tornado~=6.1
+setuptools~=57.0.0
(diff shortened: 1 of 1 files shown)


manics (changes requested):
Hi! This shouldn't be necessary. Please could you explain the problem you had, and provide us with the exact steps to reproduce the problem? Thanks!

prakashcc:
> Hi! This shouldn't be necessary. Please could you explain the problem you had, and provide us with the exact steps to reproduce the problem? Thanks!

Hi @manics, I tried to run pip install -r requirements.txt but it did not work for me because other library was not present.

Issue #414, opened 2021-02-14
[OAuth] Custom 403 error page after successful login at OAuth provider but rejection based on config


1kastner:
### Proposed change

Add the option to show a customized 403 error message when after a successful login at an OAuth provider the login is still rejected based on the config, e.g. the `allowed_users ` list or the `blocked_users` list.

### Alternative options

After installation, each JupyterHub administrator can tamper around with the code at https://github.com/jupyterhub/oauthenticator/blob/a4aac191c16cf6281f3d346615aefa75702b02d7/oauthenticator/oauth2.py#L226-L227

### Who would use this feature?

Especially if several JupyterHubs are deployed in one institution, users could accidentially try to login at the wrong server. This could be e.g. the case if there is one JupyterHub per course. For each course, only enrolled students of that specific course would be on the `allowed_users` list. Here, after a failed login some words could help to avoid confusion. As especially non-IT-people wouldn't be able to decypher the meaning of 403, that could result in a lot of unnecessary consultation.

### (Optional): Suggest a solution

The todo at https://github.com/jupyterhub/oauthenticator/blob/a4aac191c16cf6281f3d346615aefa75702b02d7/oauthenticator/oauth2.py#L226-L227 could be done. Similar to https://github.com/jupyterhub/jupyterhub/blob/4e7936056744cdad31d608388a349207196efa56/jupyterhub/auth.py#L180-L186, in the JupyterHub config the JupyterHub administrator can set an case-specific html snippet.

See https://github.com/jupyterhub/jupyterhub/issues/3356 for alternative ideas.

1kastner:
If there is some interest, I could submit a PR.

minrk:
@1kastner that sounds great!

PR #484, opened 2022-02-05
Allow and document custom 403 messages and pages

GeorgianaElena:
This PR:

- adds a config option that allows modifying the message that is shown to the user when a 403 error is raised
403-default-appearance
- documents this custom message option
- shows an example `403.html` that can extend the default hub templates. This example template uses the nw custom 403 message and also adds a button that redirects users to the hub logout endpoint. It looks like this rendered:
403-message
- exemplifies extending hub templates and links to the appropriate hub docs

Fixes https://github.com/jupyterhub/oauthenticator/issues/414
Ref: https://github.com/2i2c-org/pilot-homepage/pull/8

### DIFF
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -230,8 +230,8 @@ class OAuthCallbackHandler(BaseHandler):
         self.check_arguments()
         user = await self.login_user()
         if user is None:
-            # todo: custom error page?
-            raise web.HTTPError(403)
+            raise web.HTTPError(403, self.authenticator.custom_403_message)
+
         self.redirect(self.get_next_url(user))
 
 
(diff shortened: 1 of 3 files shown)


oauthenticator/oauth2.py
```diff
@@ -292,6 +292,14 @@ def _userdata_url_default(self):
     def _logout_redirect_url_default(self):
         return os.getenv("OAUTH_LOGOUT_REDIRECT_URL", "")
 
+    custom_403_message = Unicode(
+        config=True, help="""The message to be shown when user was not allowed"""
```
yuvipanda, line 296: Can we put the default message here as the traitlet's first arg (default) rather than use a method for this?

Also I think most installations do not use an allow-list explicitly. So perhaps the message can be more generic? "Sorry, you are not currently authorized to use this hub. Please contact the hub administrator"?

GeorgianaElena: Done in https://github.com/jupyterhub/oauthenticator/pull/484/commits/1d6c6baf4c0cf0d552314d46cf2727d788c78f5a. Thanks a lot for the review @yuvipanda ✨

yuvipanda:
Thanks, @GeorgianaElena!

PR #485, opened 2022-02-08
Implementation of Apache Airavata Custos Authenticator

isururanawaka:
Custos is a  security middleware for science gateways. See https://airavata.apache.org/custos/ for additional information.

This PR contains the implementation of Custos Authenticator for Jupyterhub as an OIDC plugin.


smarru:
Thanks @consideRatio yes it makes sense. We will create a new repo with oauthenticator as a dependency and make PR to documentation.

PR #489, opened 2022-02-24
[Generic] Support custom ca_certs bundle in GenericOAuthenticator

nngo:
Add support to use a custom certificate authority (CA) bundle for talking to GenericOAuthenticator (OAuth2) and overridable via `OAUTH2_CA_CERTS` environment variable without having to disable certificate validation with `OAUTH2_TLS_VERIFY=0` env.

Otherwise get the following error:
  - tornado.curl_httpclient.CurlError: HTTP 599: SSL certificate problem: unable to get local issuer certificate

### DIFF
--- a/oauthenticator/generic.py
+++ b/oauthenticator/generic.py
@@ -84,10 +84,18 @@ class GenericOAuthenticator(OAuthenticator):
         help="Disable basic authentication for access token request",
     )
 
+    ca_certs = Unicode(
+        os.environ.get('OAUTH2_CA_CERTS', os.environ.get('SSL_CERT_FILE')),
+        allow_none=True,
+        config=True,
+        help="Custom trusted Certificate Authorities bundle in PEM format (ca_certs) or None to use defaults. Default: None",
+    )
+
     @default("http_client")
     def _default_http_client(self):
         return AsyncHTTPClient(
-            force_instance=True, defaults=dict(validate_cert=self.tls_verify)
+            force_instance=True,
+            defaults=dict(ca_certs=self.ca_certs, validate_cert=self.tls_verify),
         )
 
     def _get_headers(self):
--- a/oauthenticator/generic.py
+++ b/oauthenticator/generic.py
@@ -85,10 +85,14 @@ class GenericOAuthenticator(OAuthenticator):
     )
 
     ca_certs = Unicode(
-        os.environ.get('OAUTH2_CA_CERTS', os.environ.get('SSL_CERT_FILE')),
+        os.environ.get('OAUTH2_CA_CERTS', os.environ.get('SSL_CERT_FILE', None)),
         allow_none=True,
         config=True,
-        help="Custom trusted Certificate Authorities bundle in PEM format (ca_certs) or None to use defaults. Default: None",
+        help="""
+        Custom trusted Certificate Authorities bundle in PEM format (ca_certs). Default: None (to use system trusted CAs).
+
+        Can be overriden with OAUTH2_CA_CERTS (higher precedence) or SSL_CERT_FILE environment variable.
+        """,
     )
 
     @default("http_client")
--- a/oauthenticator/tests/test_generic.py
+++ b/oauthenticator/tests/test_generic.py
@@ -167,3 +167,14 @@ async def test_generic_callable_groups_claim_key_with_allowed_groups_and_admin_g
     user_info = await authenticator.authenticate(handler)
     assert user_info['name'] == 'zoe'
     assert user_info['admin'] is True
+
+
+def test_generic_ca_certs():
+    authenticator = GenericOAuthenticator()
+    assert authenticator.http_client.defaults.get('ca_certs') == None
+
+    ca_certs_bundle = '/test/ca.bundle'
+    authenticator = GenericOAuthenticator(
+        ca_certs=ca_certs_bundle,
+    )
+    assert authenticator.http_client.defaults.get('ca_certs') == ca_certs_bundle


oauthenticator/generic.py
```diff
@@ -84,10 +84,18 @@ class GenericOAuthenticator(OAuthenticator):
         help="Disable basic authentication for access token request",
     )
 
+    ca_certs = Unicode(
+        os.environ.get('OAUTH2_CA_CERTS', os.environ.get('SSL_CERT_FILE')),
```
GeorgianaElena, line 88: I'm not sure if defaulting to `SSL_CERT_FILE` is a good idea if `OAUTH2_CA_CERTS`

 I'm thinking of a scenario where `SSL_CERT_FILE` is set to a path and `OAUTH2_CA_CERTS` should be a different one, like the default system cert location. Can something like this be possible?

GeorgianaElena (commented):
From what I understand, this allows setting an option already present in `AsyncHTTPClient`, so this looks good to me @nngo!

The only concern I have is if whether this is a bit hacky and the right behaviour of `AsyncHTTPClient` can be achieved by a proper config of the default certs that torndado looks at.

Also, this tornado error was reported for other authenticators too https://github.com/jupyterhub/oauthenticator/search?q=SSL+certificate+problem%3A+unable+to+get+local+issuer+certificate&type=issues, which makes me wonder if we should implement this in oauth2 directly.

cc @consideRatio and @manics who might have more insight about this.

oauthenticator/generic.py
```diff
@@ -84,10 +84,18 @@ class GenericOAuthenticator(OAuthenticator):
...
+        allow_none=True,
+        config=True,
+        help="Custom trusted Certificate Authorities bundle in PEM format (ca_certs) or None to use defaults. Default: None",
```
consideRatio, line 91: `or None to use defaults. Default: None",` is a bit confusing. I think you need to be explicit that "to use defaults" refer to fallback on the two separate environment variables you have mentioned - in sequence.

I'm not well versed in these environment variables. Are these common variable names used for these purposes etc? I think to approve this PR I need to learn more, and a key part would be to learn about these environment variables common meaning in other projects or defined meaning in this project.

Some questions:
- Does AsyncHTTPClient respect some environment variable already to configure this?
- Is there some RFC document declaring that these environment variable names are to be used for purposes like this, or is it up to anyone to invent names etc?

Thinking about this PR on a high level, allowing configuration of this makes sense, but defaulting to using environment variables that I don't know the background about etc could change the default behavior for users etc and requires more scrutiny.

oauthenticator/generic.py
nngo: @GeorgianaElena I did scan over similar tornado issues/posts and they claim they do provide the option to override the CA certs and put the onus on the client code to handle setting that properly, hence this PR.

Maybe there is a better place to do this in the oauth2, but I couldn't find it and came up with this solution to allow me to override ca-certs to get this working in my scenario.

I have implemented it to first use a new `OAUTH2_CA_CERTS` env (higher precedence) with fallback to `SSL_CERT_FILE` env (b/c that env is what is used by the python ssl module). I don't think `SSL_CERT_FILE`  can be a directory, if you want I can remove this fallback, in the end of both of those environment variables are not set, then it is the save behavior as originally.

oauthenticator/generic.py
nngo: > `or None to use defaults. Default: None",` is a bit confusing. I think you need to be explicit that "to use defaults" refer to fallback on the two separate environment variables you have mentioned - in sequence.
> 
I modeled this help from the existing overrideable config via environment variable names, see lines 69-85, i.e. OAUTH2_USERDATA_REQUEST_TYPE, OAUTH2_TLS_VERIFY, OAUTH2_BASIC_AUTH, none of which calls out the environment variable names in the help. But I can adjust it to include those if you like and reword this as I do agree, that does sound a bit confusing.

manics: Do we need to support environment variables for this config property? Some of the `OAUTH2_*` env vars are due to the previous implementation of OAuthenticator where parameters couldn't be passed between components so are kept for backwards compatibility. Unless it's required I'm in favour of keeping things simple and not looking at env-vars for new properties.

nngo: @consideRatio I just pushed update to help docs for new `ca_certs` config, please take a look if that is better.

@manics is there another way I can override the oauthenticate oauth2 ca-certs when I am running this from jupyterhub/k8s-hub image deployed to K8s via a helm chart? and I don't want to have to build another image from the k8s-hub image to inject my ca-certs into the system trusted CAs, e.g. I just want to override it via environment variables.

manics: Couldn't you set the property in the same way you set any other one, the Z2JH equivalent of `c.GenericOAuthenticator.ca_cert = ...`? Why is an environment variable also needed?

manics:
There are some unresolved issues, and this repository had undergone some refactoring so this would need careful testing.

However, we may not need this PR anymore.
https://github.com/jupyterhub/oauthenticator/pull/578
adds support for all Tornado request parameters:
https://www.tornadoweb.org/en/stable/httpclient.html#request-objects

consideRatio:
@nngo thank you for working this!!

I'm hopeful that you can use custom certificates without configuring the authenticator specifically, but by configuring tornado running in JupyterHub via #614, and add your certificates to the jupyterhub pod via [`hub.extraFiles`](https://z2jh.jupyter.org/en/stable/resources/reference.html#hub-extrafiles) chart config.

It would be great to know if this did the trick for you!

PR #497, opened 2022-04-08
Use isort for import formatting

yuvipanda:
Follows up https://github.com/jupyterhub/jupyterhub/pull/3852

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,11 +21,12 @@ repos:
     hooks:
       - id: prettier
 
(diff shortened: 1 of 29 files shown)


yuvipanda:
image

YESSSSSSSSSS

consideRatio:
🎉🎉🎉 :) i'm glad we found something we both appreciate, i think having autoformatting to group imports into python lib / third party lib / local lib has been a very helpful distinction when i see things i dont recognize in imports overall - with this enforced, it is also very reliable to trust such grouping

Issue #492, opened 2022-03-19
Adding organization (and team) information to auth_state in GithubOAuthenticator


ft-cama:
### Proposed change
I would like to have the GithubOAuthenticator store the organization/team through which a user is authenticated in e.g. auth_state.

### What can we achieve with this information?
We can customize the kubespawner so that it based on the user's organization/team does something specific, e.g. mounts an organization specific volume.

### Why is this a good idea?
It seems intuitive and natural that a user's jupyterhub instance is customized based on which organization the user belongs to. 

### Code change proposal
Link to GithubOAuthenticator: [https://github.com/jupyterhub/oauthenticator/blob/main/oauthenticator/github.py]

When checking if the user belongs to one of the allowed organizations do the following (lines 176-182 in the link):
```
        user_org = '' # NEW
        if self.allowed_organizations:
            for org in self.allowed_organizations:
                user_in_org = await self._check_membership_allowed_organizations(
                    org, username, access_token
                )
                if user_in_org:
                    user_org = org # NEW
                    break
```

And then to later update auth_state (lines 195-197 in the link):
```
        auth_state['access_token'] = access_token
        auth_state['github_user'] = resp_json
        if user_org != '': # NEW
            auth_state['user_org'] = user_org # NEW
```

meeseeksmachine:
This issue has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/github-organizations-user-permissions/13382/2

ft-cama:
I get your point @manics!

How about instantiating an empty list and then append all the organizations that a user is part of? If the list is empty after looping through the allowed_organizations list, then the user is not authorized and authenticate still returns None.

```
    userdict = {"name": username}
    userdict["auth_state"] = auth_state = {}
    auth_state['access_token'] = access_token
    auth_state['github_user'] = resp_json

    if self.allowed_organizations:
        user_orgs = []
        for org in self.allowed_organizations:
            user_in_org = await self._check_membership_allowed_organizations(
                org, username, access_token
            )
            if user_in_org:
                user_orgs.append(org)
        if not user_orgs: # user_orgs is empty
            self.log.warning("User %s is not in allowed org list", username)
            return None
        else: # user is part of one or more orgs
            auth_state['user_orgs'] = user_orgs
```
[...]

PR #498, opened 2022-04-08
[GitHub] Add populate_teams_in_auth_state option

yuvipanda:
Hoping to get https://github.com/2i2c-org/infrastructure/issues/1146
done. Longer term, we might use some of the new group management
features that JupyterHub 2.x uses - but I think this is still
pretty helpful, and we're still on 1.x :D

Fixes https://github.com/jupyterhub/oauthenticator/issues/492

### DIFF
--- a/docs/source/github.md
+++ b/docs/source/github.md
@@ -36,7 +36,7 @@ We currently use the following fields:
 - `access_token` is the token used to authenticate to GitHub.
 - `teams` is list of teams the user is part of, fetched only if
   `fetch_teams` option is set to `True` on `GitHubOAuthenticator`.
-  The `read:orgs` scope is also required for this to work.
+  The `read:org` scope is also required for this to work.
 
 To use this expanded user information, you will need to subclass your
 current spawner and modify the subclass to read these fields from
--- a/docs/source/github.md
+++ b/docs/source/github.md
@@ -34,6 +34,9 @@ We currently use the following fields:
 - `name` is the full name GitHub knows the user by.
 - `email` is the publicly visible email address (if any) for the user.
 - `access_token` is the token used to authenticate to GitHub.
+- `teams` is list of teams the user is part of, fetched only if
+  `fetch_teams` option is set to `True` on `GitHubOAuthenticator`.
+  The `read:orgs` scope is also required for this to work.
 
 To use this expanded user information, you will need to subclass your
 current spawner and modify the subclass to read these fields from
--- a/docs/source/github.md
+++ b/docs/source/github.md
@@ -35,7 +35,7 @@ We currently use the following fields:
 - `email` is the publicly visible email address (if any) for the user.
 - `access_token` is the token used to authenticate to GitHub.
 - `teams` is list of teams the user is part of, fetched only if
-  `populate_team_membership` option is set to `True` on `GitHubOAuthenticator`.
+  `populate_teams_in_auth_state` option is set to `True` on `GitHubOAuthenticator`.
   The `read:org` scope is also required for this to work.
 
 To use this expanded user information, you will need to subclass your
--- a/docs/source/github.md
+++ b/docs/source/github.md
@@ -35,7 +35,7 @@ We currently use the following fields:
 - `email` is the publicly visible email address (if any) for the user.
 - `access_token` is the token used to authenticate to GitHub.
 - `teams` is list of teams the user is part of, fetched only if
-  `fetch_teams` option is set to `True` on `GitHubOAuthenticator`.
+  `populate_team_membership` option is set to `True` on `GitHubOAuthenticator`.
   The `read:org` scope is also required for this to work.
 
 To use this expanded user information, you will need to subclass your
--- a/oauthenticator/github.py
+++ b/oauthenticator/github.py
@@ -10,6 +10,7 @@ from tornado import web
 from tornado.httpclient import HTTPRequest
 from tornado.httputil import url_concat
 from traitlets import Bool, Set, Unicode, default, validate
+from requests.utils import parse_header_links
 
 from .oauth2 import OAuthenticator
 
@@ -242,25 +243,43 @@ class GitHubOAuthenticator(OAuthenticator):
         if self.fetch_teams:
             # Number of teams to request per page
             per_page = 100
-            # Page counter
-            cur_page = 1
 
             teams = []
+            url = self.github_api + f"/user/teams?per_page={per_page}"
             while True:
                 req = HTTPRequest(
-                    self.github_api
-                    + f"/user/teams?per_page={per_page}&page={cur_page}",
+                    url,
                     method="GET",
                     headers=_api_headers(access_token),
                     validate_cert=self.validate_server_cert,
                 )
-                resp_json = await self.fetch(req, "fetching user teams")
+                resp = await self.fetch(req, "fetching user teams", parse_json=False)
+
+                resp_json = json.loads(resp.body.decode())
                 teams += resp_json
 
-                # Only try to paginate if we received the max number of teams we could have
-                if len(resp_json) != per_page:
+                # Check if a Link header is present, with a collection of pagination links
+                links_header = resp.headers.get('Link')
+                if not links_header:
+                    # If Link header is not present, we just exit
                     break
-                cur_page += 1
+                else:
+                    # If Link header is present, let's parse it.
+                    links = parse_header_links(links_header)
+
+
+                    next_url = None
+                    # Look through all links to see if there is a 'next' link present
+                    for l in links:
+                        if l.get('rel') == 'next':
+                            next_url = l['url']
+
+                    # If we found a 'next' link, continue the while loop with the new URL
+                    # If not, we're out of pages to paginate, so we stop
+                    if next_url is not None:
+                        url = next_url
+                    else:
+                        break
 
             auth_state['teams'] = teams
 
--- a/oauthenticator/github.py
+++ b/oauthenticator/github.py
@@ -9,7 +9,7 @@ from jupyterhub.auth import LocalAuthenticator
 from tornado import web
 from tornado.httpclient import HTTPRequest
 from tornado.httputil import url_concat
-from traitlets import Set, Unicode, default, Bool, validate
+from traitlets import Bool, Set, Unicode, default, validate
 
 from .oauth2 import OAuthenticator
 
@@ -125,7 +125,7 @@ class GitHubOAuthenticator(OAuthenticator):
 
         Each user is currently limited to a maximum of 100 teams.
         """,
-        config=True
+        config=True,
     )
 
     @validate('scope')
@@ -136,7 +136,6 @@ class GitHubOAuthenticator(OAuthenticator):
         if 'read:org' not in proposal.value:
             return ['read:org'] + proposal.value
         return proposal.value
-
 
     async def authenticate(self, handler, data=None):
         """We set up auth_state based on additional GitHub info if we
@@ -249,7 +248,8 @@ class GitHubOAuthenticator(OAuthenticator):
             teams = []
             while True:
                 req = HTTPRequest(
-                    self.github_api + f"/user/teams?per_page={per_page}&page={cur_page}",
+                    self.github_api
+                    + f"/user/teams?per_page={per_page}&page={cur_page}",
                     method="GET",
                     headers=_api_headers(access_token),
                     validate_cert=self.validate_server_cert,
--- a/oauthenticator/github.py
+++ b/oauthenticator/github.py
@@ -242,6 +242,8 @@ class GitHubOAuthenticator(OAuthenticator):
                 per_page = 100
 
                 teams = []
+                # Fetch all teams across all orgs this user is in and we have visibility for
+                # https://docs.github.com/en/rest/reference/teams#list-teams-for-the-authenticated-user
                 url = self.github_api + f"/user/teams?per_page={per_page}"
                 while True:
                     req = HTTPRequest(
--- a/oauthenticator/github.py
+++ b/oauthenticator/github.py
@@ -117,12 +117,18 @@ class GitHubOAuthenticator(OAuthenticator):
     populate_teams_in_auth_state = Bool(
         False,
         help="""
-        If auth_state is enabled, populate it with the list of teams the current user is a member of.
+        Populates the authentication state dictionary `auth_state` with a key
+        `teams` assigned the list of teams the current user is a member of at
+        the time of authentication. The list of teams is structured like the
+        response of the GitHub API documented in
+        https://docs.github.com/en/rest/reference/teams#list-teams-for-the-authenticated-user.
 
-        'teams' will be a key in auth_state that has the same structure listed
-        in https://docs.github.com/en/rest/reference/teams#list-teams-for-the-authenticated-user.
-
-        Requires `read:org` to be set in scopes.
+        Requires `read:org` to be set in `scope`.
+        
+        Note that authentication state is only be available to a
+        `post_auth_hook` before being discarded unless configured to be
+        persisted via `enable_auth_state`. For more information, see
+        https://jupyterhub.readthedocs.io/en/stable/reference/authenticators.html#authentication-state.
         """,
         config=True,
     )
--- a/oauthenticator/github.py
+++ b/oauthenticator/github.py
@@ -6,11 +6,11 @@ import os
 import warnings
 
 from jupyterhub.auth import LocalAuthenticator
+from requests.utils import parse_header_links
 from tornado import web
 from tornado.httpclient import HTTPRequest
 from tornado.httputil import url_concat
-from traitlets import Bool, Set, Unicode, default, validate
-from requests.utils import parse_header_links
+from traitlets import Bool, Set, Unicode, default
 
 from .oauth2 import OAuthenticator
 
@@ -122,21 +122,12 @@ class GitHubOAuthenticator(OAuthenticator):
         'teams' will be a key in auth_state that has the same structure listed
         in https://docs.github.com/en/rest/reference/teams#list-teams-for-the-authenticated-user.
 
-        Automatically adds read:org scope to requested scopes if it is not present.
+        Requires `read:org` to be set in scopes.
 
         Each user is currently limited to a maximum of 100 teams.
         """,
         config=True,
     )
-
-    @validate('scope')
-    def _validate_scope(self, proposal):
-        """
-        Ensure read:org is requested if fetch_teams is true
-        """
-        if 'read:org' not in proposal.value and self.fetch_teams:
-            return ['read:org'] + proposal.value
-        return proposal.value
 
     async def authenticate(self, handler, data=None):
         """We set up auth_state based on additional GitHub info if we
@@ -241,47 +232,54 @@ class GitHubOAuthenticator(OAuthenticator):
                     break
 
         if self.fetch_teams:
-            # Number of teams to request per page
-            per_page = 100
-
-            teams = []
-            url = self.github_api + f"/user/teams?per_page={per_page}"
-            while True:
-                req = HTTPRequest(
-                    url,
-                    method="GET",
-                    headers=_api_headers(access_token),
-                    validate_cert=self.validate_server_cert,
+            if 'read:org' not in self.scope:
+                # This means the 'read:org' scope was not set, and we can't fetch teams
+                self.log.error(
+                    'read:org scope is required for fetch_teams functionality to work'
                 )
-                resp = await self.fetch(req, "fetching user teams", parse_json=False)
+            else:
+                # Number of teams to request per page
+                per_page = 100
 
-                resp_json = json.loads(resp.body.decode())
-                teams += resp_json
+                teams = []
+                url = self.github_api + f"/user/teams?per_page={per_page}"
+                while True:
+                    req = HTTPRequest(
+                        url,
+                        method="GET",
+                        headers=_api_headers(access_token),
+                        validate_cert=self.validate_server_cert,
+                    )
+                    resp = await self.fetch(
+                        req, "fetching user teams", parse_json=False
+                    )
 
-                # Check if a Link header is present, with a collection of pagination links
-                links_header = resp.headers.get('Link')
-                if not links_header:
-                    # If Link header is not present, we just exit
-                    break
-                else:
-                    # If Link header is present, let's parse it.
-                    links = parse_header_links(links_header)
+                    resp_json = json.loads(resp.body.decode())
+                    teams += resp_json
 
-
-                    next_url = None
-                    # Look through all links to see if there is a 'next' link present
-                    for l in links:
-                        if l.get('rel') == 'next':
-                            next_url = l['url']
-
-                    # If we found a 'next' link, continue the while loop with the new URL
-                    # If not, we're out of pages to paginate, so we stop
-                    if next_url is not None:
-                        url = next_url
-                    else:
+                    # Check if a Link header is present, with a collection of pagination links
+                    links_header = resp.headers.get('Link')
+                    if not links_header:
+                        # If Link header is not present, we just exit
                         break
+                    else:
+                        # If Link header is present, let's parse it.
+                        links = parse_header_links(links_header)
 
-            auth_state['teams'] = teams
+                        next_url = None
+                        # Look through all links to see if there is a 'next' link present
+                        for l in links:
+                            if l.get('rel') == 'next':
+                                next_url = l['url']
+
+                        # If we found a 'next' link, continue the while loop with the new URL
+                        # If not, we're out of pages to paginate, so we stop
+                        if next_url is not None:
+                            url = next_url
+                        else:
+                            break
+
+                auth_state['teams'] = teams
 
         return userdict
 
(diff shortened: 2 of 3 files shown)


oauthenticator/github.py
```diff
@@ -215,6 +239,31 @@ async def authenticate(self, handler, data=None):
...
+                    self.github_api
+                    + f"/user/teams?per_page={per_page}&page={cur_page}",
+                    method="GET",
+                    headers=_api_headers(access_token),
+                    validate_cert=self.validate_server_cert,
+                )
+                resp_json = await self.fetch(req, "fetching user teams")
+                teams += resp_json
+
+                # Only try to paginate if we received the max number of teams we could have
```
manics, line 260: Does the teams API response include pagination information?
https://docs.github.com/en/github-ae@latest/rest/overview/resources-in-the-rest-api#pagination

yuvipanda: I don't see it. If I set `per_page` to `1` to trigger this easy, here's first response for me:

```
[{'name': 'Owners', 'id': 66678, 'node_id': 'MDQ6VGVhbTY2Njc4', 'slug': 'owners', 'description': None, 'privacy': 'secret', 'url': 'https://api.github.com/organizations/867692/team/66678', 'html_url': 'https://github.com/orgs/publictransit-in/teams/owners', 'members_url': 'https://api.github.com/organizations/867692/team/66678/members{/member}', 'repositories_url': 'https://api.github.com/organizations/867692/team/66678/repos', 'permission': 'admin', 'created_at': '2007-10-20T11:24:19Z', 'updated_at': '2007-10-20T11:24:19Z', 'members_count': 4, 'repos_count': 7, 'organization': {'login': 'publictransit-in', 'id': 867692, 'node_id': 'MDEyOk9yZ2FuaXphdGlvbjg2NzY5Mg==', 'url': 'https://api.github.com/orgs/publictransit-in', 'repos_url': 'https://api.github.com/orgs/publictransit-in/repos', 'events_url': 'https://api.github.com/orgs/publictransit-in/events', 'hooks_url': 'https://api.github.com/orgs/publictransit-in/hooks', 'issues_url': 'https://api.github.com/orgs/publictransit-in/issues', 'members_url': 'https://api.github.com/orgs/publictransit-in/members{/member}', 'public_members_url': 'https://api.github.com/orgs/publictransit-in/public_members{/member}', 'avatar_url': 'https://avatars.githubusercontent.com/u/867692?v=4', 'description': 'Community projects around public transport in India', 'name': 'India public transport tools', 'company': None, 'blog': 'http://publictransit.in', 'location': 'India', 'email': 'arun.planemad@gmail.com', 'twitter_username': None, 'is_verified': False, 'has_organization_projects': True, 'has_repository_projects': True, 'public_repos': 8, 'public_gists': 0, 'followers': 0, 'following': 0, 'html_url': 'https://github.com/publictransit-in', 'created_at': '2011-06-22T17:54:15Z', 'updated_at': '2016-01-26T20:18:57Z', 'type': 'Organization'}, 'parent': None}]
```

Just a list without anything else.

manics: Anything in the response headers? The docs imply there should be a `Link:`

GeorgianaElena: I'm linking here the docs that mention the Link header as the location of pagination info: https://docs.github.com/en/github-ae@latest/rest/guides/traversing-with-pagination#basics-of-pagination

yuvipanda: Ah, I didn't inspect it but makes sense it would be in the response header.

I'm slightly torn on what to do here. I can look for `Link` in the header and try use that, but that means I can't use the nice `self.fetch` function. But that's probably the 'right' thing to do, especially given the specific warning about it in the GItHub API docs...

Should I move the code to stop using `self.fetch` and use a http client directly?

manics: It's more work, but code has a habit of being copied and pasted so I think it's better to do it properly. I suspect it'll also be useful in future if others extend this class to fetch additional information in their own deployments.

oauthenticator/github.py
```diff
@@ -113,6 +113,30 @@ def _github_client_secret_changed(self, name, old, new):
...
+    )
+
+    @validate('scope')
+    def _validate_scope(self, proposal):
+        """
+        Ensure read:org is requested if fetch_teams is true
+        """
+        if 'read:org' not in proposal.value:
+            return ['read:org'] + proposal.value
+        return proposal.value
```
manics, line 138: Automatically adding a scope is a change in behaviour, previously we've relied on administrators setting the scope themselves (e.g. to allow checking of private org memberships). On the other hand automatically setting it definitely makes this easier to configure and is a lot nicer. I'm undecided on what's best, though if we keep this as automatically set I think it'd be good to add something to the docs.

oauthenticator/github.py
yuvipanda: Makes sense.. I'll do that.

oauthenticator/github.py
consideRatio: Right now this logic is adding `read:org` scope independent on `fetch_teams` being configured or not which wouldn't make sense at least - right?

I've grown vary of convenience logic for admins that kind of need to understand what goes on anyhow, and how complicated it can become to document that, thinking that it overall leads to more complexity than benefits for admins using this software and us who maintains it.

I'm trying to make my vague hesitance towards automating this more explicit by considering a few situations a bit more explicitly.

1. The admin configuring this is aware of `read:org` will be appended because the admin has read the docs and understands this is implied by enabling `fetch_teams`.
   Then this is quite reasonable in my mind. But even then, after some time, even the admin who configured this initially and read the docs once could be confused about seeing this be requested by users after looking at docs observing it isn't explicitly declared in `scopes` or similar.
2. The admin configuring use of `fetch_teams` wasn't aware that `read:org` will be appended and got a functional installation directly without adjusting `scopes`.
   This admin may end up confused why users were asked for something more than explicitly declared in `scopes`. We may need to document this behavior also in `scopes` along with `fetch_teams` due to that.

Hmmm at this point I'm a more favorable towards documenting `fetch_teams` to be a feature that should be used with `read:org` scope, which should be manually configured alongside `fetch_teams`.

I'm even more favorable towards this now that I realize that this feature would also rely on `enable_auth_state` to be configured - right? See the JupyterHub documentation for authenticators: https://jupyterhub.readthedocs.io/en/stable/reference/authenticators.html#authentication-state. Z2JH provides a auth key by default, but it does not enable saving auth state by default.

oauthenticator/github.py
```diff
@@ -215,6 +239,31 @@ async def authenticate(self, handler, data=None):
...
 
+        if self.fetch_teams:
+            # Number of teams to request per page
+            per_page = 100
+            # Page counter
+            cur_page = 1
+
+            teams = []
+            while True:
+                req = HTTPRequest(
```
consideRatio, line 250:
```suggestion
                # GitHub API reference for /user/teams:
                # https://docs.github.com/en/rest/reference/teams#list-teams-for-the-authenticated-user
                req = HTTPRequest(
```

oauthenticator/github.py
manics: Good points! `read:org` potentially means the app has to be granted by the organisation owner in addition to the user. There's a couple of old issues somewhere in this repo where we tried to figure out exactly what GitHub organisation/user related scopes are required in different situations and it was very complicated.... and depended on the GitHub organisation and user's settings. Automatically adding one scope here to handle fetching teams solves one problem, but it's inconsistent with how the rest of this authenticator is configured.

cilogin handling things differently is a good point too, but I think it reflects how separately developed authenticators have been bundled together. It would definitely be nice to refactor them all to be consistent, or alternatively to decouple them into separate repos (we've already said no new authenticators will be accepted in this repo), but for now I think we'll have to accept they all use slightly different patterns.

yuvipanda: I feel fairly strongly that we shouldn't allow config that lets admins shoot themselves in the foot, and part of that is allowing config that has 0 chance of working under any circumstances. If enabling this feature without read:org works even in some capacity, then I think it can be optional. However, I don't think it is possible for this feature to work without read:org scope (I could be wrong here tho), so auto adding is the way to go as it prevents inconsistent config that doesn't actually work....

I understand that deciding which scopes are needed for what might be tricky. But if we enforce a simple rule - if a particular feature always needs a scope, we make sure that scope is requested when that feature is turned on.

However, I recognize that is inconsistent with how the authenticator works now. I also definitely don't want to get bogged down in this as part of this PR as it is quite ancillary. So I'll just remove this for now and document that that scope is needed.

yuvipanda: I had forgotten to only add it if `fetch_teams` is set. It's now done in 39b51b22e0e0cd7b6d6938415ed1a935728e2f9d. 

If I remove the code that automatically sets the correct scopes, and set `fetch_teams` to True, logging in produces this nasty error in the logs:

```
[E 2022-04-08 11:46:49.848 JupyterHub oauth2:380] Error fetching user teams 404 GET https://api.github.com/user/teams: {
     "documentation_url": "https://docs.github.com/rest/reference/teams#list-teams-for-the-authenticated-user",
     "message": "Not Found"
    }
[E 2022-04-08 11:46:49.850 JupyterHub web:1789] Uncaught exception GET /hub/oauth_callback?code=c1e50d7dbcbf5171fcf1&state=eyJzdGF0ZV9pZCI6ICJkOTNlNDdiYjQzZWU0N2I2YmYwZmEwOGU5OTI4MWZmMCIsICJuZXh0X3VybCI6ICIifQ%3D%3D (::ffff:127.0.0.1)
    HTTPServerRequest(protocol='http', host='localhost:8000', method='GET', uri='/hub/oauth_callback?code=c1e50d7dbcbf5171fcf1&state=eyJzdGF0ZV9pZCI6ICJkOTNlNDdiYjQzZWU0N2I2YmYwZmEwOGU5OTI4MWZmMCIsICJuZXh0X3VybCI6ICIifQ%3D%3D', version='HTTP/1.1', remote_ip='::ffff:127.0.0.1')
      File "/Users/yuvipanda/.local/share/virtualenvs/oauthenticator/lib/python3.9/site-packages/tornado/web.py", line 1704, in _execute
        result = await result
      File "/Users/yuvipanda/code/oauthenticator/oauthenticator/oauth2.py", line 222, in get
      File "/Users/yuvipanda/.local/share/virtualenvs/oauthenticator/lib/python3.9/site-packages/jupyterhub/handlers/base.py", line 804, in login_user
        authenticated = await self.authenticate(data)
      File "/Users/yuvipanda/.local/share/virtualenvs/oauthenticator/lib/python3.9/site-packages/jupyterhub/auth.py", line 473, in get_authenticated_user
        authenticated = await maybe_future(self.authenticate(handler, data))
      File "/Users/yuvipanda/code/oauthenticator/oauthenticator/github.py", line 249, in authenticate
        resp = await self.fetch(req, "fetching user teams", parse_json=False)
      File "/Users/yuvipanda/code/oauthenticator/oauthenticator/oauth2.py", line 381, in fetch
        raise e
      File "/Users/yuvipanda/code/oauthenticator/oauthenticator/oauth2.py", line 360, in fetch
        resp = await self.http_client.fetch(req, **kwargs)
    tornado.httpclient.HTTPClientError: HTTP 404: Not Found
```

This currently crashes the hub. I can probably handle this error (which we should do anyway) and provide a useful log message, but this reinforces to me that adding the scope if this feature is enabled is the right thing to do.

yuvipanda: I've removed the automatic scope setting in ac49a4c3fee761991464d71961b6bb03be25db03 and added docs.

yuvipanda: @manics reworked to use the Link header now!

oauthenticator/github.py
yuvipanda: @manics reworked to use the Link header now!

oauthenticator/github.py
consideRatio: The question is probably between:
1. hard failure on startup if misconfigured.
2. warning in config that it has been added + docs it should be added
3. no warning and no suggestion in the docs to add read:org to the scopes

I lean away from the last, and initially thought i appreciated one the most but ended up thinking the second as best.

I prefer 2 over 1 since this cant be enforced in a clean way as for example using a schema for a helm chart - validated even before impacting the software operation. To have an admin disrupt a running hub with a hard error... Nah, not good

With 2, we have a purpose to allowing it - to avoid having an admin crash a deployment, but also with the docs, to help admins declare config that make sense and conveys how something works better.

yuvipanda: Let's continue this conversation elsewhere, I'll try open up something more holistic rather than piecemeal. I don't want you to have to spend the limited review energy you have on this, @consideRatio! It's more worthy reviewing the other parts :)

yuvipanda: Currently this PR adds docs + emits a warning, which is consistent with the behavior we have for it right now.

oauthenticator/github.py
```diff
@@ -113,6 +114,21 @@ def _github_client_secret_changed(self, name, old, new):
...
+        False,
+        help="""
+        Fetch list of teams user is in and put it in auth_state.
+
+        'teams' will be a key in auth_state that has the same structure listed
+        in https://docs.github.com/en/rest/reference/teams#list-teams-for-the-authenticated-user.
+
+        Requires `read:org` to be set in scopes.
+
+        Each user is currently limited to a maximum of 100 teams.
```
consideRatio, line 127: This is no longer a limitation I think.

```suggestion
```

oauthenticator/github.py
```diff
@@ -113,6 +114,21 @@ def _github_client_secret_changed(self, name, old, new):
         config=True, help="Automatically allow members of selected organizations"
     )
 
+    fetch_teams = Bool(
+        False,
+        help="""
+        Fetch list of teams user is in and put it in auth_state.
+
+        'teams' will be a key in auth_state that has the same structure listed
+        in https://docs.github.com/en/rest/reference/teams#list-teams-for-the-authenticated-user.
```
consideRatio, line 123: Instead of saying it `will be a key in auth_state` I think it would be relevant to clarify that this is put in `auth_state` during authentication, making it available to a `Authenticator.post_auth_hook` but not for a Spawner hook of some kind, which would also require `enable_auth_state` to be set to True.

I'm mainly concerned that people assume it will be available, but then it isn't.

Related links:
- Docs on `Authenticator.post_auth_hook`: https://jupyterhub.readthedocs.io/en/stable/api/auth.html#jupyterhub.auth.Authenticator.post_auth_hook
- Discourse topic on using auth state etc: https://discourse.jupyter.org/t/tailoring-spawn-options-and-server-configuration-to-certain-users/8449

oauthenticator/github.py
```diff
@@ -113,6 +114,21 @@ def _github_client_secret_changed(self, name, old, new):
         config=True, help="Automatically allow members of selected organizations"
     )
 
+    fetch_teams = Bool(
```
consideRatio, line 117: `fetch_teams` is short and brief and I accept it, but I'd prefer something more verbose if you think it also could make sense @yuvipanda. For example: `add_teams_to_auth_state` or `fetch_teams_to_auth_state`.

I appreciate that `_to_auth_state` clarifies the end result, I appreciate that `fetch` implies notable work, I appreciate that `add_teams` more directly clarifies the result that `teams` will be in `auth_state` than `fetch_teams`.

Feel free to resolve without action if you are opinionated towards `fetch_teams`.

oauthenticator/github.py
```diff
@@ -215,8 +231,68 @@ async def authenticate(self, handler, data=None):
                     auth_state['github_user']['email'] = val['email']
                     break
 
+        if self.fetch_teams:
+            if 'read:org' not in self.scope:
+                # This means the 'read:org' scope was not set, and we can't fetch teams
+                self.log.error(
+                    'read:org scope is required for fetch_teams functionality to work'
+                )
```
consideRatio, line 239: Should it be an error or a warning? I was thinking a warning is more appropriate since a user will successfully be authenticated etc.

oauthenticator/github.py
```diff
@@ -215,8 +231,68 @@ async def authenticate(self, handler, data=None):
...
+            else:
+                # Number of teams to request per page
+                per_page = 100
+
+                # Fetch all teams across all orgs this user is in and we have visibility for
+                # https://docs.github.com/en/rest/reference/teams#list-teams-for-the-authenticated-user
```
consideRatio, line 245: This kind of clarity is something I think belongs to the docstring of the configuration to enable this. I had to ask myself what it meant to fetch teams as part of a user rather than an organization etc.

Also, since it notes "we have visibility for" I become hesitant on my understanding - wouldn't that be all teams if we have requested `read:org`? Or is the complexity like partial visibility because both the user part of the team and the github org needs to agree on this being made available?

docs/source/github.md
```diff
@@ -34,6 +34,9 @@ We currently use the following fields:
 - `name` is the full name GitHub knows the user by.
 - `email` is the publicly visible email address (if any) for the user.
 - `access_token` is the token used to authenticate to GitHub.
+- `teams` is list of teams the user is part of, fetched only if
+  `fetch_teams` option is set to `True` on `GitHubOAuthenticator`.
+  The `read:orgs` scope is also required for this to work.
```
consideRatio, line 39:
```suggestion
  The `read:org` scope is also required for this to work.
```

consideRatio: Can a user grant some but not all of these permissions, if so maybe we should have a fallback on just partially granted scopes by the user or handle an unauthorized response without crashing as we would do now.

consideRatio (commented):
Thank you for your work on this @yuvipanda!!!

oauthenticator/github.py
yuvipanda: I think this should be an error though, as the software is not functioning as it is configured to.

docs/source/github.md
yuvipanda: Can I do that in a follow-up?

oauthenticator/github.py
yuvipanda: > I'm mainly concerned that people assume it will be available, but then it isn't.

I've clarified in the string that this works only if auth_state is enabled. I'd like to not repeat more details about auth_state here though, users can look through docs about auth_state for that.

docs/source/github.md
consideRatio: Sure!

oauthenticator/github.py
consideRatio: It will work for the Authenticator class post_auth_hook though, it just wont be persisted to be around in auth state for later use by a spawner.

I fully agree on avoiding repetition, linking out to relevant socs to explain details is great!

yuvipanda: @consideRatio we don't actually have a 'topic' page about auth_state, so I'm not sure what to link out to from here. I'm also running out of energy unfortunately, and want to avoid shaving too many yaks! So instead of trying to figure out best way to tell people what auth_state is here, can I just let things be as they are, and just make sure that people know this only puts things into auth_state if it is enabled?

oauthenticator/github.py
yuvipanda: I tried to find docs on GitHub's API what the 'Grant' button that shows up sometimes when i log in for some organizations, and I don't actually know why exactly. Given this is the exact situation current feature wrt allowing only users with specific teams to log in, I'm just going to remove the ambiguous comment and let that be.

oauthenticator/github.py
yuvipanda: `fetch_teams` is definitely too vague - thanks for catching that! I've renamed it to `populate_team_membership`, which I think is definitely an improvement as it tells you what it does after fetching. However, I'd like to keep `auth_state` out of it as it feels a bit too verbose to me. If you feel strongly about it I can try find a way to work it in!

yuvipanda: Somehow this particular name has bothered me, and I kept trying to figure out why. While I still can not articulate exactly why, it's given me an idea on how to actually redo this PR in a way that's more generic, where things are easier to name! I'll work on it, and I've converted this PR to draft in the meantime.

yuvipanda: @consideRatio ok i clearly couldn't actually redo this PR. Is calling it `populate_team_membership` acceptable to you? If so, this is ready for merge.

yuvipanda:
Well I clearly couldn't rework the PR to be more general in time :D I think this is ready for merge.

oauthenticator/github.py
```diff
@@ -113,6 +114,19 @@ def _github_client_secret_changed(self, name, old, new):
         config=True, help="Automatically allow members of selected organizations"
     )
 
+    populate_teams_in_auth_state = Bool(
+        False,
+        help="""
+        If auth_state is enabled, add list of teams user is part of to it.
```
GeorgianaElena, line 120: *It seems it was partially re-worded?*

```suggestion
        If auth_state is enabled, add list of team users to it.
```

oauthenticator/github.py
```diff
@@ -215,8 +229,67 @@ async def authenticate(self, handler, data=None):
...
+                break
+            else:
+                # If Link header is present, let's parse it.
+                links = parse_header_links(links_header)
+
+                next_url = None
+                # Look through all links to see if there is a 'next' link present
+                for l in links:
+                    if l.get('rel') == 'next':
+                        next_url = l['url']
```
GeorgianaElena, line 283:
```suggestion
                    if l.get('rel') == 'next':
                        next_url = l['url']
                        break
```

oauthenticator/github.py
```diff
@@ -215,8 +229,67 @@ async def authenticate(self, handler, data=None):
...
+
+            resp_json = json.loads(resp.body.decode())
+            content += resp_json
+
+            # Check if a Link header is present, with a collection of pagination links
+            links_header = resp.headers.get('Link')
+            if not links_header:
+                # If Link header is not present, we just exit
+                break
+            else:
```
GeorgianaElena, line 275: I believe we can get rid of the `else` statement here, right?

GeorgianaElena (approved):
I left a few minor style-related comments, but I believe that we should merge this PR 🚀

oauthenticator/github.py
yuvipanda: Done!

oauthenticator/github.py
yuvipanda: Done!

oauthenticator/github.py
yuvipanda: I reworded for clarity

oauthenticator/github.py
```diff
@@ -113,6 +114,19 @@ def _github_client_secret_changed(self, name, old, new):
...
 
+    populate_teams_in_auth_state = Bool(
+        False,
+        help="""
+        If auth_state is enabled, populate it with the list of teams the current user is a member of.
+
+        'teams' will be a key in auth_state that has the same structure listed
+        in https://docs.github.com/en/rest/reference/teams#list-teams-for-the-authenticated-user.
+
+        Requires `read:org` to be set in scopes.
```
consideRatio, line 125: I wanted to clarify that we populate authentication state no matter if `enable_auth_state` is configured. This state, while not persisted without `enable_auth_state`, can still be relevant to use by a `post_auth_hook`, so I don't think we should add a condition about only doing this if `enable_auth_state` is set either.

Yet, since the authentication state is perhaps mostly relevant for other purposes outside of the `post_auth_hook`, it would be good to link out to authentication state documentation where `enable_auth_state` is described in more detail.

```suggestion
        Populates the authentication state dictionary `auth_state` with a key
        `teams` assigned the list of teams the current user is a member of at
        the time of authentication. The teams have the same structure listed in
        https://docs.github.com/en/rest/reference/teams#list-teams-for-the-authenticated-user.

        Requires `read:org` to be set in `scopes`.
        
        Note that authentication state is only be available to a
        `post_auth_hook` before being discarded unless configured to be
        persisted via `enable_auth_state`. For more information, see
        https://jupyterhub.readthedocs.io/en/stable/reference/authenticators.html#authentication-state.
```

oauthenticator/github.py
consideRatio: Thank you Yuvi for your work to think about this! I've opened https://github.com/jupyterhub/oauthenticator/pull/498/files#r860745304 to make me feel happy about this documentation detail!

oauthenticator/github.py
```diff
@@ -215,8 +229,68 @@ async def authenticate(self, handler, data=None):
...
+        while True:
+            req = HTTPRequest(
+                url,
+                method="GET",
+                headers=_api_headers(access_token),
+                validate_cert=self.validate_server_cert,
+            )
+            resp = await self.fetch(req, "fetching user teams", parse_json=False)
+
+            resp_json = json.loads(resp.body.decode())
```
consideRatio, line 267: In case GitHub is down or we get a slow-down 429 response or similar it would be good to error with a message about status code instead of failing to parse the json.

```suggestion
            resp.raise_for_status()
            resp_json = json.loads(resp.body.decode())
```

consideRatio (commented):
Thanks for your work on this Yuvi! I've made two suggestions but otherwise thinks this LGTM and is ready for merge!

yuvipanda:
Thank you so much for these suggestions, @consideRatio! I've applied them :)

consideRatio:
Wieeee nice work on this @yuvipanda and thanks for your review work @manics and @GeorgianaElena!!

PR #502, opened 2022-04-25
Fix AUTH0_SUBDOMAIN default setting

alejandrosame:
The decorated function `_auth0_subdomain_default` always sets the subdomain value to an empty value due to a missing return. This PR fixes that issue.


yuvipanda:
Thanks for the catch and contribution, @alejandrosame!

PR #503, opened 2022-05-02
[CILogon] Use `get_callback_url` to determine the OAuth callback URL

brianaydemir:
When `oauth_callback_url` is not configured, this function will fallback on guessing based on the request URL, which is desirable when the JupyterHub instance is serving multiple domains.


GeorgianaElena (approved):
Thanks @brianaydemir! 🚀

PR #504, opened 2022-05-16
[GitHub] fix implementation of populate_teams_in_auth_state

consideRatio:
When using tornado's HTTPClient, there will be an error thrown by default as documented in [tornado.httpclient.HTTPClient](https://www.tornadoweb.org/en/stable/httpclient.html#tornado.httpclient.HTTPClient).

---

Fixes broken implementation of `GitHubOAuthenticator.populate_teams_in_auth_state` in #498 as caused by my code suggestion adding `resp.raise_for_status()`.


yuvipanda:
Thanks, @consideRatio!

PR #505, opened 2022-05-19
maint: remove deprecated distutils reference

consideRatio:
This is similar to what was done in https://github.com/jupyterhub/jupyterhub/pull/3562, but instead of declaring a dependency on `packaging` as is done in JupyterHub for this one-liner, I went for some basic string manipulation. I think this should be fine (robust) as we are not considering some patch or pre-release version etc, just checking if the major version is larger than 2.

### DIFF
--- a/oauthenticator/azuread.py
+++ b/oauthenticator/azuread.py
@@ -3,7 +3,6 @@ Custom Authenticator to use Azure AD with JupyterHub
 """
 import os
 import urllib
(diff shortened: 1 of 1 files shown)


consideRatio:
Actually, I'll propose another PR instead. I want to stop supporting PyJWT 1 and 2 now as that added complexity that I don't think we still need.

PR #506, opened 2022-05-19
breaking maint: assume pyjwt>=2 to reduce complexity

consideRatio:


GeorgianaElena (commented):
@consideRatio, wondering if we should keep this PR in draft mode until a `mwoauth` release or find another way to signal this dependency?

GeorgianaElena (commented):
Yuhuuuu 🎉  new version of `mwoauth` was released, right @consideRatio? Thanks for helping make that happen

GeorgianaElena (approved):
Many, many thanks @halfak ✨ 

 LGTM 🚀

yuvipanda:
@halfak is a name I haven't heard in such a long time! So good to hear it again :D

Issue #207, opened 2018-08-23
[Auth0] Support for using custom domains


isaurssaurav:
I have recently bought a custom domain for Auth0 login's Url.(a.example.com)
But there is no way to pass my new domain Url in jupyterhub's Auth0.
I have tired in with :
`AUTH0_SUBDOMAIN = a.example.com`
But, It adds .auth0.com at last . For e.g a.example.com.auth0.com

Thanks.

isaurssaurav:
i have solved it by, overriding **Auth0LoginHandler** class.
But it would be nice if we have something like `AUTH0_DOMAIN = a.example.com`

antoinekrajnc:
Hi @isaurssaurav, 


Thanks a lot for your answer. I ran into the same issue. Could you provide a little more detail? Did you override the *Auth0LoginHandler* class inside the `.yaml` config file? 

Thanks a lot for your help!

isaurssaurav:
@antoinekrajnc, 
I override the function from jupyerhub_config.py.

So. [This](https://github.com/jupyterhub/oauthenticator/blob/44121eab53ec1a6f6c957b64f438f3cd65377845/oauthenticator/auth0.py) is the class you will be overriding

It's been long, so I don't really remember the details  but here is my jupterhub_config file:

```
from oauthenticator.auth0 import Auth0OAuthenticator
from oauthenticator.auth0 import Auth0LoginHandler

c.JupyterHub.authenticator_class = CustomAuthenticator //this is where you override the function

#Custom Authenticator
class CustomAuth0LoginHandler(Auth0LoginHandler):
    AUTH0_SUBDOMAIN = os.environ['AUTH0_SUBDOMAIN']
    _OAUTH_AUTHORIZE_URL = "https://%s/authorize" % AUTH0_SUBDOMAIN
    _OAUTH_ACCESS_TOKEN_URL = "https://%s/oauth/token" % AUTH0_SUBDOMAIN


class CustomAuthenticator(Auth0OAuthenticator):
    login_handler = CustomAuth0LoginHandler

    @gen.coroutine
    def authenticate(self, handler, data=None):
        AUTH0_SUBDOMAIN = os.environ['AUTH0_SUBDOMAIN']
        code = handler.get_argument("code")
        http_client = AsyncHTTPClient()
        params = {
            'grant_type': 'authorization_code',
            'client_id': self.client_id,
            'client_secret': self.client_secret,
            'code': code,
[...]

PR #510, opened 2022-05-20
[Auth0] Allow use of custom domains

biomath-vlad:
Addresses this issue #207

### DIFF
--- a/oauthenticator/auth0.py
+++ b/oauthenticator/auth0.py
@@ -6,6 +6,7 @@ Derived using the Github and Google OAuthenticator implementations as examples.
 The following environment variables may be used for configuration:
 
     AUTH0_SUBDOMAIN - The subdomain for your Auth0 account
+    AUTH0_CUSTOM_DOMAIN - The custom domain for your Auth0 account
     OAUTH_CLIENT_ID - Your client id
     OAUTH_CLIENT_SECRET - Your client secret
     OAUTH_CALLBACK_URL - Your callback handler URL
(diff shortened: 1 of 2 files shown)


oauthenticator/auth0.py
```diff
@@ -42,35 +43,64 @@ class Auth0OAuthenticator(OAuthenticator):
 
     login_service = "Auth0"
 
-    auth0_subdomain = Unicode(config=True)
+    auth0_subdomain = Unicode(config=True, allow_none=True)
+    auth0_custom_domain = Unicode(config=True, allow_none=True)
```
consideRatio, line 47: This PR currently doesn't have an explanation or description to help me with little experience with specifically Auth0. Due to that, reviewing the PR becomes hard.

I see this configuration doesn't include a `help` keyword, but it really should. Also the `auth0_subdomain` should but didn't even though it wasn't added in this PR.

Could you add `help` sections to these configuration options?

consideRatio: Looking at the linked issue, I now understand a bit more.

I agree with @GeorgianaElena's comment about deprecating subdomain here: https://github.com/jupyterhub/oauthenticator/issues/207#issuecomment-1076402417.

consideRatio:
## Attempt on providing a background for the PR

Auth0 has documentation on [how to configure "custom domains"](https://auth0.com/docs/customize/custom-domains). What this means is not yet clear to me, but my understanding is that when you use Auth0 as an identity provider, and you have done something related to "custom domains" in the Auth0 documentation above, then this JupyterHub authenticator class needs to adjust in some way?

oauthenticator/auth0.py
```diff
@@ -6,6 +6,7 @@
 The following environment variables may be used for configuration:
 
     AUTH0_SUBDOMAIN - The subdomain for your Auth0 account
+    AUTH0_CUSTOM_DOMAIN - The custom domain for your Auth0 account
```
drhagen, line 9: The original design in #207 had this called `AUTH0_DOMAIN`, which I agree is better. That this can be a custom domain is more of a marketing term. It is not really "custom" from the perspective of `oauthenticator`. It is just a regular domain name. You could put `mycompany.auth0.com` in this variable and it would work perfectly, even though it is not a custom domain by Auth0's definition.

drhagen:
This can be closed as superseeded by the merged #534.

PR #511, opened 2022-05-20
docs/ci: use myst, fix broken links, add linkcheck test, remove deprecated distutils, avoid 2x job triggers

### LINKED ISSUE #480: Drop dependency on recommonmark - use MyST instead?
We have a dependency on recommonmark for our docs, but that makes us not able to work with Python 3.9 because its dependency commonmark hasn't updated to support Python 3.9.

I'd love to transition to have MyST instead. I think perhaps we use recommonmark to have markdown files part of Sphinx, while we could transition to use MyST instead.

consideRatio:
The docs had not been updated for a long time, so this PR is a significant overhaul. There were plenty of broken links etc.

Closes #480.

There is one thing that doesn't fit so well in this PR, which is [72ce042](https://github.com/jupyterhub/oauthenticator/pull/511/commits/72ce042c0812dc1be0755aaec36761e3d6cbe15a) that was added because I could not get `make html` or `make linkcheck` to function properly without making some changes there. Closes #505 which was a dedicated PR trying to do something about that as well.


docs/source/getting-started.rst
```diff
@@ -287,6 +283,13 @@ You can customize the sign in button text (optional):
 
    c.GoogleOAuthenticator.login_service = 'My College'
 
+.. note::
+
+   Additional notes, that seem quite outdated at the time of writing May 2022,
+   are available about authorizing users part of specific Google Groups are
+   :ref:`available here <google-groups-label>`. Contributions to update these
+   and re-verify this functionality are most welcome.
+
```
GeorgianaElena, line 292: Oh yes!  👀

.github/workflows/test.yml
```diff
@@ -5,7 +5,19 @@ name: Tests
 
 on:
   pull_request:
+    paths-ignore:
+      - "docs/**"
+      - ".github/workflows/*.yaml"
+      - "!.github/workflows/publish.yaml"
```
GeorgianaElena, line 11: Any reason not to ignore the publish workflow here?

```suggestion
```

.github/workflows/test.yml
```diff
@@ -5,7 +5,19 @@ name: Tests
...
   push:
+    paths-ignore:
+      - "docs/**"
+      - ".github/workflows/*.yaml"
+      - "!.github/workflows/publish.yaml"
```
GeorgianaElena, line 16: Same

```suggestion
```

.github/workflows/test.yml
consideRatio: Nice catch! It was copy pasted from a workflow named publish, but I had forgo to update it to de-ignore the `test.yaml` workflow. The idea was that "ignore all workflows (line before), but not this workflow"

consideRatio:
@GeorgianaElena thank you for the review!!! I applied your suggestions, I'll go for a merge right now then!

PR #513, opened 2022-06-02
Add changelog for 15.0.0

consideRatio:
There is a linkcheck failure, but I think that resolves when we publish the security advisory after we make this release.

__Note that this PR includes some non-generated text under "Breaking changes" that I think merits review.__


yuvipanda (approved):
I added a separate 'breaking security change' as a separate item, but otherwise LGTM!

yuvipanda:
@consideRatio my suggestion is to hold off until Monday, as it contains breaking changes. We can do a release and publicize it then, rather than dropping it on a friday - especially as it contains security fixes.

Issue #514, opened 2022-06-08
[CILogon] Allow hub usernames to be left unchanged


GeorgianaElena:
### Bug description

With `oauthenticator` 15.0.0, there isn't possible anymore to have the hub usernames be the value returned by the chosen claim. This is because `action` is a **required** field of `allowed_idps` dict and it can either be `strip_idp_domain` or `prefix`.

#### Expected behaviour

Migrate to 15.0.0 without requiring to change existing hub user's usernames.

#### Actual behaviour

You can either strip the domain from the username or prefix the username claim with a string. 

### Proposed solution

We could either remove the `action` from the required fields, which will translate into "no action specified on the username, then no change will be applied to it", or we could add a new explicit action that's just "don't change".

https://github.com/jupyterhub/oauthenticator/blob/42bebe989ec10e8a439bc80a011fbe64cc01669f/oauthenticator/schemas/cilogon-schema.yaml#L23-L25

consideRatio:
It would be great to get this resolved quickly, it sounds like a very troublesome experience otherwise. If you submit a fix I'll be quick to assist with review so we can get a new release out.

Issue #515, opened 2022-06-08
[CILogon] Allow specifying which domains should be allowed to login using a specific identity provider


GeorgianaElena:
### Proposed change

With latest oauthenticator version we lost the option to allow only certain domains to login into a hub. Currently we are allowing identity providers. I believe it would be useful to combine these to in order to be able to specify which domains to allow to login using a specific identity provider.

For example, if we'd like to allow login into a hub to only `2i2c.org` accounts using CILogonOAuthenticator, then `allowed_idps` isn't enough, since 2i2c doesn't have their own SSO, so we'll need to login though Google. So, to only allow `2i2c.org`, we'll also need to define an  `username_pattern` in addition to `allowed_idps`.

### (Optional): Suggest a solution

Add a new key under a specific identity provider in `allowed_idps` that would allow only certain domains to login though a specific identity provider.

PR #516, opened 2022-06-08
Don't make action a required field of CILogonOAuthenticator.allowed_idps

GeorgianaElena:
Fixes https://github.com/jupyterhub/oauthenticator/issues/514

### DIFF
--- a/oauthenticator/schemas/cilogon-schema.yaml
+++ b/oauthenticator/schemas/cilogon-schema.yaml
@@ -22,7 +22,6 @@ properties:
         type: string
     required:
       - username_claim
-      - action
     allOf:
       - if:
           properties:


minrk (approved):

GeorgianaElena:
@consideRatio, I was just working on writing a test for https://github.com/jupyterhub/oauthenticator/issues/515 (hoping to have that feature in the new release) and got an error from the schema validator from not adding the `action`. So this might be broken. Failing error:

```
jsonschema.exceptions.ValidationError: 'domain' is a required property
E           
E           Failed validating 'required' in schema['properties']['username_derivation']['allOf'][0]['then']:
E               {'required': ['domain']}
E           
E           On instance['username_derivation']:
E               {'username_claim': 'email'}
```

I'm working on fixing this and really hope to cut a release tonight after I get back to the computer or as early as possible tomorrow :(

- [ ] I would like to add some better examples to the docs too
- [ ] and tests
- [ ] and also document a possible issue with CILogon using an old OIDC endpoint for google (and open an issue upstream about this) (Old endpoint is http://google.com/accounts/o8/id, and new endpoint should be https://accounts.google.com/o/oauth2/v2/auth)

I added some comments in https://github.com/2i2c-org/infrastructure/pull/1403  about some workarounds I did to fix various issues if you want to check it out.

Also, sorry for pouring my todo list here 😅

PR #517, opened 2022-06-09
[CILogon] Don't make action a required field of CILogonOAuthenticator.allowed_idps follow-up

GeorgianaElena:
Follow-up to https://github.com/jupyterhub/oauthenticator/pull/516. Without this, it's still impossible to don't specify an `action`.

This also adds a test and documents this option.

Ref: https://github.com/jupyterhub/oauthenticator/issues/514

### DIFF
--- a/docs/source/migrations.md
+++ b/docs/source/migrations.md
@@ -2,17 +2,17 @@
 
 The following section describes how to migrate your OAuthenticator to a newer version given some upgrade scenarios.
 
-## Migrating CILogonOAuthenticator to version 15.0.0
+## Migrating CILogonOAuthenticator to version 15.0.1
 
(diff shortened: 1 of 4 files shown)


consideRatio (commented):
This will be 15.0.1 i figure, but right now the migration docs mentions only 15.0.0.

I suggest replacing 15.0.0 with 15.0.1 without further notes pretty much.

GeorgianaElena:
Thanks a lot @consideRatio! I believe I have addressed your comments.

After the merge of this PR, I believe we should go for a 15.0.1 release. I also opened https://github.com/jupyterhub/oauthenticator/pull/518, but maybe that needs more discussion and should be released later? What do you think?

consideRatio:
@GeorgianaElena yepp unless its another bugfix, let's not include it before 15.0.1 is released.

Want to go for a release? I merged a bigbucket bugfix pr just now as well btw.

GeorgianaElena:
> Want to go for a release? I merged a bigbucket bugfix pr just now as well btw.

Yep! Let's see how it goes. I'll open up a PR with the updated changelog shortly

consideRatio:
I agree!

PR #526, opened 2022-07-07
Refactor oauthenticators

GeorgianaElena:
- Creates a base `authenticate` method that consists of:
 ```python
    async def authenticate(self, handler, data=None):
        tokens_info = await self.get_tokens_info(handler)
        user_info = await self.token_to_user(tokens_info)
        auth_state = await self.create_auth_state(tokens_info, user_info)

        return auth_state
```
  The scope is to also introduce `refresh_user` in this workflow
- Implements genric `get_tokens_info` and `token_to_user` methods in `oauth2.py` and lets the oauthenticators only implement `create_auth_state` (this ca be further refactored into smaller pieces)
- Moves some attributes from `generic` to `oauth2`

The puropose of this refactoring is to make implementing https://github.com/jupyterhub/oauthenticator/issues/398 easier.

- [X] oauthenticator.oauth2
- [X] oauthenticator.auth0
- [X] oauthenticator.azuread
- [X] oauthenticator.bitbucket
- [X] oauthenticator.cilogon
- [X] oauthenticator.generic
- [X] oauthenticator.github
- [X] oauthenticator.gitlab
- [x] oauthenticator.globus
- [x] oauthenticator.google
- [x] oauthenticator.okpy
- [x] oauthenticator.openshift
- [x] oauthenticator.mediawiki

### DIFF
--- a/oauthenticator/auth0.py
+++ b/oauthenticator/auth0.py
@@ -33,17 +33,20 @@ jupyterhub_config.py :
   c.JupyterHub.authenticator_class = 'oauthenticator.auth0.Auth0OAuthenticator'
 
 """
-import json
 import os
 
 from jupyterhub.auth import LocalAuthenticator
-from tornado.httpclient import HTTPRequest
 from traitlets import Unicode, default
 
 from .oauth2 import OAuthenticator
 
 
 class Auth0OAuthenticator(OAuthenticator):
+
+    _deprecated_oauth_aliases = {
+        "username_key": ("username_claim", "15.1.0"),
+        **OAuthenticator._deprecated_oauth_aliases,
+    }
 
     login_service = "Auth0"
 
@@ -68,10 +71,17 @@ class Auth0OAuthenticator(OAuthenticator):
         )
 
     username_key = Unicode(
-        os.environ.get("OAUTH2_USERNAME_KEY", "email"),
         config=True,
-        help="Userdata username key from returned json with user data login information",
+        help="Deprecated, use `Auth0OAuthenticator.username_claim`",
     )
+
+    @default("user_auth_state_key")
+    def _user_auth_state_key_default(self):
+        return "auth0_user"
+
+    @default("username_claim")
+    def _username_claim_default(self):
+        return "email"
 
     @default("logout_redirect_url")
     def _logout_redirect_url_default(self):
@@ -85,61 +95,9 @@ class Auth0OAuthenticator(OAuthenticator):
     def _token_url_default(self):
         return "https://%s/oauth/token" % self.auth0_domain
 
-    async def authenticate(self, handler, data=None):
-        code = handler.get_argument("code")
-
-        params = {
-            'grant_type': 'authorization_code',
-            'client_id': self.client_id,
-            'client_secret': self.client_secret,
-            'code': code,
-            'redirect_uri': self.get_callback_url(handler),
-        }
-        url = self.token_url
-
-        req = HTTPRequest(
-            url,
-            method="POST",
-            headers={"Content-Type": "application/json"},
-            body=json.dumps(params),
-        )
-
-        resp_json = await self.fetch(req)
-
-        access_token = resp_json['access_token']
-
-        refresh_token = resp_json.get('refresh_token')
-        id_token = resp_json.get('id_token')
-
-        # Determine who the logged in user is
-        headers = {
-            "Accept": "application/json",
-            "User-Agent": "JupyterHub",
-            "Authorization": "Bearer {}".format(access_token),
-        }
-        req = HTTPRequest(
-            "https://%s/userinfo" % self.auth0_domain,
-            method="GET",
-            headers=headers,
-        )
-        resp_json = await self.fetch(req)
-
-        name = resp_json.get(self.username_key)
-        if not name:
-            self.log.error(
-                "Auth0 user contains no key %s: %s", self.username_key, resp_json
-            )
-            return
-
-        return {
-            'name': name,
-            'auth_state': {
-                'access_token': access_token,
-                'refresh_token': refresh_token,
-                'id_token': id_token,
-                'auth0_user': resp_json,
-            },
-        }
+    @default("userdata_url")
+    def _userdata_url_default(self):
+        return "https://%s/userinfo" % self.auth0_domain
 
 
 class LocalAuth0OAuthenticator(LocalAuthenticator, Auth0OAuthenticator):
--- a/oauthenticator/auth0.py
+++ b/oauthenticator/auth0.py
@@ -44,7 +44,7 @@ from .oauth2 import OAuthenticator
 class Auth0OAuthenticator(OAuthenticator):
 
     _deprecated_oauth_aliases = {
-        "username_key": ("username_claim", "15.1.0"),
+        "username_key": ("username_claim", "16.0.0"),
         **OAuthenticator._deprecated_oauth_aliases,
     }
 
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -22,10 +22,7 @@ from traitlets import Any, Bool, Dict, List, Unicode, default
 
 
 def guess_callback_uri(protocol, host, hub_server_url):
-    return "{proto}://{host}{path}".format(
-        proto=protocol, host=host, path=url_path_join(hub_server_url, "oauth_callback")
-    )
-
+    return f'{protocol}://{host}{url_path_join(hub_server_url, "oauth_callback")}'
 
 STATE_COOKIE_NAME = "oauthenticator-state"
 
@@ -43,12 +40,12 @@ def _deserialize_state(b64_state):
     try:
         json_state = base64.urlsafe_b64decode(b64_state).decode("utf8")
     except ValueError:
-        app_log.error("Failed to b64-decode state: %r", b64_state)
+        app_log.error(f"Failed to b64-decode state: {b64_state}")
         return {}
     try:
         return json.loads(json_state)
     except ValueError:
-        app_log.error("Failed to json-decode state: %r", json_state)
+        app_log.error(f"Failed to json-decode state: {json_state}")
         return {}
 
 
@@ -90,7 +87,7 @@ class OAuthLoginHandler(OAuth2Mixin, BaseHandler):
             ).geturl()
             if next_url != original_next_url:
                 self.log.warning(
-                    "Ignoring next_url %r, using %r", original_next_url, next_url
+                    f"Ignoring next_url {original_next_url}, using {next_url}"
                 )
         if self._state is None:
             self._state = _serialize_state(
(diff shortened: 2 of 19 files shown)

### CODE REVIEW

**Comment by minrk** [comment]
Hooray! I've wanted to do this for ages.

**Comment by GeorgianaElena** [comment]
I believe I got myself into a refactoring rabbit hole so I'll stop now and try to provide an update about the state of things so far to help with reviews 😥

### New `authenticate` workflow
I believe I managed to synthesize the `authentication` workflow into a set of methods that can be independently tweaked (or not) by the `oauthenticators` that need some of them done differently. So, here it goes:

1. build the parameters to be used in the request exchanging the oauth code for the access token
2. exchange the oauth code for an access token and get the JSON with info about it
3. use the access_token to get userdata info
4. extract the username out of the user_info dict
5. check if there any refresh_token in the tokens_info dict
6. if not, try to retrieve it from a previous auth_state if available
7. build the auth model to be persisted if authentication goes right
8. check if the username that's authenticating should be authorized
9. update the auth model with any info if available
10. return the auth model

### What changed
(I'm sure I'm going to miss some :/)
- some configs were deprecated in order to be consistent
- new config options were added
- some configs moved from being specific to only one authenticator to being part of the base class
- authenticate now returns more info to satisfy all oauthenticators

### What still needs to be improved

- there's an entire entanglement of methods either raising errors, or returning nulls on failure that should be sorted out
- not enough comments
- not enough corner case checks
- need to sort out making autodoc generate docs for methods too

### What's the connection with `refresh_user`

I believe that `refresh_user` can now be implemented in the base class and some of the newly introduced methods could be reused there.

### For the review process
While I  did mess with the logics of things here and there, the goal of this is to get all oauthenticators to the same common ground without removing or breaking any existing logic.

So because this is a monster PR that touches all authenticators, I would suggest concentrating our efforts on making sure that this doesn't break anything by mistake and we can later further improve the actual logic/code.

The tests pass, but I'm not sure they can be trusted, since they are little :(, hence my fear of breaking anything.

### Apologies
I'm very sorry for the amount of effort to review this PR is generating and I'm happy to provide any extra info, answer any question or change anything to make reviewing easier.

File: oauthenticator/auth0.py [hunk 1]
```diff
@@ -55,14 +58,21 @@ def _auth0_subdomain_default(self):
         return subdomain
 
     username_key = Unicode(
-        os.environ.get("OAUTH2_USERNAME_KEY", "email"),
         config=True,
-        help="Userdata username key from returned json with user data login information",
+        help="Deprecated, use `Auth0OAuthenticator.username_claim`",
     )
```
consideRatio, line 63: ## Important: breaking change

There is a breaking change related to the `OAUTH2_USERNAME_KEY` environment variable. Adding this line back, setting the default value of this deprecated trait, is not sufficient though.

Our logic associated with the configuration of `_deprecated_oauth_aliases` is working if we via config directly set `username_key`, but not if we set it indirectly via the default value via the `OAUTH2_USERNAME_KEY` environment variable.

consideRatio: I think it would be great to drop use of these environment variables and require direct configuration, but preferable not as part of this PR.

GeorgianaElena: Yes, I agree! How about we open up an issue about dropping env vars support (if one isnt' open already) after merging this PR and we release them both together as a big breaking version?

Alternatively, I believe we could have the new `username_claim` track OAUTH2_USERNAME_KEY and default to that if not explicitly set as config option?

(Also, I believe there might be another similar situation with another env var)

consideRatio: > I believe we could have the new username_claim track OAUTH2_USERNAME_KEY and default to that if not explicitly set as config option?

Let's go for this for now and either an inline FIXME note or dedicated issue on how this is something we should stop supporting this in the future.

GeorgianaElena (commented):
Maybe this helps:
- I tried adding some context about the bits that I felt were more sketchy or interesting.  
- Also, one idea for a review approach would be to start with the changes made in the `oauth2.py` file to get a sense of what the general `authenticate` flow is right now, and then check if this looks like it will still  works on all the authenticators. Esp if there are things were left out?

- The way I approached the refactoring was to start from one oauthenticator, then separate the big `authenticate` method into 3 main methods (one that gets the token info, one that get the user info using this token, and one that does all the rest, i.e. returns the expected dict after some processing of the data that needed to go into that), tried to generalize that and moved it to the base class.
  Then, I slowly went through other oauthenticators, and checked if those base methods still made sense for those too. This usually resulted in either adding conditionals, or adding more features/checks to the base methods to make it work for each of the new oauthenticator I was checking
- Unfortunately I heavily relied on tests to make sure the features that were "borrowed" from other oauthenticators (by moving them to the base class) won't break the authentication. But because I don't think the tests cover all edge cases, I cannot be confident the changes here won't break existing configurations under any circumstances

yuvipanda:
We're already running into rebase issues here, unfortunately :(

@consideratio @minrk would either of you have time in say the next two weeks to look at this? If not, I can try find time.

oauthenticator/oauth2.py
```diff
@@ -277,8 +288,45 @@ def _token_url_default(self):
...
+        The field in the userdata response from which to get the JupyterHub username.
+        Examples include: email, username, nickname
+
+        What keys are available will depend on the scopes requested and the authenticator used.
+        """,
+    )
+
+    @default("username_claim")
+    def _username_claim_default(self):
+        return "username"
```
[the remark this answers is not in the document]
minrk: Don't need a dynamic default for a static value. Subclasses can still override with a dynamic default:

```suggestion
    username_claim = Unicode(
        "username",
        config=True,
        help="""Field in userdata reply to use for username
        The field in the userdata response from which to get the JupyterHub username.
        Examples include: email, username, nickname

        What keys are available will depend on the scopes requested and the authenticator used.
        """,
    )

```

oauthenticator/oauth2.py
```diff
@@ -277,8 +288,45 @@ def _token_url_default(self):
...
+
+    # Originally a GenericOAuthenticator only trait
+    userdata_token_method = Unicode(
+        os.environ.get("OAUTH2_USERDATA_REQUEST_TYPE", "header"),
+        config=True,
+        help="Method for sending access token in userdata request. Supported methods: header, url. Default: header",
+    )
+
+    # Originally a GenericOAuthenticator only trait
+    extra_params = Dict(
```
minrk, line 326: Should we take this opportunity to call it `token_params` to be consistent with `userdata_params`?

```suggestion
    token_params = Dict(
```

oauthenticator/oauth2.py
```diff
@@ -426,8 +481,274 @@ def get_handlers(self, app):
...
+        Called by the :meth:`oauthenticator.OAuthenticator.token_to_user`
+        """
+        return {
+            "Accept": "application/json",
+            "Content-Type": "application/json",
+            "User-Agent": "JupyterHub",
+            "Authorization": "{} {}".format(token_type, access_token),
+        }
+
+    def build_token_info_req_headers(self):
```
minrk, line 496: Consistency with userdata req->request:

```suggestion
    def build_token_info_request_headers(self):
```

oauthenticator/oauth2.py
```diff
@@ -426,8 +481,274 @@ def get_handlers(self, app):
...
+            auth_state = await decrypt(encrypted)
+
+            return auth_state.get('refresh_token')
+        except (ValueError, InvalidToken, EncryptionUnavailable) as e:
+            self.log.warning(
+                f"Failed to retrieve encrypted auth_state for {username}. Error was {e}.",
+            )
+            return
+
+    def build_access_tokens_req_params(self, handler):
```
minrk, line 559: Same as above, consistency:

```suggestion
    def build_access_tokens_request_params(self, handler):
```

oauthenticator/oauth2.py
```diff
@@ -426,8 +481,274 @@ def get_handlers(self, app):
...
+            'grant_type': 'authorization_code',
+            'client_id': self.client_id,
+            'client_secret': self.client_secret,
+            'redirect_uri': self.get_callback_url(handler),
+        }
+        params.update(self.extra_params)
+
+        return params
+
+    async def get_tokens_info(self, handler, params):
```
minrk, line 580:
```suggestion
    async def get_token_info(self, handler, params):
```

oauthenticator/oauth2.py
```diff
@@ -426,8 +481,274 @@ def get_handlers(self, app):
...
+            the JSON response to the `token_url` the request.
+
+        Called by the :meth:`oauthenticator.OAuthenticator.authenticate`
+        """
+        url = url_concat(self.token_url, params)
+
+        req = HTTPRequest(
+            url,
+            method="POST",
+            headers=self.build_token_info_req_headers(),
```
minrk, line 594:
```suggestion
            headers=self.build_token_info_request_headers(),
```

oauthenticator/oauth2.py
```diff
@@ -426,8 +481,274 @@ def get_handlers(self, app):
...
+            body=json.dumps(params),
+            validate_cert=self.validate_server_cert,
+        )
+
+        tokens_info = await self.fetch(req)
```
minrk, line 599: Consistency: let's use `token_info` everywhere, instead of plural `tokens_info`.

oauthenticator/oauth2.py
```diff
@@ -426,8 +481,274 @@ def get_handlers(self, app):
...
+
+        if "error_description" in tokens_info:
+            raise web.HTTPError(
+                403,
+                "An access token was not returned: {}".format(
+                    tokens_info["error_description"]
+                ),
+            )
+        elif "access_token" not in tokens_info:
```
minrk, line 608: Can this key vary, or is it the same for every provider?

minrk (commented):
This is great! A few comments about consistency of things, but the changes all look really great.

oauthenticator/oauth2.py
GeorgianaElena: According to https://www.oauth.com/oauth2-servers/access-tokens/access-token-response/ I believe this is a required key for every oauth2 provider

GeorgianaElena:
Thanks a lot for the feedback. I believe I have addressed all the comments and also rebased the PR.
I have also updated some of the deprecation messages that this introduces to mention `16.0.0` instead of `15.1.0` as the new version when these will be shipped. We can change it later if this version gets released in the meantime though.

minrk:
Terrific job, @GeorgianaElena! I think we should probably do a bit of a write-up in the docs about what changed - in particular noting any config that no longer works or is deprecated due to the normalization of config names.

GeorgianaElena:
Yaaay! Thank you 🚀 I've opened https://github.com/jupyterhub/oauthenticator/issues/549 about the docs part and self-assigned to it.

consideRatio:
Wieee!!!! Amazing effort on this @GeorgianaElena!! And @minrk for going all the way with the associated review efforts!

:Tada: :heart: 🌻

consideRatio:
Following this work, is this known to be breaking or not in some way - besides possible fixes already merged?

PR #531, opened 2022-08-08
[All] Add post_logout_redirect_uri configuration

pylanglois:
To logout, the openid specs requires a id_token_hint as parameter which is provided along with the access_token. So I added that value to the auth_state. After logout, it is possible to provide a URI where the user is redirected. This URI can be set with the traitlets post_logout_redirect_uri.

https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout


pylanglois:
Note the you need to request the 'openid' scope to get an 'id_token'

`c.GenericOAuthenticator.scope = ["openid"]`

nguyentritai2906:
@GeorgianaElena Any update on this? :)

n0k0m3:
Quick workaround without the need of `id_token_hint`: use `client_id=<clientid>`.
This will allows logout to work, however user will need to confirm on Keycloak/OAuth provider side to logout (2 clicks instead of just 1)

pylanglois:
For some reason, extra_params with client_id and secret needs to be present in your config file.

`c.GenericOAuthenticator.extra_params = {
    "client_id": "your-client",
    "client_secret": "your-secret",
}`

yavorivanov-cw:
Any news on that?
We want to use `post_logout_redirect_uri` but for now we have to use it with `client_id` instead of `id_token_hint`. This requires the user the press logout on the keycloak logout page instead only on logout in jupyterhub UI.

Issue #536, opened 2022-09-06
[Generic] Allow passing period-separated string for nested `claim_groups_key` in generic_client


dingobar:
### Proposed change

Allow passing for example `claim_groups_key=permissions.groups` to `GenericOAuthenticator`, to extract the groups from a claim with the form
```json
{
    "permissions": {
        "groups" : ["mygroup1", "mygroup2"]
    }
}
```

### Alternative options

It is currently possible to do this by passing a `Callable()` that returns the groups given the user object, but I find this not configurable enough. It is also very finnicky (probably impossible) to use with the kubernetes-hosted jupyter hub.

### Who would use this feature?

For example, KeyCloak exposes its group claim under `realm_access.roles`. Therefore, I cannot currently implement authorization on our kubernetes-hosted jupyterhub instance.

### (Optional): Suggest a solution

 Similar classes in for example Java spring boot use a period to separate nested keys. I think we should do the same.

**This will be a breaking change for users who have a key with a period in it**. I think this is an edge case but do you think we should design for it?

I implemented a quick fix for this in this pr: https://github.com/jupyterhub/oauthenticator/pull/537

Any feedback is appreciated on this.

minrk:
Are there any specs for what are legal characters for claim names? I think both `.` and `/` are logical separators for nested JSON paths (i think the '.' you've chosen is the more natural of the two), and should be safe if either is _illegal_ in a field name.

Including `.` in a field name would be weird for JSON, since it would mess up JSON->javascript object deserialization (`obj["field.key"] != obj.field.key`).

dingobar:
Yes, I think it is rather safe, and I would call it an edge case if someone were using it like that, but it is still valid JSON. I don't have a strong opinion on `.` vs `/`, even making that configurable.

minrk:
I think that's sensible, and something like the basics of jq query syntax for accessing list items only if/when someone needs it. A callable will always be there as an escape hatch.

dingobar:
Implemented in https://github.com/jupyterhub/oauthenticator/pull/537

What is your release cycle like? Could we get this one out as a minor bump some time this month maybe?

consideRatio:
I think only a minor version bump is relevant for this, I'll look into making one directly because why not.

consideRatio:
Done!

PR #537, opened 2022-09-06
[Generic] Allow passing a string separated by periods for group claim key

dingobar:
This implements support for nested dict keys in the group claim key, by allowing the user to specify the key as a string of period-separated keys.

See https://github.com/jupyterhub/oauthenticator/issues/536


dingobar:
First question:
This will throw `TypeError` if the key does not exist, whereas before it would just return None. Is it fair enough if I add a try-except for that, and just set the groups to `None` if `TypeError` is caught?

minrk (approved):

yuvipanda:
Thanks, @dingobar!

PR #552, opened 2022-11-25
Docs and test refactoring

GeorgianaElena:
Before https://github.com/jupyterhub/oauthenticator/issues/549, I believe some maintenance work needs to happen.
This PR:
- moves `repo_root/oauthenticator/tests` to `repo_root/tests` (to be in line with other jupyterhub repos)
- changes the `.rst` files to `.md` files and hopefully much of the old syntax with myst markdown one

## Note

The docs content was not changed more than a few typos I spotted here and there, to not bloat this PR more.

PR #553, opened 2022-11-25
Relocate tests to the root repo directory

GeorgianaElena:
Ref: https://github.com/jupyterhub/oauthenticator/pull/552#issuecomment-1327421312 and https://github.com/jupyterhub/team-compass/issues/588


consideRatio (commented):
I'm not confident about merging this. I see that there is mocks.py with things that I think could be relevant and used by someone downstream developing their own subclass, but I'm not sure.

If nobody is using these things, it is fine to merge in my mind, but its hard to know if someone does.

I searched for `oauthenticator.tests.mocks` and got two hits.

- This doesn't look like a problem, unsure what it is but its just listing a lot of stuff:
  https://github.com/search?q=oauthenticator.tests.mocks&type=code
- This on the other hand would be a problem, making it a breaking change.
  https://github.com/brazil-data-cube/bdc-jupyterhub-oauth

manics:
I think we should make the decision based on how we intend the tests to be used or not used, taking into account all the points raised in https://github.com/jupyterhub/team-compass/issues/588

If we merge this it should be highlighted as a breaking change in a major release. It's disruptive to anyone relying on importing tests, but if we're keeping it as supported that implies we should avoid making breaking changes to the tests in a minor release, and potentially increases the maintenance burden (what parts of the tests are considered public and reusable?)

Either option is fine as long as it's carefully considered.

consideRatio:
I figure this is a judgement call, I'm okay with either merge or close, leaning towards wanting to do what other people opine towards.

@GeorgianaElena how do you think about this at the moment?

GeorgianaElena:
Given the context in https://github.com/jupyterhub/team-compass/issues/588, I believe I will close this and stick to the rule of having the tests dir in the top level directory for new projects.

Thanks a lot for all your input @consideRatio and @manics!

PR #559, opened 2022-12-06
maint: drop support for python 3.6

consideRatio:
I don't think we need to call it breaking to drop support for the python version 3.6 as its reached end of life, but I'm not sure.


manics (approved):
Other projects don't, but I think it's good to create a major release as it communicates a big change. It's officially EOL but in practice it's still supported by some enterprise Linux distributions.

Issue #549, opened 2022-11-18
Document the changes introduced by latest refactor


GeorgianaElena:
### Context

https://github.com/jupyterhub/oauthenticator/pull/526 introduces some changes like configuration deprecations in order to achieve a certain "normalization" of the authenticators.

### Proposal

- [ ] Document the changes in #526  in the changelog and also create a migration guide for when a new version will be released
- [ ] Improve the  documentation of the base `OAuthenticator` class with latest info

consideRatio:
Wieee excellent work!

I removed the maintenance label and retained documentation label, to help github-activity tool generate a changelog where this PR shows up only in one category: documentation.

adnion:
tornado.auth still uses the old name 'extra_params' instead of 'token_params' (https://github.com/tornadoweb/tornado/blob/341e8900524846a2c5e7a9ce5be4448e13d2d3d7/tornado/auth.py#L558)
```
    Traceback (most recent call last):
      File "/usr/local/lib/python3.10/dist-packages/tornado/web.py", line 1711, in _execute
        result = method(*self.path_args, **self.path_kwargs)
      File "/usr/local/lib/python3.10/dist-packages/oauthenticator/oauth2.py", line 106, in get
        self.authorize_redirect(
    TypeError: OAuth2Mixin.authorize_redirect() got an unexpected keyword argument 'token_params'
```

GeorgianaElena:
Thanks for reporting this bug @bcm0! I've opened https://github.com/jupyterhub/oauthenticator/pull/551 to fix it.

https://github.com/jupyterhub/oauthenticator/pull/526 was a big PR and there might be other things missing. If you spot anything else, please feel free to open a new bug report

jimdigriz:
...remember to include "breaking the Access Token Request described in [RFC6749 section 4.1.3](https://www.rfc-editor.org/rfc/rfc6749#section-4.1.3)".

Since the refactor, `application/json` is now used rather than `application/x-www-form-urlencoded` which you can see in https://github.com/jupyterhub/oauthenticator/blob/6ba490b00c26b91a8b7f3ed679b3f39db25c6eef/oauthenticator/oauth2.py#L573-L590

It causes some endpoints to just reject the request altogether.

[...]

PR #561, opened 2022-12-16
Refactor the documentation structure

GeorgianaElena:
I wanted to pair this PR with https://github.com/jupyterhub/oauthenticator/issues/549. But I'm not sure of my time availability and how fast I will be able to do it. So I wouldn't want this more general maintenance PR to get blocked by that.

So, I'm marking this as ready for review noting the following:
- The docs' structure will change following the diataxis framework from:
Screenshot 2023-01-04 at 18 19 56

to
Screenshot 2023-01-04 at 18 27 41

- it also extracts documentation about getting started with specific providers into their own file instead of having them all inside a very big file https://github.com/jupyterhub/oauthenticator/pull/561/commits/72f39ff4310562458bce0b5a6df841d76e1dbd22. This should provide the grounds for improving provider specific setup.


consideRatio (commented):
@GeorgianaElena nice work! I've not looked at the files changed tab now but your overview of changes via picture was very helpful.

- [x] Add a sphinx-rediraffe extension and setup redirects so that links to the previous locations remain functional. For reference on how to do this, see https://github.com/2i2c-org/team-compass/pull/596.
- [ ] What do you think about renaming "Changes in oauthenticator" to "Changelog" like most projects name this section?
- [ ] I think "Migrations" is quite technical and is considering if we should rename it to help users get what its about easier. Its under the how-to section, hmm... what do you think about creating "15.0 migration guide" and "16.0 migration guide" etc or "Upgrading from 14 to 15" etc as how-to guides?

GeorgianaElena:
Thanks for the review @consideRatio. I've just updated the PR with some changes

docs/source/reference/api/gen/oauthenticator.auth0.rst
```diff
@@ -0,0 +1,9 @@
+oauthenticator.auth0
+====================
+
+.. automodule:: oauthenticator.auth0
+
+
+.. autoconfigurable:: Auth0OAuthenticator
+
+.. autoconfigurable:: LocalAuth0OAuthenticator
```
consideRatio, line 9: The `authenticator.rst.tpl` is used to generate these files, which are files that render nicely with the `autodoc_traits` extension.

Since we generate them, we don't need them to be comitted. So let's remove them and update the .gitignore file to reflect this folder was relocated and no longer gitignored.

consideRatio: Also, update the `make devenv` entry in the Makefile / make.bat files as they use an `--ignore` flag to not rebuild because the generated files are updated by the build itself.

consideRatio (changes requested):
Nice work, this looks good with some minor technical things to fix!

docs/source/reference/api/gen/oauthenticator.auth0.rst
GeorgianaElena: Good idea to delete them, thank you! I've just added a commit that does that and also added a `clean` target, so that they get cleanup up on `make clean` too  https://github.com/jupyterhub/oauthenticator/pull/561/commits/60d7c07b00accb5a5e750dcf7b2e12d95eb2e28c

consideRatio: Good idea to let `clean` support cleaning up those files.

The commit removed `--ignore` for sphinx-autobuild rather than updating it to reflect the location of files to not watch for changes, so when trying `make devenv` i got a flurry of rebuilds that never stopped.

GeorgianaElena: Ah, I understand! Fixing it now.

GeorgianaElena: Done @consideRatio! Had to include the `gen` directory too in the `ignore` glob https://github.com/jupyterhub/oauthenticator/pull/561/commits/6f09254fc72eeb12e3c2d0b361aff7b166d28d43

docs/make.bat
```diff
@@ -47,7 +47,7 @@ if errorlevel 9009 (
 	echo.The 'sphinx-autobuild' command was not found. Open and read README.md!
 	exit /b 1
 )
-sphinx-autobuild -b html --open-browser "%SOURCEDIR%" "%BUILDDIR%/html" %SPHINXOPTS%
+sphinx-autobuild -b html --open-browser "%SOURCEDIR%" --ignore " "*/reference/api/*" "%BUILDDIR%/html" %SPHINXOPTS%
```
consideRatio, line 50: Three quote symbols here!

docs/make.bat
```diff
@@ -39,7 +47,7 @@ if errorlevel 9009 (
...
```
GeorgianaElena, line 50: @consideRatio, this one?

```suggestion
sphinx-autobuild -b html --open-browser "%SOURCEDIR%" --ignore "*/reference/api/*" "%BUILDDIR%/html" %SPHINXOPTS%
```

consideRatio:
After building stuff with `make html`, this shows up. I expected no changes, so I think there are things we should update in .gitignore

image

GeorgianaElena:
@consideRatio, do you think this is in a good enough shape to be merged?

minrk (approved):

consideRatio:
Thank you soo much @GeorgianaElena for your work on this!!!! :heart: :tada:

PR #566, opened 2023-01-17
ci: add dependabot for github actions and update misc versions in workflows

consideRatio:
I'm adding the label `ci` (Continuous integration) as this is changes related to _running tests_ etc, not even development of the tests themselves. If looking for a bug in the actual software, one should be able to dismiss that it has been introduced a bug because this PR is solely related to `ci` stuff.

With dependabot (an official github service) setup like this, it will open PRs for us when versions are outdated once a month. It won't update all versions I've updated in this PR (not ubuntu version or python version), but it will update all referenced github action's versions (`actions/checkout@v2` -> `@v3`). For example https://github.com/jupyterhub/jupyterhub/pull/4300.

Typically we have configured dependabot to provide weekly updates to github actions, but I figured we should go with monthly and will seek agreement about that in todays JupyterHub team meeting as well.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,16 @@
+# dependabot.yaml reference: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+#
+# Notes:
+# - Status and logs from dependabot are provided at
+#   https://github.com/jupyterhub/oauthenticator/network/updates.
+# - YAML anchors are not supported here or in GitHub Workflows.
+#
+version: 2
+updates:
+  # Maintain dependencies in our GitHub Workflows
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      interval: monthly
+      time: "05:00"
+      timezone: "Etc/UTC"
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -24,12 +24,12 @@ on:
 
 jobs:
   build-release:
-    runs-on: ubuntu-20.04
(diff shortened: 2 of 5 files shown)


Sheila-nk:
@consideRatio This looks great to me! 👍 

I didn't have sufficient knowledge of workflows but I had to do a bit of research before leaving a review. I can say this PR has helped me quite a bit. I am sure that knowledge will come in handy as we continue building the JupyterHub Pytest Plugin. 

Just so I understand completely, you manually updated the github actions versions and then added dependabot.yaml which will automatically do that work for you monthly?

Sheila-nk:
> I'm adding the label `ci` (Continuous integration) as this is changes related to _running tests_ etc, not even development of the tests themselves. If looking for a bug in the actual software, one should be able to dismiss that it has been introduced a bug because this PR is solely related to `ci` stuff.

Could you please also expound on this part?

consideRatio:
> Just so I understand completely, you manually updated the github actions versions and then added dependabot.yaml which will automatically do that work for you monthly?

Exactly so! Dependabot would have generated one PR for action (actions/setup-python, actions/checkout is two actions for example), so i liked to get it done myself to avoid several PRs.


About ci label and focus of the PR. I meant that if we stick to making a PR that dont modify the software we publish, but only modify the documentation or stuff to run tests etc - its quite helpful to highlight that.

Such PRs are not of relevance for a user considering "is there some new feature for me to use?" or "is this something that caused something to break as part of upgrading?". As a maintainer, one often need to track down a bug that has showed up between a version upgrade, so it also helps us as maintainers.

The labels can be used to generate the changelog by grouping entries based on labels.

Issue #565, opened 2023-01-17
Incorrectly sending the `client_client & secret` when basic auth is enabled


GeorgianaElena:
### Bug description

As per https://www.rfc-editor.org/rfc/rfc6749#section-2.3.1, the `client_id` and `client_secret` should not be included in the access token request body when basic authentication is used.

A very useful thread for referece is https://github.com/jupyterhub/oauthenticator/issues/273

#### Actual behaviour

 - This is where the client creds are set https://github.com/jupyterhub/oauthenticator/blob/fd9c7762a3f4567b06522a271c695cafcfd6cbd0/oauthenticator/oauth2.py#L561-L569

 - and this is where is sent https://github.com/jupyterhub/oauthenticator/blob/fd9c7762a3f4567b06522a271c695cafcfd6cbd0/oauthenticator/oauth2.py#L588

- and here is where the basic auth Authorization header is set https://github.com/jupyterhub/oauthenticator/blob/fd9c7762a3f4567b06522a271c695cafcfd6cbd0/oauthenticator/oauth2.py#L495-L500

consideRatio:
Thank you soo much @Georgiana for how thorough yet concise you write issues!!! Perfect!! :heart: :tada: :sunflower:

Issue #273, opened 2019-06-13
[Generic] client_id / client_secret not passed in POST request to get an access token


chucklay:
In generic.py, the POST made to exchange the auth code for an auth token does not include the client_id and client_secret by default:
https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/generic.py#L89-L93

client_id and client_secret are required by oauth spec when making this exchange: 
https://developer.okta.com/blog/2018/04/10/oauth-authorization-code-grant-type#exchange-the-authorization-code-for-an-access-token

It seems like they could be passed in via `self.extra_params`, but so far I can't find where those extra_params should be set.

inviscid:
Is there a legitimate reason not to include the client_id and client_secret values?  The quick fix is to simply include them in the POST parameters.  Examples where this is done in the other service specific POST params:

Github: https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/github.py#L102-L103

Gitlab: https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/gitlab.py#L100-L101

Bitbucket: https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/bitbucket.py#L64-L65

Azure: https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/azuread.py#L73-L74

manics:
Is this still an issue for you?

consideRatio:
Reading in https://tools.ietf.org/html/rfc6749#section-3.2 I conclude that the specification say that we MAY send `client_id`, but it doesn't mention sending along `client_secret`, so it seem like a practice not part of the OAuth2 standard when interacting with the Access Token endpoint.

Apparently though, some Identity providers require it, [like Okta seem to do](https://developer.okta.com/blog/2018/04/10/oauth-authorization-code-grant-type#exchange-the-authorization-code-for-an-access-token).

I think it makes sense at least to pass along `client_id` because that is not a problem security wise to include I think, but I think since its not part of the OAuth2 standard I dislike sending it along to the Access Token endpoint alongside the code which should be enough. I wonder if we really need to send along the client_secret to GitHub's OAauth2 along with the provided code...

Anyhow, this is how you would configure it I think.

```python
c.GenericOAuthenticator.client_secret = "my secret"
c.GenericOAuthenticator.extra_params = {"client_secret": c.GenericOAuthenticator.client_secret}
```

chucklay:
[...]

PR #568, opened 2023-01-18
Adjust the params of the access token request when basic auth is enabled

GeorgianaElena:
Fixes https://github.com/jupyterhub/oauthenticator/issues/565

### DIFF
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -492,7 +492,7 @@ class OAuthenticator(Authenticator):
         """
         headers = {"Accept": "application/json", "User-Agent": "JupyterHub"}
 
-        if self.basic_auth:
+        if not self.basic_auth:
             b64key = base64.b64encode(
                 bytes("{self.client_id}:{self.client_secret}", "utf8")
             )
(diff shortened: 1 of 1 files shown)


oauthenticator/oauth2.py
```diff
@@ -561,11 +561,18 @@ def build_access_tokens_request_params(self, handler, data=None):
...
             "data": data,
         }
+
+        # the client_id and client_secret should not be included in the access token request params
+        # when basic authentication is used
+        # ref: https://www.rfc-editor.org/rfc/rfc6749#section-2.3.1
+        if self.basic_auth:
+            params.update(
+                [("client_id", self.client_id), ("client_secret", self.client_secret)]
+            )
```
consideRatio, line 574: Should this be `if not ...`?

GeorgianaElena: haha, of course :D

consideRatio:
@GeorgianaElena is this a bugfix of a regression introduced in the refactoring PR, or something more as well?

consideRatio:
Ah, I think I understand.

Before refactoring PR we had a bug
After refactoring PR we had another bug
Now this is fixed fully

GeorgianaElena:
I think it's more. I believe, according to https://github.com/jupyterhub/oauthenticator/issues/273, the GenericOAuthenticator (and possibly other oauthenticators also) wasn't including the client creds in the request at all.

So, the refactor, added the creds to the request for _all_ authenticators, but didn't take into account the basic_auth thing.

So, now this fixes that. 

I believe labeling it as a bug, is enough? 

Anyway, after we get https://github.com/jupyterhub/oauthenticator/pull/561 merged, I will focus on documenting the refactorization as best as I can, so we can make a release and see if it actually broke more things than expected 😬

PR #569, opened 2023-01-18
Document the notable changes of the refactorization 

GeorgianaElena:
Not sure if this is the right place for this docs?
I might be missing things, but don't think they should be major.

Fixes https://github.com/jupyterhub/oauthenticator/issues/549

### DIFF
--- a/docs/source/how-to/migrations/upgrade-to-16.md
+++ b/docs/source/how-to/migrations/upgrade-to-16.md
@@ -1,0 +1,70 @@
+# Upgrading your OAuthenticator to version 16.0
+
+The following section describes what to keep in mind when upgrading to OAuthenticator 16.0.
+
+````{important}
+(token_request)=
+
+```{rubric} The token request
+```
+Whenever _token request_ is used, it refers to the HTTP request exchanging the [oauth code](https://www.rfc-editor.org/rfc/rfc6749#section-1.3.1) for the [access token](https://www.rfc-editor.org/rfc/rfc6749#section-1.4).
+
+This request is sent to the [`token_url`](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.oauth2.html#oauthenticator.oauth2.OAuthenticator.token_url) in [`get_token_info()`](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.oauth2.html#oauthenticator.oauth2.OAuthenticator.get_token_info) method.
+
+(userdata_request)=
+
+```{rubric} The userdata request
+```
+Whenever _userdata request_ is used, it refers to the HTTP request that's exchanging the the `access token` to get the `userdata`.
+
+This request is sent to [`userdata_url`](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.oauth2.html#oauthenticator.oauth2.OAuthenticator.userdata_url) in [`token_to_user`](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.oauth2.html#oauthenticator.oauth2.OAuthenticator.get_token_info).
+````
+
+## Deprecations
+
+1. `username_key` replaced by `username_claim` in _all oauthenticators_
+- [`GenericOAuthenticator.username_key`](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.generic.html#oauthenticator.generic.GenericOAuthenticator.username_key) is deprecated and replaced by [`username_claim`](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.oauth2.html#oauthenticator.oauth2.OAuthenticator.username_claim).
+
+- [`Auth0OAuthenticator.username_key`](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.auth0.html#oauthenticator.auth0.Auth0OAuthenticator.username_key) is deprecated and replaced by [`username_claim`](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.oauth2.html#oauthenticator.oauth2.OAuthenticator.username_claim).
+
+    ```{note}
+    The `username_claim` and the deprecated `username_key` refers to the field in the `userdata` response from which to get the JupyterHub username. Examples include: email, username, nickname. What keys are available depend on the scopes requested and the authenticator used.
+    ```
+
+2. [`GenericOAuthenticator.extra_params`](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.generic.html#oauthenticator.generic.GenericOAuthenticator.extra_params) is deprecated and replaced by [`token_params`](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.oauth2.html#oauthenticator.oauth2.OAuthenticator.token_params).
+
+
+3. The following public functions were removed:
+- `OkpyOAuthenticator.get_auth_request(self, code)`
+- `OkpyOAuthenticator.get_user_info_request(self, access_token)`
+
+## New
+
+1. The name of the user key expected to be present in `auth_state` is now configurable through [`user_auth_state_key`](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.oauth2.html#oauthenticator.oauth2.OAuthenticator.user_auth_state_key) for _all oauthenticators_, and it defaults to their prior values for each authenticator.
+
+2. The [`refresh_pre_spawn`](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.oauth2.html#oauthenticator.oauth2.OAuthenticator.refresh_pre_spawn) is enabled by default if [`enable_auth_state`](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.oauth2.html#oauthenticator.oauth2.OAuthenticator.enable_auth_state) is set.
+
+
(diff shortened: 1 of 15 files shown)


docs/source/how-to/migrations/upgrade-to-16.md
```diff
@@ -0,0 +1,70 @@
...
+
+This request is sent to the [`token_url`](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.oauth2.html#oauthenticator.oauth2.OAuthenticator.token_url) in [`get_token_info()`](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.oauth2.html#oauthenticator.oauth2.OAuthenticator.get_token_info) method.
+
+(userdata_request)=
+
+```{rubric} The userdata request
+```
+Whenever _userdata request_ is used, it refers to the HTTP request that's exchanging the the `access token` to get the `userdata`.
+
+This request is sent to [`userdata_url`](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.oauth2.html#oauthenticator.oauth2.OAuthenticator.userdata_url) in [`token_to_user`](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.oauth2.html#oauthenticator.oauth2.OAuthenticator.get_token_info).
```
consideRatio, line 20: It seems like there are triple-backticks that should encapsulate this section, but are instead just start-end on line 16-17

docs/source/how-to/migrations/upgrade-to-16.md
```diff
@@ -0,0 +1,70 @@
+# Upgrading your OAuthenticator to version 16.0
+
+The following section describes what to keep in mind when upgrading to OAuthenticator 16.0.
+
+````{important}
+(token_request)=
```
consideRatio, line 6: Is this important section providing a background to understand other things in the section below?

It seems like descriptions on how things work that may not belong in a version upgrade guide. The technical level here is very high, so I worry a bit that a very limited audience can make use of it.

docs/source/how-to/migrations/upgrade-to-16.md
```diff
@@ -0,0 +1,70 @@
...
+````
+
+## Deprecations
+
+1. `username_key` replaced by `username_claim` in _all oauthenticators_
+
+- [`GenericOAuthenticator.username_key`](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.generic.html#oauthenticator.generic.GenericOAuthenticator.username_key) is deprecated and replaced by [`username_claim`](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.oauth2.html#oauthenticator.oauth2.OAuthenticator.username_claim).
+
+- [`Auth0OAuthenticator.username_key`](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.auth0.html#oauthenticator.auth0.Auth0OAuthenticator.username_key) is deprecated and replaced by [`username_claim`](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.oauth2.html#oauthenticator.oauth2.OAuthenticator.username_claim).
```
consideRatio, line 29: Some indentation matters here and elsewhere, where the content of `1.` isn't indented.

docs/source/how-to/migrations/upgrade-to-16.md
```diff
@@ -0,0 +1,70 @@
...
+The following section describes what to keep in mind when upgrading to OAuthenticator 16.0.
+
+````{important}
+(token_request)=
+
+```{rubric} The token request
+```
+Whenever _token request_ is used, it refers to the HTTP request exchanging the [oauth code](https://www.rfc-editor.org/rfc/rfc6749#section-1.3.1) for the [access token](https://www.rfc-editor.org/rfc/rfc6749#section-1.4).
+
+This request is sent to the [`token_url`](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.oauth2.html#oauthenticator.oauth2.OAuthenticator.token_url) in [`get_token_info()`](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.oauth2.html#oauthenticator.oauth2.OAuthenticator.get_token_info) method.
```
consideRatio, line 12: This is a document parsed by sphinx, so you can make reference that doesn't write like `https://oauthenticator` etc, but instead use the sphinx relative references.

Even the generated reference fields will be sphinx linkable I'm quite sure, hmmm, but how? I think its done by `autodoc_traits` which relies on `sphinx.ext.autodoc`... Hmmm...

I suspect jupyterhub/jupyterhub has made use of this. Maybe something like... Hmmm...

```
{class}`OAuthenticator`
{class}`oauthenticator.oauth.OAuthenticator`
```

@minrk do you have knowledge on how to link to autodoc_traits generated functions/classes etc?

consideRatio (commented):
I had to press send before I had finished this review, but some notes!

Thank you for very thorough work, this looks great overall!!!

docs/source/how-to/migrations/upgrade-to-16.md
GeorgianaElena: It is my understanding that the syntax of a [`rubric` directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-rubric) is to only encapsulate a title. 

I believe other content is not allowed inside the directive. And this is the error got when trying to do it.

```
 ERROR: Directive 'rubric': No content permitted
 ```

### Some context about the `rubric` directive and how I found out about it

The reason why I used this directive was because I wanted to have a heading and to be able to cross-reference it from different places in the documentation.

But a heading inside an ```{admonition} is not permitted and [they get turned into rubrics](https://myst-parser.readthedocs.io/en/latest/develop/_changelog.html#internal-improvements).

So, I thought it's best to be explicit instead of relying on the default behavior.

Do you agree with this reasoning @consideRatio ?

Comment (consideRatio) [reply docs/source/how-to/migrations/upgrade-to-16.md:20]: image

Ah, I had never used this and benefited from your explanations and links. This looks smart to me!!

File: docs/source/how-to/migrations/upgrade-to-16.md (hunk 4 above)
Comment (GeorgianaElena) [reply docs/source/how-to/migrations/upgrade-to-16.md:12]: Good idea @consideRatio! I found this reference in the `jupyterhub` docs for example and I'll try to play with it and see how it goes.

https://github.com/jupyterhub/jupyterhub/blob/36f74689c4dd7c9825da1fae9175469dde926cf9/docs/source/reference/authenticators.md?plain=1#L3

Comment (consideRatio) [reply docs/source/how-to/migrations/upgrade-to-16.md:12]: Ah nice! Yes that is exactly what I was thinking about!

I see that if one clicks it, one goes from...

image

To...

image

File: docs/source/how-to/migrations/upgrade-to-16.md (hunk 2 above)
Comment (GeorgianaElena) [reply docs/source/how-to/migrations/upgrade-to-16.md:6]: > Is this important section providing a background to understand other things in the section below?

It's here more in order to don't repeat this explanation and the links to outside resources multiple times in the sections below.

And also because when reading through the RFC standards, I noticed that for example what we call `token_request` it's called `access token request` there and wanted to make sure people know they're the same thing and I'm not confusing them with possible new names?

Comment (consideRatio) [reply docs/source/how-to/migrations/upgrade-to-16.md:6]: Maybe you could add some introduction sentence between "Important" and the first rubric?

We want this to be something people dare keep reading about, and this is "upgrade instructions", so end users shouldn't need to know the details _unless we have a breaking change we must explain to them_. Currently, its like a snippet of facts and I don't know if its relevant if I'd upgrade and I used for example the GitHubOAuthenticator, etc.

So I think we need some kind of guidance for this notice if its to remain at this location, to clarify for whom this is relevant to read and try to understand.

Comment (GeorgianaElena) [reply docs/source/how-to/migrations/upgrade-to-16.md]: Good point @consideRatio. I've just added a commit that adds these two "definitions" into  footnotes to be less intrusive, but also at hand in case someone wants to double check their meaning:

Check it out rendered at https://oauthenticator--569.org.readthedocs.build/en/569/how-to/migrations/upgrade-to-16.html

File: docs/source/how-to/migrations/upgrade-to-16.md [hunk 5]
```diff
@@ -1,23 +1,6 @@
...
-(userdata_request)=
-```{rubric} The userdata request
-```
-Whenever _userdata request_ is used, it refers to the HTTP request that's exchanging the the `access token` to get the `userdata`.
-
-This request is sent to {attr}`.OAuthenticator.userdata_url` in {meth}`.OAuthenticator.token_to_user` method.
-````
+The following section describes what to pay attention to when upgrading to OAuthenticator 16.0.
 
 ## Deprecations
```
consideRatio, line 5: The title is deprecations which makes me thing that various things still work but is going to stop working. Is this the case, or is it just clean breaking changes, or a mix of breaking changes and deprecations?

It seems that part 1 and 2 are deprecations, but part 3 is a breaking change. Since part 1 and 2 is mentioned to be deprecations, which still could fall under a category of "breaking changes" in my mind, what do you think about changing the title to "Breaking changes" or similar?

image

docs/source/how-to/migrations/upgrade-to-16.md
```diff
@@ -50,21 +33,34 @@ This request is sent to {attr}`.OAuthenticator.userdata_url` in {meth}`.OAuthent
...
+6. It is now possible to set whether or not to use basic authentication for the access token request[^token_request] using {attr}`.OAuthenticator.basic_auth` for _all of the oauthenticators_.
    Currently it defaults to `False`.
 
    ```{note}
    Previously, a GenericOAuthenticator only trait.
    ```
+
+[^token_request]: **The token request.**
+
+    Whenever _token request_ is used, it refers to the HTTP request exchanging the [oauth code](https://www.rfc-editor.org/rfc/rfc6749#section-1.3.1) for the [access token](https://www.rfc-editor.org/rfc/rfc6749#section-1.4).
```
consideRatio, line 57: I consider a section heading saying "foot notes" or similar, feel free to ignore - uncertain what I think.

docs/source/how-to/migrations/upgrade-to-16.md
GeorgianaElena: Makes sense! Thanks @consideRatio

consideRatio (approved):
I think there is a broken link of relevance to update:

(reference/changelog: line   55) broken    https://oauthenticator.readthedocs.io/en/latest/migrations.html#migrating-cilogonoauthenticator-to-version-15-0 - Anchor 'migrating-cilogonoauthenticator-to-version-15-0' not found

Besides that, this LGTM.

GeorgianaElena:
Thanks a lot for reviewing this @consideRatio <3

PR #571, opened 2023-02-01
[All] Adding overrides to user_is_authorized method

npapapietro:
# Motivation

The overall goal of this PR is to expose some override functionality to the `user_is_authorized` [method](https://github.com/jupyterhub/oauthenticator/blob/main/oauthenticator/oauth2.py#L706) that will allow downstream consumers (most notably the [Kubespawner](https://github.com/jupyterhub/kubespawner)) of the authenticator make further decisions on a smaller subset of group/role type properties. My goal is to create a path for allowing auth logic in `KubeSpawner.profile_list`.

# Changes

In each implementation where the concept of `allowed` user groups/projects/roles exist, there is a simple override kwarg option passed.

```python

async def user_is_authorized(self, auth_model, **overrides):
  allowed_groups = overrides.pop("allowed_groups", self.allowed_groups)
  # rest of logic

```

Each override option has been documented in the respective docstring as well as testing added to cover the new changes. Some of the implementations have no concept of allowed groups currently implemented and the contributing guidelines favor moving towards `GenericOAuthenticator`, so they were ignored (cilogin for example).

# Risks

This method is called by the `authenticate` method.

# Mitigations

Testing was created and no changes were made inside the `authenticate` method.

PS:
I also had to bump `isort` bugfix version to fix the pre-commit scripts. Also `flake8==6.0.0` hook breaks for `python==3.7`.


GeorgianaElena (commented):
Thanks a lot for opening this PR @npapapietro! 

Can you please help me understand it better, in order to be able to review it?

> expose some override functionality to the user_is_authorized [method](https://github.com/jupyterhub/oauthenticator/blob/main/oauthenticator/oauth2.py#L706) that will allow downstream consumers (most notably the [Kubespawner](https://github.com/jupyterhub/kubespawner)) of the authenticator make further decisions on a smaller subset of group/role type properties

Can you please describe in more detail how this override functionality added to `user_is_authorized` would be used in a `profile_list` context?

> My goal is to create a path for allowing auth logic in KubeSpawner.profile_list.

I believe @consideRatio created a blog post in https://discourse.jupyter.org/t/tailoring-spawn-options-and-server-configuration-to-certain-users/8449 describing how can the `auth_state` can be used to influence the `profile_list`. And [this is where `auth_state` is built](https://github.com/jupyterhub/oauthenticator/blob/82b1627fc87360f244350fe23d3cb17491cac697/oauthenticator/oauth2.py#L645-L686) before being returned by a successful `authenticate` call.

npapapietro:
Absolutely,

In side the [Kubespawner._render_options_form](https://github.com/jupyterhub/kubespawner/blob/21642eb7de58d66379ea83dc9f73e19531989bb9/kubespawner/spawner.py#L2922) method, one can check/filter the rendered profile list by the base class property `Spawner.authenticator` [here](https://github.com/jupyterhub/jupyterhub/blob/63f164ca53d222d9e65d35f5422c9fd635b577a0/jupyterhub/spawner.py#L161). This would require a PR to the kubespawner to expand this method and altering the schema of the profile list. My suggestion would be to add the respective `**override` kwarg option to the profile list depending on the chosen Authenticator.

For example

```python
c.KubeSpawner.profile_list = [
  {
      'display_name': 'Training Env',
      'slug': 'training-python',
      'default': True,
      # for example
      'authenticator_override': {
         'allowed_groups': ['group1', 'group2'], # generic
         'allowed_organizations': ['org1', 'org2'], # github
      },
      'kubespawner_override': {
        'image': 'datascience/bigGPU:latest',
        'cpu_limit': 48,
        'mem_limit': '96G',
        'extra_resource_guarantees': {"nvidia.com/gpu": "2"},
    }
  ]

```

From here, one can use the z2jh [values](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/46a632f5d07ba8520e5fb5e7539b3f7b89c5fcc7/jupyterhub/values.yaml#L403) to define profile lists depending on Auth without needing to inline python code inside `.Values.hub.extraConfig`

consideRatio:
I currently understand this PR's intent to be to adjust the authorization check (now `check_allowed`) to be usable by other components (like KubeSpawner) to check if an authenticated and authorized user is authorized under different authenticator configuration.

I'd like to steer towards the alternative approach of relying on Authenticator class agnostic information like:
- the user's auth state
- the user's JupyterHub RBAC scopes (permissions)
- the user's JupyterHub groups membership

PR #575, opened 2023-03-09
[CILogonOAuthenticator] Small bugfixes

GeorgianaElena:
Related: https://github.com/jupyterhub/oauthenticator/pull/551

I'm also adding "profile" to the list of default scopes. This scope is needed in order to be able to get the username when authenticating with GitHub. Otherwise, only the email would be available.

Note that `scope` is configurable and can always be modified per need.

### DIFF
--- a/oauthenticator/cilogon.py
+++ b/oauthenticator/cilogon.py
@@ -31,14 +31,14 @@ class CILogonLoginHandler(OAuthLoginHandler):
 
     def authorize_redirect(self, *args, **kwargs):
         """Add idp, skin to redirect params"""
-        token_params = kwargs.setdefault('token_params', {})
+        extra_params = kwargs.setdefault('extra_params', {})
         if self.authenticator.shown_idps:
             # selected_idp must be a string where idps are separated by commas, with no space between, otherwise it will get escaped
             # example: https://accounts.google.com/o/oauth2/auth,https://github.com/login/oauth/authorize
             idps = ",".join(self.authenticator.shown_idps)
-            token_params["selected_idp"] = idps
+            extra_params["selected_idp"] = idps
         if self.authenticator.skin:
-            token_params["skin"] = self.authenticator.skin
+            extra_params["skin"] = self.authenticator.skin
 
         return super().authorize_redirect(*args, **kwargs)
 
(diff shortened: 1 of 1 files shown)


oauthenticator/cilogon.py
```diff
@@ -273,7 +273,7 @@ def user_info_to_username(self, user_info):
             user_info_keys = sorted(user_info.keys())
             if len(claimlist) < 2:
                 self.log.error(
-                    f"Username claim {user_info_keys} not found in response: {self.username_claim}"
+                    f"Username claim {claimlist} not found in response: {user_info_keys}"
```
consideRatio, line 276: I'm not confident about this code, but it seems off that `username claim` (singular) will be followed by a list of some kind.

Should it be this?

```suggestion
                    f"Username claim {self.username_claim} not found in response: {user_info_keys}"
```

But besides that, it seems very odd that the check is `if len(claimlist) < 2`. What makes that relate to if a username claim was found in the response or not?

consideRatio (commented):
Super happy that you find time to keep contributing to this work @GeorgianaElena!!!

- For the changelog, can you update to reflect that this is related to CILogon?
- For the PR label and PR description, I also wonder if this fixes a bug in the main branch and/or in a released version.
  - Min has a practice of labelling fixed to unreleased versions to `maintenance`, and I've labelled them `bug` or `maintenance` historically but start to think it makes more sense to go with `maintenance`.

oauthenticator/cilogon.py
GeorgianaElena: The username claim can either be `self.username_claim` + `self.additional_claims` (this used to exist before self.allowed_idps was introduced)
https://github.com/jupyterhub/oauthenticator/blob/462275b05d84b79966978a3c2ae1f9c2d48acd08/oauthenticator/cilogon.py#L253-L256

Or whatever the specific idp claim in `self.allowed_idps` if that's specified
https://github.com/jupyterhub/oauthenticator/blob/462275b05d84b79966978a3c2ae1f9c2d48acd08/oauthenticator/cilogon.py#L261-L265

So, claimlist can either be a list with one claim or with multiple claims to check if one fails.

I agree the `if len(claimlist) < 2` check is pretty unhappy and confusing, and the different error messages don't make a lot fo sense. I will add a commit to turn them into one.

PR #576, opened 2023-03-10
Add support for signing in with certificate credentials using client …

ctoma73:
This implementation supports the scenario where signing into JupyterHub is needed to be performed using a certificate instead of client secret. This is achieved by passing in a client assertion type and a client assertion in JWT format to exchange the code for an access token.

To securely handle the certificate, it is assumed that the certificate is stored in an Azure KeyVault. A managed identity is needed/recommended to be provisioned with RBAC access to the Azure KeyVault containing the certificate.

Note: In addition to that the new authenticator uses /oauth2/v2.0/token (OAuth2 V2.0) endpoint.

Implementation provides a new authenticator AzureAdOAuthenticatorWithCertificate which subclasses AzureAdOAuthenticator.
Instead of expecting a client_secret a client_assertion_handler must be provided which is a function that is responsible to build a client assertion in JWT format as described in this documentation links:

Reference documentation:
https://learn.microsoft.com/en-us/azure/active-directory/develop/active-directory-certificate-credentials
https://learn.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#request-an-access-token-with-a-certificate-credential

### DIFF
--- a/examples/azuread/sample_jupyterhub_config_with_certificate.py
+++ b/examples/azuread/sample_jupyterhub_config_with_certificate.py
@@ -1,0 +1,74 @@
+import base64
+import datetime
+import jwt, os, uuid
+
+from azure.keyvault.secrets import SecretClient
+from azure.identity import DefaultAzureCredential
+
+from oauthenticator.azureadcert import AzureAdOAuthenticatorWithCertificate
+
+from cryptography.hazmat.primitives.serialization import pkcs12
+from cryptography.hazmat.primitives.asymmetric import rsa
+from cryptography.hazmat.primitives import serialization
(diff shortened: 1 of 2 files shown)


manics:
Thanks for looking into this!

We'd like to avoid adding new authenticators to this repo due to maintenance issues, and inability to properly test authenticators that depend on external providers: https://github.com/jupyterhub/oauthenticator/blob/main/CONTRIBUTING.md

It looks like this only depends on the public interface of OAuthenticator, so it should be straightforward to create your own Python package that has oauthenticator as a requirement. Advantages of this include being easier to release updates, and to add a real production test since you're in full control of any secret credentials required for testing.

Issue #403, opened 2021-01-21
Jupyterhub fails to login using Azure aad with proxy setup


Madhuriv23:
#### Actual behaviour

Our VMs with jupyterhub and oauth installed are behind a proxy. Jupyter comes up fine but fails to login azure-active directory user with the proxy set.
i.e when proxy is set in the shell before starting the jupyter as follows
http_proxy=http://x.x.x.x:xxxx
https_proxy=http://x.x.x.x:xxxx
no_proxy="xxx"

jupyter comes up fine with aadlogin screen, however it fails to login the user with the following error
[D 2021-01-21 16:55:32.169 JupyterHub log:181] 200 GET /hub/logo (@::ffff:10.250.6.187) 0.74ms
[D 2021-01-21 16:55:32.236 JupyterHub log:181] 200 GET /hub/static/favicon.ico?v=fde5757cd3892b979919d3b1faa88a410f28829feb5ba22b6cf069f2c6c98675fceef90f932e49b510e74d65c681d5846b943e7f7cc1b41867422f0481085c1f (@::ffff:10.250.6.187) 0.56ms
[I 2021-01-21 16:55:33.537 JupyterHub oauth2:104] OAuth redirect: 'https://vm-dsvmv2-ihprodtest2.bdaa.prod.ams.azu.dbgcloud.io:8000/hub/oauth_callback'
[D 2021-01-21 16:55:33.538 JupyterHub base:519] Setting cookie oauthenticator-state: {'httponly': True, 'secure': True, 'expires_days': 1}
[I 2021-01-21 16:55:33.539 JupyterHub log:181] 302 GET /hub/oauth_login?next= -> https://login.microsoftonline.com/e00ddcdf-1e0f-4be5-a37a-894a4731986a/oauth2/authorize?response_type=code&redirect_uri=https%3A%2F%2Fvm-dsvmv2-ihprodtest2.bdaa.prod.ams.azu.dbgcloud.io%3A8000%2Fhub%2Foauth_callback&client_id=de9b193b-6b87-4ee8-954e-de1e6e6d99d9&state=[secret] (@::ffff:10.250.6.187) 1.94ms
[...]

Issue #217, opened 2018-09-26
Configuring web-requests to use a proxy


dtandersen:
I think there is something preventing CurlAsyncHTTPClient from accepting defaults.

`AsyncHTTPClient.configure("tornado.curl_httpclient.CurlAsyncHTTPClient", defaults=defaults)` didn't work from config.yaml.

I had to do this hack:

```
  extraConfig: |
    import pycurl
    from tornado.httpclient import HTTPRequest

    def configure_proxy(curl):
        logging.error(curl.getinfo(pycurl.EFFECTIVE_URL))
        # we only want google oauth to use the proxy
        if "google" in curl.getinfo(pycurl.EFFECTIVE_URL):
            logging.error("adding proxy")
            curl.setopt(pycurl.PROXY, "proxy.example.com")
            curl.setopt(pycurl.PROXYPORT, 8080)

    # never do this
    HTTPRequest._DEFAULTS['prepare_curl_callback'] = configure_proxy
```

I don't know why this doesn't work:

```
    import certifi
    from tornado.httpclient import AsyncHTTPClient

    defaults2 = dict(ca_certs=certifi.where())
    defaults2['proxy_host'] = 'proxy.example.com'
    defaults2['proxy_port'] = 8080
    AsyncHTTPClient.configure("tornado.curl_httpclient.CurlAsyncHTTPClient", defaults=defaults2)
```

Actually, I do have one guess. Maybe this overwrite the defaults?

https://github.com/jupyterhub/jupyterhub/blob/8437f47f361aab42d11801703145ababa7372538/jupyterhub/app.py#L1622

zneudl:
I'm having the same issue with Azure AD.  Your dirty hack worked for me as well.  It would be nice to have this fixed though.

`## Setup proxy for OAUTH2
os.environ['PYCURL_SSL_LIBRARY'] = 'nss'
subprocess.call([sys.executable, '-m', 'pip', 'install', '--compile', '--proxy', 'http://149.180.136.225:3128', 'pycurl'])
import pycurl
#defaults = {'proxy_host':'149.180.136.225', 'proxy_port':3128, 'request_timeout':300, 'connect_timeout':60}
#AsyncHTTPClient.configure("tornado.curl_httpclient.CurlAsyncHTTPClient")

def configure_proxy(curl):
        logging.error(curl.getinfo(pycurl.EFFECTIVE_URL))
        # we only want google oauth to use the proxy
[...]

PR #578, opened 2023-03-12
[All] Add `http_request_kwargs` config option

manics:
Adds a new config property `http_request_kwargs` that allows additional keyword args to be passed by default to `tornado.HTTPRequest`.

This can for example be used to pass HTTP proxy configuration.

Closes https://github.com/jupyterhub/oauthenticator/issues/217
Closes https://github.com/jupyterhub/oauthenticator/issues/403


minrk (approved):

manics:
I was hoping to test this in a real environment with a proxy but haven't had time. However I have tested it with a pretend proxy (basically setting up a http listener, tell oauthneticator to use that as a proxy and verifying that a connection request is sent. It doesn't actually work beyond that since it's not a real proxy so the auth fails, but it shows the proxy is being used).

consideRatio (approved):

consideRatio:
Nice!!! Super happy to see how this managed addressed multiple authenticators issues in one go!

PR #579, opened 2023-03-28
feat: Implement `OAuthenticator.refresh_user`

### LINKED ISSUE #398: [OAuth or Generic] Implement refresh_user

Wykiki:
This MR aims to implement the `refresh_user()`'s `Authenticator` method in `OAuthenticator` class.

Following #398 [comment](https://github.com/jupyterhub/oauthenticator/issues/398#issuecomment-1400319340).

At the time of writing, nothing has been tested, and provider specific code would be needed, as the OAuth2 spec does not force the `expires_in` field in the `access_token` response body.

I may not have bandwidth soon to work again on this, feel free to take the lead !

closes #475 
closes #490 
closes #398

### DIFF
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -1018,7 +1018,7 @@ class OAuthenticator(Authenticator):
         # the client_id and client_secret should not be included in the access token request params
         # when basic authentication is used
         # ref: https://www.rfc-editor.org/rfc/rfc6749#section-2.3.1
-        if self.basic_auth:
+        if not self.basic_auth:
             params.update(
                 [("client_id", self.client_id), ("client_secret", self.client_secret)]
             )
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -1320,6 +1320,7 @@ class OAuthenticator(Authenticator):
         """
         if not self.enable_auth_state:
             # auth state not enabled, can't refresh
+            self.log.debug("auth_state disabled, no auth state to refresh")
             return True
         auth_state = await user.get_auth_state()
         if not auth_state:
@@ -1332,12 +1333,15 @@ class OAuthenticator(Authenticator):
         auth_model = None
         try:
             auth_model = await self._token_to_auth_model(token_info)
-        except Exception as e:
-            # usually this means the access token has expired
-            # handle more specific errors?
-            self.log.info(
-                f"Error refreshing auth with current access_token for {user.name}: {e}. Will try to refresh, if possible."
-            )
+        except HTTPClientError as e:
+            # assume any client error means an expired token
+            # most likely 401 or 403 for well-behaved providers
+            if 400 <= e.code < 500:
+                self.log.info(
+                    f"Error refreshing auth with current access_token for {user.name}: {e}. Will try to refresh, if possible."
+                )
+            else:
+                raise
         refresh_token = auth_state.get("refresh_token", None)
         if refresh_token and not auth_model:
             self.log.info(f"Refreshing oauth access token for {user.name}")
@@ -1352,6 +1356,10 @@ class OAuthenticator(Authenticator):
                     f"Error using refresh_token for {user.name}: {e}. Requiring fresh login."
                 )
                 return False
+            else:
+                self.log.debug(
+                    f"Received fresh access_token for {user.name} via refresh_token"
+                )
             # refresh_token may not be returned when refreshing a token
             # in which case, keep the current one
             if not token_info.get("refresh_token"):
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -746,16 +746,6 @@ class OAuthenticator(Authenticator):
                 return client_secret
         return os.getenv("OAUTH_CLIENT_SECRET", "")
 
-    access_token_expiration = Unicode(
-        config=True,
-        default="3600",
-        help="""
-        If the `expires_in` field is  omitted in the OAuth 2.0 token response
-        then this value will be the default expiration, in seconds, of the
-        access token.
-        """,
-    )
-
     validate_server_cert_env = "OAUTH_TLS_VERIFY"
     validate_server_cert = Bool(
         config=True,
@@ -1082,7 +1072,7 @@ class OAuthenticator(Authenticator):
         # ref: https://www.rfc-editor.org/rfc/rfc6749#section-2.3.1
         if not self.basic_auth:
             params.update(
-                [("client_id", self.client_id), ("client_secret", self.client_secret)]
+                {"client_id": self.client_id, "client_secret": self.client_secret,}
             )
 
         return params
@@ -1190,8 +1180,6 @@ class OAuthenticator(Authenticator):
         Returns:
             auth_state: a dictionary of auth state that should be persisted with the following keys:
                 - "access_token": the access_token
-                - "created_at": creation date, in seconds, of the access_token
-                - "expires_in": expiration date, in seconds, of the access_token
                 - "refresh_token": the refresh_token, if available
                 - "id_token": the id_token, if available
                 - "scope": the scopes, if available
@@ -1206,8 +1194,6 @@ class OAuthenticator(Authenticator):
 
         # We know for sure the `access_token` key exists, otherwise we would have errored out already
         access_token = token_info["access_token"]
-        created_at = token_info.get("created_at", time.time())
-        expires_in = token_info.get("expires_in", self.access_token_expiration)
 
         refresh_token = token_info.get("refresh_token", None)
         id_token = token_info.get("id_token", None)
@@ -1218,8 +1204,6 @@ class OAuthenticator(Authenticator):
 
         return {
             "access_token": access_token,
-            "created_at": created_at,
-            "expires_in": expires_in,
             "refresh_token": refresh_token,
             "id_token": id_token,
             "scope": scope,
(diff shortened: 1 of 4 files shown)


oauthenticator/oauth2.py
```diff
@@ -381,6 +382,15 @@ def _client_secret_default(self):
...
 
+    access_token_expiration_env = "OAUTH_ACCESS_TOKEN_EXPIRATION"
+    access_token_expiration = Unicode(
+        config=True,
+        help="""Default expiration, in seconds, of the access token."""
+    )
+
+    def _access_token_expiration_default(self):
+        return os.getenv(self.access_token_expiration_env, "3600")
+
```
Wykiki, line 392: May be replaced by [Authenticator.auth_refresh_age](https://jupyterhub.readthedocs.io/en/stable/api/auth.html#jupyterhub.auth.Authenticator.auth_refresh_age) ?

oauthenticator/oauth2.py
```diff
@@ -644,6 +675,38 @@ async def token_to_user(self, token_info):
...
+        Args:
+            token_info: the dictionary returned by the token request (exchanging the OAuth code for an Access Token)
+
+        Returns:
+            lifetime: a number representing the access token lifetime, in seconds
+
+        Called by the :meth:`oauthenticator.OAuthenticator.build_auth_state_dict`
+        """
+        return token_info.get("expires_in", self.access_token_expiration)
```
Wykiki, line 707: Need specific methods to retrieve refresh-related informations, as other providers may need custom logic.

But what if the required fields are provider via response headers ? May need provider implementation to see if that's relevant.

oauthenticator/oauth2.py
```diff
@@ -718,8 +787,42 @@ async def user_is_authorized(self, auth_model):
...
+        if not auth_state:
+            self.log.info(
+                "No auth_state found for user %s refresh, need full authentication",
+                user,
+            )
+            return False
+
+        created_at = auth_state.get('created_at', 0)
+        expires_in = auth_state.get('expires_in', 0)
+        is_expired = created_at + expires_in - time.time() < 0
```
Wykiki, line 807: May need to refresh before delta reaches 0, to avoid disruption while working.

oauthenticator/oauth2.py
```diff
@@ -579,6 +588,27 @@ def build_access_tokens_request_params(self, handler, data=None):
...
 
+    def build_refresh_token_request_params(self, refresh_token):
+        """
+        Builds the parameters that should be passed to the URL request
+        that renew Access Token from Refresh Token.
+        Called by the :meth:`oauthenticator.OAuthenticator.refresh_user`.
+        """
+        params = {
+            "grant_type": "refresh_token",
+            "refresh_token": refresh_token,
```
johnpmayer, line 599: need to include 'scope' here too - the same one used on initial login, to ensure you get back an equivalent id_token/access_token

johnpmayer: apologies, looks like this is only required on the original authorization code request. the scopes for the token_info and refresh_token requests do not affect the responses

yuvipanda:
I'll be super excited to try get this to eventually land :)

Wykiki:
@yuvipanda  I don't have bandwidth to work on it anymore, so feel free to take over the code !

yuvipanda:
@Wykiki I'll poke around and try to find someone!

yuvipanda:
And thank you for your contribution :)

epstein6:
Hello!  Was any progress made here?  This would be useful for my work, and this looks like a fairly complete implementation (thanks!).

epstein6:
I've been using this with gitlab with a couple of compatibility changes (adding the redirect_url and the client info) and it seems to work.  Are there any other particular changes that would need to be done here for this to be merged?

minrk:
I finally have time to look this over, and I think there is some simplification we can do. In particular, I think we should rely on JupyterHub's own `refresh_user` expiration (`Authenticator.auth_refresh_age`) config rather than adding our own that does the same thing _within_ `refresh_user`, which makes two separate timers interact.

I also think `refresh_user` should _always_ refresh if there is a refresh token available. Motivating use case: `refresh_pre_spawn = True` to maximize the lifetime of a token for a user session. This should always refresh the token at spawn time.

Another affected case where we need to make a choice: oauth without refresh tokens. In the absence of refresh_tokens, should refresh_user:

1. always expire (may be annoying, likely not correct - e.g. GitHub OAuth [does not provide](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/refreshing-user-access-tokens#configuring-your-app-to-use-user-access-tokens-that-expire) a refresh token when using access tokens that don't expire)
2. check if access_token is valid (can check expiration and/or make an API request, e.g. calling `token_to_user`)
3. check if access_token _will be_ valid for a given amount of time (requires `expires_in` and some config about minimum lifetime)

minrk:
This now implements `refresh_user` for the cases:

- `enable_auth_state` is False - never refresh (current default)
- `refresh_token` defined - always refresh
- `refresh_token` unavailable - refresh auth model with current token, force login if no longer valid

I think the one problem with this always-refresh design is that using a refresh_token often _invalidates_ existing access tokens (GitHub does this, for. This is a problem for the case:

1. spawn server with access token
2. later request to Hub after `auth_max_age` calls `refresh_user`, access token provided to server is revoked.

This is very likely to happen, as the activity notifications coming from the user server will eventually trigger `refresh_user` after `auth_max_age`.

As a result, I think preemptively refreshing the access token is not something we should do, even though I really want to do it on spawn start. Unfortunately, JupyterHub doesn't provide Authenticators the info they need to refresh auth on spawn, but not the rest of the time. Plus, it would be a problem for named servers, where multiple spawners may be concurrent - forcefully refreshing the token on every start would actually mean only the latest spawn has a valid token.

minrk:
This now only refreshes access tokens after they expire.

The flow:

- `refresh_user` always refreshes the user model with the current access token
- _if_ that refresh fails (likely due to expired token) and refresh_token is defined, request a new access token

There is still no need to check an expiration date, because we need to attempt to use the token to refresh anyway, and we can assume failure means expiration is likely (note: this handling is currently unconditional - all failures are assumed to indicate a need for a fresh login. Intermittent network issues on the Hub could trigger expired tokens).

GeorgianaElena (approved):
Thanks for picking this up @minrk and for explaining the reasoning behind simplifying the logic and why this doesn't preemptively refreshes the access token.

I think `get_prev_refresh_token` needs to be removed now as it not used and confusing. Other than this, the PR looks ready to merge to me. I think creating a fixture for running the mock-oauth2-provider should be tracked by another issue and PR as a more general improvement to the testing infra.

oauthenticator/oauth2.py
```diff
@@ -1288,24 +1310,84 @@ async def authenticate(self, handler, data=None, **kwargs):
...
         username = self.user_info_to_username(user_info)
         username = self.normalize_username(username)
 
-        # check if there any refresh_token in the token_info dict
-        refresh_token = token_info.get("refresh_token", None)
-        if self.enable_auth_state and not refresh_token:
-            self.log.debug(
-                "Refresh token was empty, will try to pull refresh_token from previous auth_state"
-            )
-            refresh_token = await self.get_prev_refresh_token(handler, username)
```
manics, line 1305: This was the only use of `OAuthenticator.get_prev_refresh_token()` in this repo, so it's never going to be tested. Is the method still useful to others, or can we deprecate/remove it? If we keep it we should update the method doc since it's no longer called by authenticate:
https://github.com/jupyterhub/oauthenticator/blob/90da9d7c21d9b57ec46ed96bda486aca1600ca93/oauthenticator/oauth2.py#L991-L995

oauthenticator/oauth2.py
```diff
@@ -1288,24 +1310,84 @@ async def authenticate(self, handler, data=None, **kwargs):
...
+        return await self._token_to_auth_model(token_info)
+
+    async def refresh_user(self, user, handler=None, **kwargs):
+        """
+        Renew the Access Token with a valid Refresh Token
+        """
+        if not self.enable_auth_state:
+            # auth state not enabled, can't refresh
+            self.log.debug("auth_state disabled, no auth state to refresh")
+            return True
```
manics, line 1324: Can you update the method docstring with a description of the return value? Returning True when the token isn't refreshed is confusing.

manics (commented):
The docstring for `build_auth_state_dict` isn't quite correct anymore, it still says it's (only) called by `authenticate`:
https://github.com/jupyterhub/oauthenticator/blob/90da9d7c21d9b57ec46ed96bda486aca1600ca93/oauthenticator/oauth2.py#L1153
If it's easier you could take out the cross reference and similarly for other docstrings. It's helpful for understanding the code, but it's a pain to keep in sync.

minrk:
Thanks for the review! I think I've addressed it. I removed the unused method, updated the docstrings, and updated all the method cross-links so they resolve properly. See [here](https://oauthenticator--579.org.readthedocs.build/en/579/reference/api/gen/oauthenticator.oauth2.html#oauthenticator.oauth2.OAuthenticator.refresh_user) for the rendered refresh_user docstring.

I think the last question is if this should be enabled by default, and I think the answer is yes. You can opt-out by setting:

```python
c.Authenticator.auth_refresh_age = 0
```

manics (approved):
Should we treat the removal of `get_prev_refresh_token` as breaking, or an internal implementation detail?

minrk:
I don't think it should be considered breaking, but I can put it back as deprecated if folks want.

YStrauchP4:
I'm very keen on this PR and it looks like it's ready to be merged? (Hopeful smile)

PR #590, opened 2023-04-05
maint: replace test-requirements.txt with opt. dependencies

### LINKED ISSUE #583: Declare test as an optional dependency and remove test-requirements.txt
This package bundles tests with the package, so I think it makes sense if we also put the dependencies from test-requirements.txt as an optional dependency (extras_require in setup.py).

consideRatio:
- Closes #583


GeorgianaElena (approved):
Thanks @consideRatio

PR #593, opened 2023-04-10
Drop support for Python 3.7

consideRatio:


GeorgianaElena (approved):
👍🏼

consideRatio:
@minrk wdyt? JuoyterHub 4 retains py37 still, is it okay for this commonly dependency to go for py38+ at this point?

minrk:
I think this is okay, if you want to and it makes things easier. I'm not personally a big fan of dropping support without cause (pure Python packages have a substantially easier maintenance story re: python version than Extensions like numpy).

FWIW, Python 3.7 is the third most popular version to download oauthenticator in the last 30 days, after 3.10 and before 3.9 or 3.11:



python_version | num_downloads
-- | --
3.8 | 9250
3.10 | 6350
3.7 | 4704
3.9 | 4327
3.11 | 1499
3.6 | 984
2.7 | 7
3.5 | 4


<details>
<summary>Google BigQuery</summary>

```sql
SELECT
  REGEXP_EXTRACT(details.python, r"[0-9]+\.[0-9]+") AS python_version,
  COUNT(*) AS num_downloads,
FROM `bigquery-public-data.pypi.file_downloads`
WHERE
  file.project = 'oauthenticator'
  AND details.python IS NOT NULL
  AND DATE(timestamp)
    BETWEEN DATE_TRUNC(DATE_SUB(CURRENT_DATE(), INTERVAL 30 DAY), MONTH)
    AND CURRENT_DATE()
GROUP BY `python_version`
ORDER BY `num_downloads` DESC
```

</details>

And JupyterHub itself is slightly different, in that 3.9 accounts for fully half of all installs and 3.10 is still less widely used than 3.7:


python_version | num_downloads
-- | --
3.9 | 125119
3.8 | 66085
3.7 | 25569
3.10 | 20892
3.11 | 8430
3.6 | 5777
2.7 | 967
3.5 | 177
3.12 | 5
3.4 | 2


All that said, dropping support for a Python version isn't introducing any new problems for folks on old Python, it's just preventing them from getting updates, which is usually fine since they are already avoiding updates by sticking to a  very old Python.

consideRatio:
Thanks for the statistics, this is interesting!

I figure it makes sense to distinguish "known cause" and "precautionary cause", where the precautionary cause is to avoid running into known cause.

> All that said, dropping support for a Python version isn't introducing any new problems for folks on old Python, it's just preventing them from getting updates, which is usually fine since they are already avoiding updates by sticking to a very old Python.

This is a key point in my mind overall.

I figure sometime after Python 3.7 has reached end of life (27 Jun 2023) we drop support for Python 3.7 without necessitating a major release. Is that okay with you @minrk?

Like this, anyone using Python 3.7 can't expect a security patch from oauthenticator going onwards, and that aligns with Python 3.7.x versions as well.

consideRatio:
I opened #596 about dropping support in the future after Python 3.7's EOL.

minrk:
After Python EOL seems reasonable. I think that's a good date to drop support _if no issues are encountered_. The later it is, the lower the bar for me before dropping support for an old Python becomes a reasonable solution to any given problem, but I do believe it should generally be in response to an actual problem (maintenance burden counts as a problem!).

> we drop support for Python 3.7 without necessitating a major release

Yes. In general, I do not believe dropping support for a version of Python is a major revision, despite the conclusion in https://github.com/jupyterhub/jupyterhub/issues/3982 . Similarly, if we choose to define bumping required Python as a major revision, that doesn't have to mean anything significant or a big release. Major releases can be small!

consideRatio:
Okay, Python 3.7 has reached EOL and I saw jupyterhub/jupyterhub dropped support as well, so I'll go for a merge!

consideRatio:
Added breaking label so that its easy to find when writing the changelog's manually written parts

PR #595, opened 2023-04-11
[Maintainance] Remove dynamic defaults when not needed and rm the io_loop

GeorgianaElena:
Enable `asyncio_mode = "auto"`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,3 +32,9 @@ target_version = [
     "py310",
     "py311",
 ]
+
+[tool.pytest.ini_options]
+addopts = "--verbose --color=yes --durations=10"
+asyncio_mode = "auto"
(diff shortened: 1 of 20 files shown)


pyproject.toml
```diff
@@ -32,3 +32,9 @@ target_version = [
     "py310",
     "py311",
 ]
+
+[tool.pytest.ini_options]
+addopts = "--verbose --color=yes --durations=10"
+asyncio_mode = "auto"
+# Ignore thousands of tests in dependencies installed in a virtual environment
+norecursedirs = "lib lib64"
```
minrk, line 40: This is perhaps simplified as `testpaths = [ "oauthenticator/tests" ]` rather than _excluding_ some files

minrk (approved):

Issue #598, opened 2023-04-19
Azure AD Support in master does not work


jabbera:
### Bug description

AzureAD support doesn't work in master because the implementation assumes that the /token endpoint supports json. It does not:

https://learn.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#request-an-access-token-with-a-client_secret

it only supports application/x-www-form-urlencoded

Trying to use it currently results in:

"error_description": "AADSTS900144: The request body must contain the following parameter: 'grant_type'.\r\nTrace ID: 41591746-1955-475c-910b-a1ec80d94900\r\nCorrelation ID: d19ed803-ed48-4b6c-949b-76bc6106112b\r\nTimestamp: 2023-04-19 14:14:31Z"

#### Expected behaviour

It should work!

#### Actual behaviour

It does not!

Issue #564, opened 2023-01-17
The content type of the access token request is missing


GeorgianaElena:
### Bug description

_Originally reported by @jimdigriz in https://github.com/jupyterhub/oauthenticator/issues/549#issuecomment-1367408298_

In the `build_token_info_request_headers` function https://github.com/jupyterhub/oauthenticator/blob/fd9c7762a3f4567b06522a271c695cafcfd6cbd0/oauthenticator/oauth2.py#L493

there's no `"Content-Type"` header to specify the type of the request body, https://github.com/jupyterhub/oauthenticator/blob/fd9c7762a3f4567b06522a271c695cafcfd6cbd0/oauthenticator/oauth2.py#L561-L569

which according to https://www.rfc-editor.org/rfc/rfc6749#section-4.1.3 it should be `"application/x-www-form-urlencoded"`

Also, we're sending it as json instead https://github.com/jupyterhub/oauthenticator/blob/fd9c7762a3f4567b06522a271c695cafcfd6cbd0/oauthenticator/oauth2.py#LL588C18-L588C23

#### Expected behaviour

Suggested fix available at https://github.com/jupyterhub/oauthenticator/commit/d547ac7b6252826954d715f6fce750958c8e1067

PR #599, opened 2023-04-19
Fix Content-Type header, should be x-www-form-urlencoded for token request, and not passed for other GET requests

jabbera:
Use correct encoding for token request. 

Fixes: https://github.com/jupyterhub/oauthenticator/issues/598
Fixes: https://github.com/jupyterhub/oauthenticator/issues/564

### DIFF
--- a/oauthenticator/tests/mocks.py
+++ b/oauthenticator/tests/mocks.py
@@ -2,7 +2,6 @@
 import json
 import os
 import re
-import urllib
 import uuid
 from io import BytesIO
 from unittest.mock import Mock
@@ -141,30 +140,17 @@ def setup_oauth_mock(
         Replies with JSON model for the token.
         """
         assert request.method == 'POST', request.method
-        if token_request_style == 'json':
-            body = request.body.decode('utf8')
-            try:
-                body = dict(urllib.parse.parse_qsl(body))
-            except ValueError:
-                return HTTPResponse(
-                    request=request,
-                    code=400,
-                    reason="Body not JSON: %r" % body,
-                )
-            else:
-                code = body['code']
-        else:
-            query = urlparse(request.url).query
-            if not query:
-                query = request.body.decode('utf8')
-            query = parse_qs(query)
-            if 'code' not in query:
-                return HTTPResponse(
-                    request=request,
-                    code=400,
-                    reason=f"No code in access token request: url={request.url}, body={request.body}",
-                )
-            code = query['code'][0]
+        query = urlparse(request.url).query
+        if not query:
+            query = request.body.decode('utf8')
+        query = parse_qs(query)
+        if 'code' not in query:
+            return HTTPResponse(
+                request=request,
+                code=400,
+                reason=f"No code in access token request: url={request.url}, body={request.body}",
+            )
+        code = query['code'][0]
         if code not in oauth_codes:
             return HTTPResponse(
                 request=request, code=403, reason=f"No such code: {code}"
--- a/oauthenticator/tests/mocks.py
+++ b/oauthenticator/tests/mocks.py
@@ -2,6 +2,7 @@
 import json
 import os
 import re
+import urllib
 import uuid
 from io import BytesIO
 from unittest.mock import Mock
@@ -143,7 +144,7 @@ def setup_oauth_mock(
         if token_request_style == 'json':
             body = request.body.decode('utf8')
             try:
-                body = json.loads(body)
+                body = dict(urllib.parse.parse_qsl(body))
             except ValueError:
                 return HTTPResponse(
                     request=request,
--- a/oauthenticator/tests/test_auth0.py
+++ b/oauthenticator/tests/test_auth0.py
@@ -29,7 +29,6 @@ def auth0_client(client):
         host=auth0_domain,
         access_token_path='/oauth/token',
         user_path='/userinfo',
-        token_request_style='json',
     )
     return client
 
--- a/oauthenticator/tests/test_globus.py
+++ b/oauthenticator/tests/test_globus.py
@@ -156,7 +156,6 @@ def globus_client(client, mock_globus_token_response):
         access_token_path='/v2/oauth2/token',
         user_path='/v2/oauth2/userinfo',
         token_type='bearer',
-        token_request_style='post',
     )
     set_extended_token_response(
         client, 'auth.globus.org', '/v2/oauth2/token', mock_globus_token_response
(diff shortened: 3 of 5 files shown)


GeorgianaElena (changes requested):
Thank you very much for reporting and fixing this @jabbera.

The issue in https://github.com/jupyterhub/oauthenticator/issues/598 seems to be the same with https://github.com/jupyterhub/oauthenticator/issues/564. And since this is impacting all the oauthenticators I believe it's worth modifying the base OAuth class directly instead of allowing it it to be customized by different oautnenticatot classe.

jabbera:
@GeorgianaElena  all set

jabbera:
Does this need anything else?

consideRatio (commented):
# Investigation

<details>
<summary>Click to see investigation</summary>

I looked in the previous release of oauthenticator, version 15.1.0, and concluded that Content-Type was set to x-www-form-urlencoded for: azuread, bitbucket, google.

But:
- azuread, bitbucket, google set Content-Type: application/x-www-form-urlencoded with details about utf-8 etc.
- auth0 set Content-Type: application/json
- cilogon, generic, github, gitlab, globus, okpy, openshift provided no Content-Type header
- mediawiki relied on mwoauth library

Since Content-Type influence what is passed, the fix ought to be ensure we declare Content-Type correctly for all requests based on what data we pass.

- auth0, seems like `"Content-Type": "application/json"` is correct
  ```python
  body=json.dumps(params),
  ```
- azread, seems like `'Content-Type': 'application/x-www-form-urlencoded; charset=UTF-8'` is correct
  ```python
  data = urllib.parse.urlencode(params, doseq=True, encoding='utf-8', safe='=')
  ```
- bitbucket, seems like `"Content-Type": "application/x-www-form-urlencoded;charset=utf-8"` is correct
  ```python
  body=urllib.parse.urlencode(params).encode('utf-8'),
  ```
- google, seems like `"Content-Type": "application/x-www-form-urlencoded"` is correct
  ```python
  body = urllib.parse.urlencode(
              dict(
                  code=code,
                  redirect_uri=self.get_callback_url(handler),
                  client_id=self.client_id,
                  client_secret=self.client_secret,
                  grant_type="authorization_code",
              )
          )
  ```
- generic, not providing a Content-Type, actually urlencodes params, not setting Content-Type header is likeley incorrect
  ```python
  body=urlencode(params),
  ```
- cilogon, github, gitlab, openshift provides an empty string as POST body - not setting Content-Type header is likely correct
- globus, okpy, makes use of a GET request where there is no POST body - not setting Content-Type header is correct

</details>

# Conclusion

Overall, I think we can't provide a standardized Content-Type across our authenticators, but need to let the authenticators decide on it setting that header or not, and to what.

- For GET requests (globus, okpy) and POST requests with a empty-string body (cilogon, github, gitlab, openshift), we should not set Content-Type at all I think.
- For POST requests where we urlencode params (azuread, google, bitbucket, generic), we should set `application/x-www-form-urlencoded` I think.
- For POST requests where we pass json encoded params (auth0), we should set `application/json` I think.

> **EDIT: New conclusion**
> I got this wrong, I do think we can provide a standarized Content-Type across our authenticators, and that should be what all authenticators support thanks to the OAuth2 specification, x-www-form-urlencoded with an utf-8 charset. But we need to be explicit about the content-type, otherwise some assume json etc.

# Related reading

- [About Content-Type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type)
- [About Content-Type with no body for POST requests](https://stackoverflow.com/a/29784642/2220152)

jabbera:
Thanks. This is the opposite of the other feedback I received which asked me to centralize it instead of doing it at the provider level. I'm happy to revert it and add the other changes you mentioned in the other providers. Are we fairly certain this is the desired end state?

oauthenticator/tests/mocks.py
```diff
@@ -140,30 +140,17 @@ def access_token(request):
...
-            try:
-                body = json.loads(body)
-            except ValueError:
-                return HTTPResponse(
-                    request=request,
-                    code=400,
-                    reason="Body not JSON: %r" % body,
-                )
-            else:
-                code = body['code']
```
consideRatio, line 154: This section was removed and only influenced tests of auth0 that had been sent JSON in the Post body before, but now sends an urlencoded body like other does.

oauthenticator/tests/mocks.py
```diff
@@ -140,30 +140,17 @@ def access_token(request):
...
+        if not query:
+            query = request.body.decode('utf8')
+        query = parse_qs(query)
+        if 'code' not in query:
+            return HTTPResponse(
+                request=request,
+                code=400,
+                reason=f"No code in access token request: url={request.url}, body={request.body}",
+            )
+        code = query['code'][0]
```
consideRatio, line 153: This is unchanged code beside intentation change when it was no longer nested in an `else` clause.

oauthenticator/tests/test_globus.py
```diff
@@ -156,7 +156,6 @@ def globus_client(client, mock_globus_token_response):
         access_token_path='/v2/oauth2/token',
         user_path='/v2/oauth2/userinfo',
         token_type='bearer',
-        token_request_style='post',
```
consideRatio, line 159: This was the default value anyhow.

oauthenticator/tests/test_auth0.py
```diff
@@ -29,7 +29,6 @@ def auth0_client(client):
         host=auth0_domain,
         access_token_path='/oauth/token',
         user_path='/userinfo',
-        token_request_style='json',
```
consideRatio, line 32: This was no longer accurate.

consideRatio (approved):
I had to think a lot before feeling confident about this change, the comment https://github.com/jupyterhub/oauthenticator/pull/599#pullrequestreview-1396378578 was a starting point for me to understand this well enough.

I pushed a few commits to finish up some details, but this looks good to me now!

Changes I made:
- Removed irrelevant Content-Type header on GET requests
- Switched to rely on Python's standard library to do the urlencoding
- Updated docstrings with details on Content-Type
- Cleaned up unused test configuration

consideRatio:
Thank you @jabbera for helping us get this fixed! I was wrong to suggest we needed to adjust the Content-Type for individual authenticators I think.

I added an EDIT note in my comment above:

image

jabbera:
Glad we could get this there. The new extension points in the refactor are awesome.

GeorgianaElena (approved):
Awesome work! Thanks @jabbera, thanks @consideRatio 🎉 🚀

jabbera:
It works!

PR #603, opened 2023-04-23
maint: import Callable traitlet from jupyterhub

consideRatio:
We make use of a traitlet configuration type `Callable` that we imported by copy-paste to this project in #305 because we depended on jupyterhub>=0.5 at that time, and jupyterhub 0.5 didn't have the traitlet defined. This was discussed in https://github.com/jupyterhub/oauthenticator/pull/305#pullrequestreview-319401911.

Now we depend on `jupyterhub>=1.2`, and [the traitlet has been around](https://github.com/jupyterhub/jupyterhub/blob/1.2.0/jupyterhub/traitlets.py#L98-L112) and [unchanged since then](https://github.com/jupyterhub/jupyterhub/blob/4.0.0/jupyterhub/traitlets.py#L98-L112).

### Linkcheck failure

- Resolved in #604

### Breaking changes

- The `oauthenticator.traitlets` module with the `Callable` class was removed but can now be imported from `jupyterhub.traitlets` instead.


manics (approved):
LGTM.

Is this a breaking change since any subclasses that relied on this will need an update?

consideRatio:
Maaaybe, leta mark it as breaking to be safe now that we make a major relase anyhow!

consideRatio:
I updated the PR description to include a Breaking change comment for the changelog for now.

GeorgianaElena (approved):

Issue #605, opened 2023-04-24
Cannot override username_claim for helm chart anymore when using master


jabbera:
### Bug description

This change: https://github.com/jupyterhub/oauthenticator/pull/595 removed the ability to have a custom claim name from the helm chart. For example the AzureAD authenticator now forces you to use name: https://github.com/jupyterhub/oauthenticator/blob/296aa57007909bd5baff53c1b4d6159155deff54/oauthenticator/azuread.py#L23

I use a custom attribute called samAccountName that maps back to people on prem values. Why was this configuration option removed?

consideRatio:
Nice catch @jabbera!! I this is an unintended regression from #595.

### What happened?

A change like below was made.

```diff
-    @default("username_claim")
-    def _username_claim_default(self):
-        return "email"

+    username_claim = "email"
```

The `username_claim` that previously was a traitlet configurable object got overridden with a string, so there is no longer a traitlet configurable via `c.Auth0Authenticator.username_claim`.

### What to do?

I think mostly revert #595, I think the principles should be:

- If a class redefines the config default for a super class' traitlet, then I think it needs do so with the `@default` decorator to avoid overriding it.

  **Like this**

  ```python
      @default("username_claim")
      def _username_claim_default(self):
          return "email"
  ```
  
  **Not like this**
  
  ```python
      username_claim = "email"
  ```

  **Not like this either**

  I think doing this could cause a subtle issue, where configuring `c.OAuthenticator.username_claim` becomes different from configuring `c.AzureADOAuthenticator.username_claim`.

  ```python
      username_claim = Unicode(
          "my default value",
          config=True,
      )
  ```
  
- If a class defines a new traitlet on its own, then I think it should set its default value as part of defining the traitlet.

  ```python
      new_config = Unicode(
          "my default value",
          config=True,
      )
  ```

PR #608, opened 2023-04-26
Fix defaults

GeorgianaElena:
- Reverts the removal of the dynamic default setting in each provider for `username_claim` as that had unwanted consequences, see https://github.com/jupyterhub/oauthenticator/issues/605
- Also, updates the tests of the providers that were impacted and although they were testing setting `username_claim`, the bug remained uncaught, because we were actually testing python's class attributes set, instead of the configurable traitlet
- From #595 this PR reverts the following attrs to being `traitlets.config.Config` again
   - `username_claim`
   - `scope` (for Google OAuthenticator)
   
   **but keeps the following attributes as const strings**
   - `user_auth_state_key`
   - `authorize_url`
   - `token_url`
   - `userdata_url`
   - `revocation_url` (globus) (todo)
   - `globus_groups_url`(globus) (todo)
   - `google_api_url` (todo)

- [ ] todo add reasoning for why keeping those strings


oauthenticator/tests/test_azuread.py
```diff
@@ -33,7 +34,7 @@ def user_model(tenant_id, client_id, name):
             "iat": now,
             "nbf": now,
             "name": name,
-            "preferred_username": name,
+            "preferred_username": "preffered",
```
consideRatio, line 37: A typo here, should be "preferred", but I don't understand the change from `name` to `"preferred"`.

GeorgianaElena: I just wanted to generate another different test. To have preffered be its own thing. But we can leave it like before if it's confusing

consideRatio: Ah well its a mocked response, I think it will be `name` consistently, but we should reasonably respect the `preferred_username` rather than pick a username based on `name`, so good point about testing something different in this mocked response!

GeorgianaElena:
@consideRatio, how do you feel about making `authorize_url`, `userdata_url` etc strings and not have them be configurable? I believe that these don't make sense to be changed by the user, but expect these to change when the provider endpoint changes.

But, for Auth0, for example, [as these links depend on the auth0 domain](https://github.com/jupyterhub/oauthenticator/blob/61786096f02422283250b37218e5688a0733c4f0/oauthenticator/auth0.py#L85-L87), etc, it makes sense to have them be set dynamically. Otherwise if we're keeping them as strings and the `auth0` domain  changes after the authneticator instance is created, then the urls won't know to change.

Hmm, do you agree with this reasoning?

consideRatio:
@GeorgianaElena I _think_ it would be okay to hardcode this, reasoning that if they change, for example an api version goes from v2 to v3, we would still likeley need to fix something in oauthenticator.

- The key value I see in this is hiding the configuration option for the users if they aren't expected to configure it etc.
- I think the complexity of the code is similar, because we still need to allow GenericOAuthenitcator to let these be configurable etc.

I'm not sure overall what I think overall.

consideRatio:
Since I'm unsure what I think, and you now lean towards letting them remain configurable, I figure that is what we should do.

A good thing about doing that is that we don't have to think about those changes as part of this release and don't take a risk of causing a regression or warnings for users about traits not recognized etc.

GeorgianaElena:
I agree to be less disruptive 👍🏼 Then from the list in the top comment, the only thing that I believe it makes sense to be kept hardcoded is `user_auth_state_key` which defines the name of the key under which the data about the user will be stored in `auth_model[auth_state]` and returned by the `authenticate` function. 

For example, this is called `cilogon_user` by cilogon. I don't see why this would need to be configurable, other than personal preference?

consideRatio:
@GeorgianaElena I agree, `user_auth_state_key` doesn't make sense to be configurable by users - removing that as an option is a feature to make users less distracted/confused when reading API reference.

If users have configured this for some reason, and, if they also themselves use the state from a `post_auth_hook` using a hardcoded string instead of using `user_auth_state_key`, then they have issues. Its quite a wild case, and I think we can consider this not breaking. But I don't think it merits `breaking` in the changelog.

GeorgianaElena:
@consideRatio, I've updated the PR to only have `user_auth_state_key` be const and left all the urls to providers' endpoint configurable.

oauthenticator/okpy.py
```diff
@@ -11,21 +11,13 @@
...
 
-    @default("token_url")
-    def _token_url_default(self):
-        return "https://okpy.org/oauth/token"
+    token_url = "https://okpy.org/oauth/token"
 
-    @default("userdata_url")
-    def _userdata_url_default(self):
-        return "https://okpy.org/api/v3/user"
+    userdata_url = "https://okpy.org/api/v3/user"
```
consideRatio, line 20: This was unexpected, they go the other way

consideRatio (commented):
This looks great! I spotted something unexpected in okpy but besides that this is ready to be merged!

I also updated the title btw as I thought it was too brief and technical

oauthenticator/okpy.py
GeorgianaElena: Oh, oops, forgot I made changes in the other way too.

consideRatio (approved):
Nice LGTM!

This is ready to be merged in my mind, do you agree? If so, go for a self-merge!

PR #612, opened 2023-05-03
WIP: changelog for next release

manics:
`github-activity --since 15.1.0`, followed by recategorisation of some entries


manics:
Actually I don't think this is so urgent, AFAICT the code fixed in https://github.com/jupyterhub/oauthenticator/pull/610 wasn't released in the first place.

consideRatio:
I see no urgency besides wanting it bundled with z2jh 3 and tljh 1.

PR #625, opened 2023-06-26
[All] breaking, add allow_all config defaulting to False (CILogon: require allowed_idps)

### LINKED ISSUE #609: Discussion: don't default to allowing all users?
### Proposed change
Don't default to allowing all users to login.

### Alternative options
Leave unchanged

Sysadmins who expect a "secure by default" configuration

As mentioned in https://github.com/jupyterhub/the-littlest-jupyterhub/issues/884 authenticators default to allowing anyone to login. Whilst this has been the case for a while it'd be safer to default to no-one unless explicitly configured, e.g. with a new configurable, or perhaps by using a special placeholder such as `allowed_users = "*"` (deliberately overriding the configurable to be list-or-string to allow this placeholder string).

The big downside is it's a breaking change, but with the changes in https://github.com/jupyterhub/oauthenticator/pull/594 now might be a good time to do this.

### LINKED ISSUE #615: [CILogon] Require specification of `allowed_idps`?
I understand the list in `allowed_idps` as the identity providers possible to use via CILogon. Is there a use case to allow that to be unset? It seems like allowing for that adds some complexity in implementation and documentation on communicating how this authenticator should be configured.

I suggest that if we don't see a use case for allowing it to be unset, we make a breaking change and forcing it to be set to something.

### LINKED ISSUE #620: [All] Stop allowing all authenticated users by default
Our authenticators default to allowing all authenticated users if no set of allowed users are configured.

I think this default should change to not allow all authenticated users in those situations. At the same time, it should be possible to allow all authenticated users.

To accomplish this, I suggest we don't allow any user unless explicitly allowed by some config, and then also introduce the `allow_all` config defaulting to False, allowing all users to be allowed.

### Related

- This is a practical proposal on how to address #609
- This is a practical proposal in line with the idea for adding `allow_all` config in the Authenticator base class, https://github.com/jupyterhub/jupyterhub/issues/4484

consideRatio:
# [All] `allow_all` config

When setting up a JupyterHub you may with a minimal config using an OAuthenticator based Authenticator end up allowing all authenticated users. Even if they were allowed for a brief amount of time, they can have ended up creating a user for access after login is restricted. This PR addresses this situation by defaulting to not allowing all authenticated users.

## Example use

```python
c.JupyterHub.authentication_class = "github" 
c.GitHubOAuthenticator.client_id = "..."
c.GitHubOAuthenticator.client_secret = "..."

# previously with a config like above all github users were allowed access
# now though allow_all needs to be configured explicitly for that situation
# using the allow_all config, like below
c.OAuthenticator.allow_all = True
```

## Related

- Fixes #609 and fixes #620

# [CILogon] `allowed_idps` required

For CILogon, it was hard to add the `allow_all` functionality independently from the planned change of making `CILogonOAuthenticator.allowed_idps` to be required. Due to that, that change is bundled in this PR.

WIth `CILogonOAuthenticator.allowed_idps` required, it made sense to also stop using `CILogonOAuthenticator`'s `username_claim`, `additional_username_claims`, and `shown_idps`, so use of them is now also causing loud errors. The main motivation for erroring loudly instead of providing a warning is that a misconfiguration here can be a security issue.

## Related

- Fixes #615

# For the reviewer

This is a humongous PR as it was hard to avoid it but I hope you don't panic! I've thought a lot on the commits I've made, making a few relatively small commits with actual changes free from updates to comments and tests which instead are put in other larger commits. With this in mind, I recommend reviewing this commit by commit.

Here are the commits, and only three contain relevant changes, and two of those are cilogon specific changes. That makes this PR include one key commit to review (223c89b - all, breaking change: add allow_all config, default to False). When reviewing that commit, I recommend starting by looking at the oauth2.py file's changes as the other changes build on logic changed in the OAuthenticator base class.

1. 0bd5809 - (trivial) refactor, auth0: add help strings for auth0_domain and auth0_subdomain
2. 2ade700 - (trivial) refactor, globus: fix help string indentation and misc whitespace
3. 2e11e44 - cilogon, breaking change: allowed_idps must be configured
4. 223c89b - all, breaking change: add allow_all config, default to False
5. 0e2352e - (only tests/ folder touched) maint: update tests for recent change
6. 561f13c - (~only comments etc touched) maint: update comments, helpstrings, and docstrings for other commits
7. 89f7d9d - cilogon, breaking change: shown_idps, username_claim, additional_username_claims 
8. 84e1aac - (trivial) maint, google: use lower case when comparing domain names
9. 032d5d2 - (trivial) maint, mediawiki: specify user_auth_state_key

**Post review update**

Following review, I've made the following commits so far.

1. 8d1da50 - (trivial) cilogon: switch to another status code in an error response
2. 69535f0 - (trivial) maint: update some additional docstrings
3. e7f864d - (trivial) maint, generic: refactor to at least partially rely on base class

## About changes in tests

- I've added significant test coverage for `allowed_all` and other `allow_...` config by using `@pytest.mark.parametrize` one test per authenticator to run with a lot of different `allow_` config. This parametrized test could largely be re-used across authenticators with small tweaks.
- I've removed quite a few standalone tests that were covered well via the parametrized test, and some other tests. Overall, I've been very cautious to avoid reducing the tested situations.

### DIFF
--- a/oauthenticator/cilogon.py
+++ b/oauthenticator/cilogon.py
@@ -314,7 +314,8 @@ class CILogonOAuthenticator(OAuthenticator):
 
     def _user_info_to_unprocessed_username(self, user_info):
         """
-        Returns a username from
+        Returns a username from user_info without also applying the "action"
+        specified under "username_derivation" for the associated idp.
         """
         user_idp = user_info["idp"]
         username_derivation = self.allowed_idps[user_idp]["username_derivation"]
@@ -330,8 +331,8 @@ class CILogonOAuthenticator(OAuthenticator):
 
     def _get_processed_username(self, username, user_info):
         """
-        This method optionally adjusts a username from user_info based on the
-        "action" specified under "username_derivation" for the associated idp.
+        Optionally adjusts a username from user_info based on the "action"
+        specified under "username_derivation" for the associated idp.
         """
         user_idp = user_info["idp"]
         username_derivation = self.allowed_idps[user_idp]["username_derivation"]
--- a/oauthenticator/cilogon.py
+++ b/oauthenticator/cilogon.py
@@ -30,7 +30,14 @@ class CILogonLoginHandler(OAuthLoginHandler):
     """See https://www.cilogon.org/oidc for general information."""
 
     def authorize_redirect(self, *args, **kwargs):
-        """Add idp, skin to redirect params"""
+        """
+        Optionally add "skin" to redirect params, and always add "selected_idp"
+        (aka. "idphint") based on allowed_idps config.
+
+        Related documentation at https://www.cilogon.org/oidc#h.p_IWGvXH0okDI_.
+        """
+        # kwargs is updated to include extra_params if it doesn't already
+        # include it, we then modify kwargs' extra_params dictionary
         extra_params = kwargs.setdefault('extra_params', {})
         if self.authenticator.shown_idps:
             # selected_idp must be a string where idps are separated by commas, with no space between, otherwise it will get escaped
@@ -92,17 +99,22 @@ class CILogonOAuthenticator(OAuthenticator):
         Unicode(),
         default_value=['openid', 'email', 'org.cilogon.userinfo', 'profile'],
         config=True,
-        help="""The OAuth scopes to request.
+        help="""
+        The OAuth scopes to request.
 
-        See cilogon_scope.md for details.
-        At least 'openid' is required.
+        See cilogon_scope.md for details. At least 'openid' is required.
         """,
     )
 
     @validate('scope')
     def _validate_scope(self, proposal):
-        """Ensure `openid` is always requested and `org.cilogon.userinfo`
-        is requested when allowed_idps is specified.
+        """
+        Ensure `openid` and `org.cilogon.userinfo` is requested.
+
+        - The `idp` claim is required, and its documented to associate with
+          requesting the `org.cilogon.userinfo` scope.
+
+        ref: https://www.cilogon.org/oidc#h.p_PEQXL8QUjsQm
         """
         scopes = proposal.value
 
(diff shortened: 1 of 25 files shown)


oauthenticator/cilogon.py
```diff
@@ -241,144 +273,105 @@ def _validate_allowed_idps(self, proposal):
...
+        #
+        user_idp = user_info.get("idp")
+        if not user_idp:
+            message = "'idp' claim was not part of the response to the userdata_url"
+            self.log.error(message)
+            raise web.HTTPError(500, message)
+        if not self.allowed_idps.get(user_idp):
+            message = f"Login with identity provider {user_idp} is not pre-configured"
+            self.log.error(message)
+            raise web.HTTPError(500, message)
```
GeorgianaElena, line 308: Personally, I think a [403 status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) might be more suited here, since the idp is not **alowed** to login, not just not configured.
Also, since the idps made visible through cilogon's `shown_idps` feature now come from the actual `allowed_idps`, someone seeing this error would need to edit the url and remove the shown_idps query parameters to be presented the entire idp list.

oauthenticator/cilogon.py
```diff
@@ -241,144 +273,105 @@ def _validate_allowed_idps(self, proposal):
...
+        username = self._get_processed_username(unprocessed_username, user_info)
 
         return username
 
-    def user_info_to_username(self, user_info):
-        username_claims = self._get_final_username_claim_list(user_info)
-        username = self._get_username_from_claim_list(user_info, username_claims)
+    def _user_info_to_unprocessed_username(self, user_info):
+        """
+        Returns a username from
```
GeorgianaElena, line 317: I think this comment was left unfinished by mistake?

oauthenticator/cilogon.py
```diff
@@ -241,144 +273,105 @@ def _validate_allowed_idps(self, proposal):
...
         """
-        # A workaround for JupyterHub<=4.0.1, described in
-        # https://github.com/jupyterhub/oauthenticator/issues/621
-        if auth_model is None:
-            return True
+        user_idp = user_info["idp"]
+        username_derivation = self.allowed_idps[user_idp]["username_derivation"]
+
+        # Optionally execute action "strip_idp_domain" or "prefix"
+        action = username_derivation.get("action")
```
GeorgianaElena, line 340: I believe the presence of the action itself must also be optional if for example we want the username to just be email claim with no processing whatsoever.

GeorgianaElena: Ignore this ⬆️ , this is already happening <3

oauthenticator/cilogon.py
```diff
@@ -180,16 +204,19 @@ def _validate_scope(self, proposal):
     def _validate_allowed_idps(self, proposal):
```
GeorgianaElena, line 204: Since using the value of `username_claim` as the default is now deprecated, we should validate that an `allowed_idps.username_derivation` always provides a username claim, I believe

GeorgianaElena (commented):
Thank you very much @consideRatio! This looks very good and describing what each commit did, was very helpful ❤️

oauthenticator/cilogon.py
consideRatio: > we should validate that an allowed_idps.username_derivation always provides a username claim, I believe

The schema does this aready! `username_derivation` is required, and `username_claim` is required in it!

GeorgianaElena: Oh, totally forgot about that, sorry!

oauthenticator/cilogon.py
consideRatio: Nice catch, resolved by 69535f0!

oauthenticator/cilogon.py
consideRatio: Deal, updated in 8d1da50!

consideRatio:
Thank you soo much for reviewing @GeorgianaElena!!!!

I pushed three commits, where e7f864d was not directly related to your review comments. When documenting GenericOAuthenticator, I concluded that it wad adding functionality in username_claim that the base class didn't have, and that it duplicated things done in the base class besides checking if `username_claim` was callable. The commit refactors away the duplicated code, now referencing the base class logic.

GeorgianaElena (approved):
This is amazing @consideRatio ❤️ Thank you very much for all this work and for making the review so straightforward. This is ready to be merged IMO 🚀

consideRatio:
WIEEEEE thank you @GeorgianaElena !!!

PR #627, opened 2023-06-28
docs: a major refresher of the documentation

### LINKED ISSUE #567: Fix warnings about missing `help` strings when generating docs
When building the documentation with `make html` we get plenty of warnings about the following traitlets that have `.config(True)` which makes them configurable, but we lack a `help` string for them. This warning comes from a modern version of `autodoc_traits` to help catch this issue.

We should simply provide this!

```
/home/erik/dev/jupyterhub/autodoc-traits/autodoc_traits.py:95: FutureWarning: 
                    Documenting Auth0OAuthenticator.auth0_subdomain without a help string because it has config=True.

                    Including undocumented config=True traits is deprecated in autodoc-traits 1.1.
                    Add a help string:

                        auth0_subdomain = Unicode(
                            help="...",
                        )

                    to keep this trait in your docs,
                    or include it explicitly via :members:
```

```
Auth0OAuthenticator.auth0_domain
Auth0OAuthenticator.auth0_subdomain
Auth0OAuthenticator.client_id
Auth0OAuthenticator.client_secret
Auth0OAuthenticator.validate_server_cert
CILogonOAuthenticator.cilogon_host
GenericOAuthenticator.login_service
GitHubOAuthenticator.github_api
GitHubOAuthenticator.github_url
GitLabOAuthenticator.gitlab_api
GitLabOAuthenticator.gitlab_api_version
GitLabOAuthenticator.gitlab_url
GoogleOAuthenticator.google_api_url
GoogleOAuthenticator.user_info_url
LocalOpenShiftOAuthenticator.ca_certs
LocalOpenShiftOAuthenticator.openshift_auth_api_url
LocalOpenShiftOAuthenticator.openshift_rest_api_url
LocalOpenShiftOAuthenticator.openshift_url
```

consideRatio:
We had a lot of legacy documentation around. In this PR I've gone through a lot of the documentation and tried to make it easier to maintain long term as part of making some updates.

For example, instead of writing again how to use `GitLabOAuthenticator.allowed_project_ids`, I let the gitlab specific docs link to the configuration reference that includes a good enough help string to be useful.

old|new
-|-
image|image

## Related

- Fixes #567


consideRatio:
If you have time @GeorgianaElena, I'd love a review of this as well. Thank you soo much for your help reviewing already in other PRs!!!

consideRatio:
Argh sorry @GeorgianaElena, I requested a review a bit too early and pushed two commits with additional things =/

There is a docs build failure now, and it made me want to cover a bit more content while I'm at it. I'll comment again when its done.

consideRatio:
Pweh! Okay @GeorgianaElena sorry for the noise, I did a final self-review of everything and pushed some details, it is now fully ready for review!

docs/source/how-to/custom-403.md
```diff
@@ -0,0 +1,27 @@
...
+```
+
+You can configure a custom error message like this:
+
+```python
+c.OAuthenticator.custom_403_message = "Your message for the user"
+```
+
+You can also show a customized 403 HTML page by creating a [custom HTML
+template], and configuring JupyterHub to find it.
```
GeorgianaElena, line 17: Or is this supposed to be pointing to a link?
```suggestion
You can also show a customized 403 HTML page by creating a custom HTML
template, and configuring JupyterHub to find it.
```

GeorgianaElena: I just realized you put the links at the end. Hmm ,I think it's nice they're separated. I like it!

oauthenticator/gitlab.py
```diff
@@ -88,27 +80,69 @@ def _authorize_url_default(self):
     def _token_url_default(self):
         return f"{self.gitlab_url}/oauth/token"
 
+    gitlab_api_version = CUnicode(
```
GeorgianaElena, line 83: Small typo:

```suggestion
    gitlab_api_version = Unicode(
```

oauthenticator/openshift.py
```diff
@@ -13,11 +13,15 @@
...
 
-    scope = ['user:info']
+    @default("scope")
+    def _scope_default(self):
+        return ["user:info"]
 
-    user_auth_state_key = "openshift_user"
+    @default("login_service")
+    def _login_service_default(self):
+        return os.environ.get("LOGIN_SERVICE", "OpenShift")
```
GeorgianaElena, line 24: Wondering why a dynamic default here? Do we wish this to be configurable?

oauthenticator/openshift.py
```diff
@@ -27,33 +31,65 @@ def _username_claim_default(self):
...
+        certificates from a specific CA.
+        """,
     )
 
-    admin_groups = Set(
+    # FIXME: validate_cert is defined here, but OAuthenticator also defines
+    #        validate_server_cert. If both should exist separately its too
+    #        confusing without further documentation, and if only one should
+    #        exist the one here should be deprecated in favor of the other.
+    #
```
GeorgianaElena, line 70: To be honest, I think this should be deprecated in favour of the one from the base class. It might just be a leftover from the refactor

GeorgianaElena (approved):
This is AMAZING work @consideRatio ❤️ 🎉 Love how the documentation is also now consistent across providers 🚀

oauthenticator/gitlab.py
consideRatio: It isn't a change in this PR to add the `C`, but I'm not sure about the type. I think its related to converting from one type to another.

Looking at https://traitlets.readthedocs.io/en/stable/trait_types.html#traitlets.CUnicode, its not so clear:

> Casting variants. When a value is assigned to the attribute, these will attempt to convert it to their type. They will not automatically encode/decode between unicode and bytes, however.

What does "convert it to their type" really mean? I figure it becomes an int if the value assigned is a string `"4"`?

oauthenticator/openshift.py
consideRatio: It is configurable in Authenticator, then it was overridden to a string constant in OAuthenticator, and then a few but not all classes made it configurable again.

I'd love to deprecate a few environment variables used to configure things, but I figure its out of scope for this PR trying to be documentation focused.

oauthenticator/openshift.py
consideRatio: Ah I suspected it could be!

I have another smaller PR I plan to open if this resolves related to deprecated config, I can have a look after

consideRatio:
Thank you @GeorgianaElena for your help to review this so quickly!!!

oauthenticator/openshift.py
consideRatio: Ah, so CUnicode is something casted to a string. So, giving it the intger will cast it to a string. So instead of `Unicode` erroring if its set with `4`, it would instead work out and have it casted to `"4"`.

consideRatio: There is also `CInt` for example, which then would cast to an Int.

consideRatio:
Working on future PRs I realized we needed `default_role` in docs/source/conf.py set to `literal` still, which makes the rST parsed helpstrings like \`code?\` format as `code?` instead of formatting as _code?_.

I pushed a commit to change that setting, and another to update docs/source/conf.py a bit to look more like the one in jupyterhub-python-repo-template.

oauthenticator/gitlab.py
consideRatio: Figured out in https://github.com/jupyterhub/oauthenticator/pull/627#discussion_r1246840709, this is as it should be!

consideRatio:
There were notable differences in the .py files docstrings, where auth0 for example included a lot of outdated documentation and cilogon included caveats that I figure can be removed. I pushed another commit making them consistent, and making this view look consistent as well as a consequence:

image

@GeorgianaElena sorry for appending this PR with more and more after pinging for review. I'll stop now and open additional PRs instead with additional changes.

consideRatio:
I didn't realize you approved this PR before Georgiana! I'll go for a self-merge to enable other PRs on top of this to be easier to review.

Thank you soo much for your help reviewing @GeorgianaElena!!!

PR #628, opened 2023-06-30
refactor: separate deprecated config for readability

consideRatio:
Some classes had quite a bit of no longer relevant config, and since they also had long help strings it was a bit tricky to overview effectively what was relevant. With this PR I've:

- relocated `_deprecated_oauth_aliases` list of deprecated config and the deprecated/removed config to the end of config declarations
- reduced help strings of deprecated/removed config to ``"[Deprecated|Removed], use :attr:`.<class>.<new_config>`"`` strings

## Code refactoring preview

image

## Config reference docs preview

image

### DIFF
--- a/oauthenticator/bitbucket.py
+++ b/oauthenticator/bitbucket.py
@@ -45,7 +45,7 @@ class BitbucketOAuthenticator(OAuthenticator):
     }
     team_whitelist = Set(
         config=True,
-        help="Deprecated, use :attr:`.BitbucketOAuthenticator.allowed_teams`.",
+        help="Deprecated in 0.12, use :attr:`.BitbucketOAuthenticator.allowed_teams`.",
     )
 
     async def _fetch_user_teams(self, access_token, token_type):
--- a/oauthenticator/bitbucket.py
+++ b/oauthenticator/bitbucket.py
@@ -48,7 +48,7 @@ class BitbucketOAuthenticator(OAuthenticator):
         help="""
         .. deprecated:: 0.12
 
-           Use :attr:`.BitbucketOAuthenticator.allowed_teams`.
+           Use :attr:`allowed_teams`.
         """,
     )
 
--- a/oauthenticator/bitbucket.py
+++ b/oauthenticator/bitbucket.py
@@ -11,11 +11,6 @@ from .oauth2 import OAuthenticator
 
 
 class BitbucketOAuthenticator(OAuthenticator):
-    _deprecated_oauth_aliases = {
-        "team_whitelist": ("allowed_teams", "0.12.0"),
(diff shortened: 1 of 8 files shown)


oauthenticator/bitbucket.py
```diff
@@ -36,18 +31,21 @@ def _token_url_default(self):
...
+    _deprecated_oauth_aliases = {
+        "team_whitelist": ("allowed_teams", "0.12.0"),
+        **OAuthenticator._deprecated_oauth_aliases,
+    }
+    team_whitelist = Set(
         config=True,
-        help="""
-        Allow members of selected Bitbucket teams to sign in.
-        """,
+        help="Deprecated, use :attr:`.BitbucketOAuthenticator.allowed_teams`.",
```
minrk, line 48: I've learned that it's a good practice to always have `in VERSION` everywhere you have a 'deprecated' or 'removed' message, so that folks who see it know how to coordinate the deprecations with their version requirements without having to dig around and find the relevant version where things changed.

minrk (commented):
LGTM, one comment about making sure version numbers are in deprecation messages

consideRatio:
Thanks for reviewing @minrk! I was thinking "should I write "Deprecated in 0.12" or "Deprecated in v0.12" and searched for examples in jupyterhub/jupyterhub, where I found the following:

image
image

- Should we use the `.. deprecated:: <version>` admonition?
- Should we write a prefix `v` in this context?

minrk:
> we should use the .. deprecated:: <version> admonition?

Yeah, since it gets rendered in docs, that makes sense.

> if so, what to do about the "Removed in .."?

Hm. A few options - you could use the deprecated admonition and put "this has been removed" in the _body_, like:

```
.. deprecated:: 1.2
    Using this option will result in an error, ...
```

or use `versionchanged` with the same explanation.

or even try [this extension](https://github.com/MrSenko/sphinx-removed-in) so we could use `.. versionremoved::`.

> we should write a prefix v in this context?

No, `v` is not part of the version, it's an abbreviation used some places (e.g. in a git tag) for the word 'version' to put _next to_ the actual version number. So `version v16` would be saying 'version version 16'.

consideRatio:
Thanks @minrk! I've opened https://github.com/sphinx-doc/sphinx/issues/11480 about `.. versionremoved` btw, it seems reasonable for the sphinx project to consider adding that to complement the other directives to capture the full lifecycle of things.

minrk:
Since it's simple like that, I think it's fine. Good find! It's probably not that often that we'll have attributes that are still present and documented but 'removed'.

minrk (approved):

consideRatio:
Thanks for reviewing @minrk!!!

PR #637, opened 2023-07-03
docs: coalesce v16 upgrade page into changelog, improve helpstrings

consideRatio:
- breaking: add allow_existing_users config
- Avoid building a big set of allowed_users if allow_all is configured
- docs: add warning for allow_existing_users about removing users
- docs: let breaking changes referenced config be links to config ref
- docs: relocate breaking change descriptions to changelog
- docs: update helpstrings for oauth urls etc
- docs: remove outdated note about now a constant would be configurable
- docs: coalesce v16 upgrade notes to changelog

PR #640, opened 2023-07-04
[OpenShift] Remove ca_certs, deprecate validate_cert, fix unreleased regression

### LINKED ISSUE #639: [OpenShift] Fix no longer respected ca_certs config
In https://github.com/jupyterhub/oauthenticator/pull/363, the OpenShiftOAuthenticator got introduced with a `ca_certs` configuration that was used as an additional argument to the client making web requests when authenticating. After unreleased refactoring in #526, this config is no longer respected.

Fixing this can be done in a few ways I'm considering:

1. (breaking) We declare that `ca_certs` should be specified using `http_request_kwargs` (from #578 by @manics)
2. (broken) We declare its no longer supported
3. (maintenance need escalation) We lift `ca_certs` as a config to OAuthenticator from openshift

I'm leaning towards a breaking change, where we say it should be configured using `http_request_kwargs` instead.

@manics and @GeorgianaElena, what do you think?

consideRatio:
Based on top of #637. This PR includes only the last commit 4bfb274.

- Fixes #639


docs/source/reference/changelog.md
```diff
@@ -29,8 +30,39 @@ command line for details.
...
 - [Google] Authentication state's `google_groups` is now a set, not a list.
 - [CILogon] {attr}`.CILogonOAuthenticator.allowed_idps` is now required config,
-  and `shown_idps`, `username_claim`, `additional_username_claims` must no
-  longer be configured.
+  and `shown_idps`, `username_claim`, `additional_username_claims` were removed.
+- [Okpy] The public functions `OkpyOAuthenticator.get_auth_request` and
+  `OkpyOAuthenticator.get_user_info_request` were removed.
+- [OpenShift] The config `ca_certs` was removed but
+  {attr}`.OpenShiftOAuthenticator.http_request_kwargs` can still be configured
+  with a `ca_certs` key for the same result. OpenShift's default `ca_certs`
```
manics, line 38:
```suggestion
- [OpenShift] The config `ca_certs` was removed. Use
  {attr}`.OAuthenticator.http_request_kwargs`
  with a `ca_certs` key instead. OpenShift's default `ca_certs`
```

consideRatio:
@manics thank you for reviewing!!!!!

consideRatio:
Applied code suggestions and rebased #641 on top of this

PR #644, opened 2023-07-05
Ensure login_service remain configurable

### LINKED ISSUE #643: GenericOAuthenticator.login_service no longer customizable in 16.0.0
### Bug description

Using just recently released `3.0.0-beta.2`, setting `login_service:` option in helm chart does not seem to work anymore

```
hub:
  config:
    GenericOAuthenticator:
      .... 
      login_service: Keycloak
```

The log also says `JupyterHub configurable:200] Config option login_service not recognized by GenericOAuthenticator.  Did you mean client_secret?` which is clearly incorrect as `client_secret` is secret token for OAuth2.0

#### Expected behaviour

Would have been nice if we can have the ability to customize back ...

image

#### My setup

Helm chart on Kubernetes cluster, using this version of the helm chart that was released today https://jupyterhub.github.io/helm-chart/jupyterhub-3.0.0-beta.2.tgz

consideRatio:
- Fixes #643

I didn't think it needed to be defined here as it was defined in the JupyterHub `Authenticator` class, but it was needed because it didn't have `config=True` set there.

### DIFF
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -241,6 +241,21 @@ class OAuthenticator(Authenticator):
     # dictionary that user info will be saved
     user_auth_state_key = "oauth_user"
 
+    login_service = Unicode(
+        "OAuth 2.0",
+        config=True,
+        help="""
+        Name of the login service or identity provider that this authenticator
+        is using to authenticate users.
+
+        This config influences the text on a button shown to unauthenticated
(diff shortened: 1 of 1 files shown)


consideRatio:
Going for a quick bugfix and 16.0.1 release to prepare tljh 1.0.0b1 and z2jh 3.0.0

PR #650, opened 2023-07-07
Authorization by GitHub repository access

koendelaat:
Allows for GitHub authentication based on repository access, which is required when using outside-collaborators.

Current situation only allows authorization based on org or org/team access. This requires all users to be a member of this org.

With outside-collaborators, we can grant them access to specific repositories, but they don't become member of the organization nor the teams.

This feature allow authorization based on repository access, making JupyterHub accessible to outside-collaborators.

### DIFF
--- a/oauthenticator/github.py
+++ b/oauthenticator/github.py
@@ -114,6 +114,16 @@ class GitHubOAuthenticator(OAuthenticator):
         """,
     )
 
+    allowed_repositories = Set(
+        config=True,
+        help="""
+        Allow users with read access to specified repositories by specifying
+        repositories like `org-a/repo-1`.
+
+        Requires `repo` to be set in `scope`.
+        """,
+    )
+
     populate_teams_in_auth_state = Bool(
         False,
         config=True,
@@ -182,6 +192,17 @@ class GitHubOAuthenticator(OAuthenticator):
                 ):
                     return True
             message = f"User {username} is not part of allowed_organizations"
+            self.log.warning(message)
+
+        if self.allowed_repositories:
+            access_token = auth_model["auth_state"]["token_response"]["access_token"]
+            token_type = auth_model["auth_state"]["token_response"]["token_type"]
+            for repo in self.allowed_repositories:
+                if await self._check_membership_allowed_repository(
+                    repo, username, access_token, token_type
+                ):
+                    return True
+            message = f"User {username} does not have access to allowed_repositories"
             self.log.warning(message)
 
         # users should be explicitly allowed via config, otherwise they aren't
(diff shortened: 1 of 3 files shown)


koendelaat:
@consideRatio any comments?

oauthenticator/github.py
```diff
@@ -205,23 +226,32 @@ async def update_auth_model(self, auth_model):
...
+            or "user:email" in granted_scopes
+            or not granted_scopes
         ):
-            resp_json = await self.httpfetch(
+            resp = await self.httpfetch(
                 f"{self.github_api}/user/emails",
                 "fetching user emails",
                 method="GET",
+                parse_json=False,
+                raise_error=False,
```
manics, line 245: This is silently hiding all errors. Can you instead check for only the error case you want to catch, and otherwise allow the error to propagate upwards?

oauthenticator/github.py
```diff
@@ -205,23 +226,32 @@ async def update_auth_model(self, auth_model):
...
         token_type = auth_model["auth_state"]["token_response"]["token_type"]
-        granted_scopes = auth_model["auth_state"].get("scope", [])
+        granted_scopes = [
+            scope for scope in auth_model["auth_state"].get("scope", []) if scope
+        ]
         if not user_info["email"] and (
-            "user" in granted_scopes or "user:email" in granted_scopes
+            "user" in granted_scopes
+            or "user:email" in granted_scopes
+            or not granted_scopes
```
manics, line 238: Can you explain this change? If this is called with empty scopes won't it fail to get the user's emails?

oauthenticator/github.py
```diff
@@ -114,6 +114,16 @@ def _userdata_url_default(self):
         """,
     )
 
+    allowed_repositories = Set(
+        config=True,
+        help="""
+        Allow users with read access to specified repositories by specifying
```
manics, line 120: We also need to think about future maintainability. Could you say a bit about your design decisions to make this conditional on read access as opposed to e.g. write access, making it configurable etc?

oauthenticator/github.py
koendelaat: The retrieved scopes will be empty for GitHub Apps https://docs.github.com/en/developers/apps.
So to facilitate this behavior, empty scopes are also allowed. 
This is also why we need a change in error handling of the email fetch. This might cause an error, but we don't know upfront what is causing this.

oauthenticator/github.py
koendelaat: Addressed in new commit

koendelaat:
@manics Any comments on my replies?

koendelaat:
@manics, @consideRatio Can you maybe have a look to this PR?

consideRatio (commented):
I'm super low on time december, but quick input: who to allow isnt obvious based on triager/maintainer etc status. I think the config should specify that (repo x + permission y) should have access, and not just repo x and assume a specific permission level is enough for all repos configured.

PR #655, opened 2023-07-28
Add ORCID iD Authenticator

### LINKED ISSUE #284: Add documentation for ORCID-based authentication
Almost every researcher has (or will sooner or later have) and ORCID, see https://orcid.org/

It would be useful to be able to _whitelist_ a number of ORCIDs for a JupyterHub instance, e.g. via a database, or via a file with one ORCID per line.

This could allow journals or universities, who are likely to require their authors to have an ORCID, to use authentication without needing their own secure password storage or OAuth mechanism.

matthewwiese:
This PR adds an [ORCID iD](https://orcid.org) authenticator class inheriting from `OAuthenticator`, based on the implementation of the `GitHubOAuthenticator` class.

The tests are rudimentary - I'm happy to implement suggestions from maintainers.

This ought to resolve #284


matthewwiese:
> Hi! We'd like to avoid adding more single service authenticators due to the maintenance and testing burden. Is there anything in here that can't be done by configuring the GenericAuthenticator?

Nope!

Should a code snippet using `GenericAuthenticator` be added to the docs? As it stands, #284 remains open without a usable example, and ORCID is popular enough that I imagine you'll be getting questions related to it in the future.

manics:
A documentation PR would be great!

Is it necessary to keep the upper case `X`? These are JupyterHub usernames so the requirement is a one-to-one mapping from the ORCID username to the JupyterHub username. Are ORCID usernames case sensitive?

matthewwiese:
Closing this PR.

> A documentation PR would be great!

I can do this. @manics: to what page(s) should this be added?

> Are ORCID usernames case sensitive?

At least according to my reading of [this page](https://support.orcid.org/hc/en-us/articles/360006897674-Structure-of-the-ORCID-Identifier), the capital "X" is necessary for recognizing the decimal value of 10 for the checksum.

> In general you can't rely on going from the JupyterHub username back to the original username as downstream authenticators/spawners may need to manipulate it in other ways

That's good to know, thank you. If at some point my use case needs the username to map 1-to-1 then I will use the custom authenticator, otherwise `GenericOAuthenticator` works just fine.

manics:
@matthewwiese There's a page with example configurations:
- https://oauthenticator.readthedocs.io/en/latest/tutorials/provider-specific-setup/providers/generic.html
- https://github.com/jupyterhub/oauthenticator/edit/main/tutorials/provider-specific-setup/providers/generic.md

But there's always room for improving the docs, so let us know if you think there's a better way to present this.

PR #668, opened 2023-08-16
[Google, Globus, BitBucket] Ensure auth_state is JSON serializable (lists are, not sets)

### LINKED ISSUE #667: [Google, Globus, BitBucket] Configuring `enable_auth_state` causes `set is not JSON serializable`
### Bug description

Upon upgrading to z2jh 3.0.1, a hub with google oauth emits a "500 : Internal Server Error". The logs indicate that a set object cannot be serialized to JSON.

#### Expected behaviour

I expected to be able to authenticate via Google

#### Actual behaviour

The hub emits an error.

### How to reproduce

1. A user visits the hub URL.
2. The user clicks on the button to sign in.
3. The user sees the `500 : Internal Server Error`.

### Your personal set up

Some of the configuration is at https://github.com/berkeley-dsep-infra/datahub/tree/staging/deployments/highschool.

Initially there were three hosted domains and I tried trimming them down to one, however that was not successful.

Related configuration, https://github.com/berkeley-dsep-infra/datahub/blob/staging/deployments/highschool/config/common.yaml#L23C1-L34

And secrets:
```
jupyterhub:
    hub:
        config:
            GoogleOAuthenticator:
                client_id: [snipped].googleusercontent.com
                client_secret: [snipped]
                oauth_callback_url: https://highschool-staging.datahub.berkeley.edu/hub/oauth_callback
            Authenticator:
                admin_users:
                    # infrastructure
                    - user1@berkeley.edu
                    - user2@berkeley.edu
```

We also have:
```
      Authenticator:
        enable_auth_state: true
        manage_groups: true
      OAuthenticator:
        allow_all: true
```

I also tried disabling `manage_groups`.

The error occurs even without spawning, i.e. if the user tries to go to /hub/admin or /hub/home, where the latter doesn't start a server.

```
[I 2023-08-15 23:19:14.910 JupyterHub oauth2:102] OAuth redirect: https://our_hub/hub/oauth_callback
[I 2023-08-15 23:19:14.911 JupyterHub log:191] 302 GET /hub/oauth_login?next=%2Fhub%2F -> https://accounts.google.com/o/oauth2/v2/auth?response_type=code&redirect_uri=https%3A%2F%2Four_hub%2Fhub%2Foauth_callback&client_id=_snipped_.apps.googleusercontent.com&state=[secret]&scope=openid+email (@10.0.0.37) 2.06ms
[E 2023-08-15 23:19:15.521 JupyterHub web:1871] Uncaught exception GET /hub/oauth_callback?state=_snipped_&code=_snipped_&scope=email+https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fuserinfo.email+openid&authuser=0&hd=our.domain&prompt=none (10.0.0.37)
    HTTPServerRequest(protocol='https', host='our_hub', method='GET', uri='/hub/oauth_callback?state=_snipped_&code=_snipped_&scope=email+https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fuserinfo.email+openid&authuser=0&hd=our.domain&prompt=none', version='HTTP/1.1', remote_ip='10.0.0.37')
      File "/usr/local/lib/python3.11/site-packages/tornado/web.py", line 1786, in _execute
        result = await result
                 ^^^^^^^^^^^^
      File "/usr/local/lib/python3.11/site-packages/oauthenticator/oauth2.py", line 202, in get
               ^^^^^^^^^^^^^^^^^^^^^^^
      File "/usr/local/lib/python3.11/site-packages/jupyterhub/handlers/base.py", line 832, in login_user
        user = await self.auth_to_user(authenticated)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/usr/local/lib/python3.11/site-packages/jupyterhub/handlers/base.py", line 821, in auth_to_user
        await user.save_auth_state(auth_state)
      File "/usr/local/lib/python3.11/site-packages/jupyterhub/user.py", line 344, in save_auth_state
        self.encrypted_auth_state = await encrypt(auth_state)
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/usr/local/lib/python3.11/concurrent/futures/thread.py", line 58, in run
        result = self.fn(*self.args, **self.kwargs)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/usr/local/lib/python3.11/site-packages/jupyterhub/crypto.py", line 141, in _encrypt
        return self.fernet.encrypt(json.dumps(data).encode('utf8'))
                                   ^^^^^^^^^^^^^^^^
      File "/usr/local/lib/python3.11/json/__init__.py", line 231, in dumps
        return _default_encoder.encode(obj)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/usr/local/lib/python3.11/json/encoder.py", line 200, in encode
        chunks = self.iterencode(o, _one_shot=True)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/usr/local/lib/python3.11/json/encoder.py", line 258, in iterencode
        return _iterencode(o, 0)
               ^^^^^^^^^^^^^^^^^
      File "/usr/local/lib/python3.11/json/encoder.py", line 180, in default
        raise TypeError(f'Object of type {o.__class__.__name__} '
    TypeError: Object of type set is not JSON serializable
```

I see from the changelog that some objects were converted to Sets, but I can't tell which item is being serialized here and if its something I need to coax into a list.

consideRatio:
This PR adds a test to ensure the `auth_state` is set to a serializable object, because if it isn't, then use of `c.Authenticator.enable_auth_state = True` will cause JupyterHub to error with a broken assumption.

This PR doesn't resolve the situation. The situation is caused by a change for various group related fields from list to set in the 16.0.0 release. The change was meant to reduce back and forth casting and confusion stemming from config being Set based but info about groups/teams ended up list based.

## Fix implementation brainstorming

I think there are various strategies to consider to resolve this caused by putting `set` objects under `auth_state`.

1. To always cast back and forth between set/list when writing/reading from `auth_state`
2. To change config of Set type to List type if that relates to avoiding all set based `auth_state` without needing to cast back and forth
3. To not put set based content under `auth_state`, but in `auth_model` (then it won't be persisted)

I'm leaning on going for 1 or 2. To evaluate if 2 is an alternative, we should know if the traitlet based List type can accept being passed at set also. Probably towards 1?

## Action points - help wanted

I'm on vacation so if someone wants to resolve this by completing this PR by pushing commits or opening another one, please go for it! Make a comment that you are in case I'll end up wanting to work this during my vacation.

- [x] Fix failing tests for BitBucket
- [x] Fix failing tests for Globus
- [x] Fix failing tests for Google

closes #667

### DIFF
--- a/oauthenticator/bitbucket.py
+++ b/oauthenticator/bitbucket.py
@@ -93,7 +93,7 @@ class BitbucketOAuthenticator(OAuthenticator):
 
         if self.allowed_teams:
             user_teams = set(auth_model["auth_state"].get("user_teams", []))
-            if any(user_teams & self.allowed_teams):
+            if user_teams & self.allowed_teams:
                 return True
 
         # users should be explicitly allowed via config, otherwise they aren't
--- a/oauthenticator/bitbucket.py
+++ b/oauthenticator/bitbucket.py
@@ -73,11 +73,13 @@ class BitbucketOAuthenticator(OAuthenticator):
         Fetch and store `user_teams` in auth state if `allowed_teams` is
         configured.
         """
+        user_teams = set()
         if self.allowed_teams:
             access_token = auth_model["auth_state"]["token_response"]["access_token"]
             token_type = auth_model["auth_state"]["token_response"]["token_type"]
             user_teams = await self._fetch_user_teams(access_token, token_type)
-            auth_model["auth_state"]["user_teams"] = user_teams
+        # sets are not JSONable, cast to list for auth_state
+        auth_model["auth_state"]["user_teams"] = list(user_teams)
 
         return auth_model
 
@@ -90,7 +92,7 @@ class BitbucketOAuthenticator(OAuthenticator):
             return True
 
         if self.allowed_teams:
-            user_teams = auth_model["auth_state"]["user_teams"]
+            user_teams = set(auth_model["auth_state"].get("user_teams", []))
             if any(user_teams & self.allowed_teams):
                 return True
 
--- a/oauthenticator/generic.py
+++ b/oauthenticator/generic.py
@@ -159,7 +159,7 @@ class GenericOAuthenticator(OAuthenticator):
             # admin status should in this case be True or False, not None
             user_info = auth_model["auth_state"][self.user_auth_state_key]
             user_groups = self.get_user_groups(user_info)
-            auth_model["admin"] = any(user_groups & self.admin_groups)
+            auth_model["admin"] = bool(user_groups & self.admin_groups)
 
         return auth_model
 
--- a/oauthenticator/globus.py
+++ b/oauthenticator/globus.py
@@ -314,7 +314,7 @@ class GlobusOAuthenticator(OAuthenticator):
 
         if self.allowed_globus_groups:
             user_groups = set(auth_model["auth_state"]["globus_groups"])
-            if any(user_groups & self.allowed_globus_groups):
+            if user_groups & self.allowed_globus_groups:
                 return True
             self.log.warning(f"{username} not in an allowed Globus Group")
 
@@ -344,7 +344,7 @@ class GlobusOAuthenticator(OAuthenticator):
 
         if self.admin_globus_groups:
             # admin status should in this case be True or False, not None
-            auth_model["admin"] = any(user_groups & self.admin_globus_groups)
+            auth_model["admin"] = bool(user_groups & self.admin_globus_groups)
 
         return auth_model
 
--- a/oauthenticator/globus.py
+++ b/oauthenticator/globus.py
@@ -313,7 +313,7 @@ class GlobusOAuthenticator(OAuthenticator):
             return True
 
         if self.allowed_globus_groups:
-            user_groups = auth_model["auth_state"]["globus_groups"]
+            user_groups = set(auth_model["auth_state"]["globus_groups"])
             if any(user_groups & self.allowed_globus_groups):
                 return True
             self.log.warning(f"{username} not in an allowed Globus Group")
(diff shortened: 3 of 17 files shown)


minrk:
Looking at this now. I think we should go with 1.

minrk:
>  To evaluate if 2 is an alternative, we should know if the traitlet based List type can accept being passed at set also. Probably towards 1?

Yes, set traits accept lists as input and cast them. But I think it's still the right thing to do for the config to be a set.

I think casting to list to/from auth_state is a simple and straightforward change. The other simplest change would be to leave it a list everywhere use the set config as the left hand side and pass the list, because `set.intersection` accepts lists (any iterable, in fact):

```python
In [2]: {'a', 'b'}.intersection(['b', 'c', 'd'])
Out[2]: {'b'}
```

I've implemented 1. and fixed some logic in the computation of checking for intersections, because:

```python
if any(set_a & set_b):
```

is not the same as

```python
if set_a & set_b:
```

which is what we want. These checks are meant to look for a non-empty intersection (i.e. truthy set, aka set with length > 0), whereas `any(set_a & set_b)` checks if any item _in_ the intersection is truthy. This isn't a serious bug because the only falsy string (`''`) doesn't make sense as a group name, but the principle of the logic is not right.

The only situation where the answers would differ is a set containing only the empty string, i.e.:

```python
In [3]: any({'', 'a'} & {''})
Out[3]: False

In [4]: bool({'', 'a'} & {''})
Out[4]: True
```

which can't really come up in practice.

minrk:
I removed 'breaking' because I don't see any breaking changes here. Only a fix to allow auth_state to work. The type of an auth_state field is changed, but that's not an API to be consumed, and auth_state doesn't work at all until this PR because it's not persisted.

consideRatio:
Thank you @minrk for working this!!!

I think its breaking by reverting a very small breaking change in 16.0.0. I consider it breaking as one can have have logic in a post_auth_hook that make an assumption about the type we stored the group/team as. I don't suggest a new major release, but that we merge this and make a notice in the 16.0.0 changelog that we reverted the change of making the auth_state store groups/teams etc as sets. Apparently that is something we only wrote about for Google, not also Globus and BitBucket.

image

minrk:
I think that's fair, but I also think there shouldn't be an expectation of using auth_state when it doesn't work with auth_state enabled. I think "auth_state didn't work until now" invalidates expectations of the structure of auth_state while it didn't work.

consideRatio:
Having reviewed Mins work on this, I'll go for a merge!

ryanlovett:
Thanks @minrk and @consideRatio !

PR #680, opened 2023-09-13
docs: fix `"generic"` renamed to `"generic-oauth"`

mehalter:
I noticed that at some point `"generic"` was changed to `"generic-oauth"` so this just updates the docs to reflect this. This is my first contribution, so let me know if there is anything else I should do to get this merged in.

Thanks for an awesome project!

### DIFF
--- a/docs/source/how-to/writing-an-oauthenticator.md
+++ b/docs/source/how-to/writing-an-oauthenticator.md
@@ -21,7 +21,7 @@ and configuration to set the necessary configuration variables.
 Example config:
 
 ```python
-c.JupyterHub.authenticator_class = "generic"
+c.JupyterHub.authenticator_class = "generic-oauth"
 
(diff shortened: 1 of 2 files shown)


consideRatio:
Oh great catch, if this is a change of behavior that introduced over time, I think its to be considered a regression / bug that we should have fixed.

Since users may have realized this and perhaps started writing `"generic-oauth"`, I think it would be fine if either would be made to work as well.

Ah, it seems that `"generic"` has never worked based on looking at https://github.com/jupyterhub/oauthenticator/blame/main/setup.py#L52-L53, so it has been a documentation regression only. Then this makes perfect sense to merge as it is!

consideRatio:
Thank you @mehalter!!! :heart: :tada: :sunflower:

Contratulations to becoming a code contributor to the jupyterhub ecosystem of software!

PR #684, opened 2023-09-23
[CILogon] Add allow_all as a idp specific config

### LINKED ISSUE #682: [CILogon] How to authorize all users of one idp but not another - add idp specific `allow_all` config?
Using CILogonOAuthenticator, you configure access to one or more identity providers (idps). That makes it an odd authenticator class among the other classes in this project.

In 15.1.0, we could have idp config like below:

```yaml
      CILogonOAuthenticator:
        allowed_idps:
          # google accounts
          http://google.com/accounts/o8/id:
            username_derivation:
              username_claim: "email"
            allowed_domains: ["google-allowed.org"]

          # microsoft accounts
          http://login.microsoftonline.com/common/oauth2/v2.0/authorize:
            username_derivation:
              username_claim: "email"
```

Assuming the goal was to allow all microsoft accounts, but only the google accounts with an email domain of "google-allowed.org", then this config in oauth v15 made it happen.

In v16, this is not something we can configure any more. While access would be granted correctly to google accounts with domains explicitly allowed, oauth v16 would no longer allow the microsoft accounts. For other oauthenticators this was resolved by adding the `allow_all` config to explicitly allow all users. In this case though, configuring `allow_all` would allow both google users and microsoft users, even though we only wanted to allow all microsoft users.

## Proposal

- We add an `allow_all` option for individual IDPs as well.

consideRatio:
- Fixes #682 by adding a `allow_all` sub-config next to `allowed_domains`.

Without this, a user may be forced to use OAUthenticator v15 to acquire a functionality where all users authenticated by an IDP are allowed, but not all users of another IDP. So in practice, this mitigates a feature regression - where something was possible in oauthenticator 15 that no longer was possible in oauthenticator 16.

Here is an example config where we allow all users from one IDP together with one github user being an admin user.

```python
c.CILogonOAuthenticator.allowed_idps = {
    "https://idpz.utorauth.utoronto.ca/shibboleth": {
        "username_derivation": {
            "username_claim": "email",
        },
        "allow_all": True,
    },
    "https://github.com/login/oauth/authorize": {
        "username_derivation": {
            "username_claim": "preferred_username",
        },
    },
}
c.Authenticator.admin_users = ["some-github-user1"]
```

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -6,7 +6,6 @@
 autodoc-traits
 importlib_metadata>=3.6; python_version < '3.10'
 myst-parser
-sphinx>=2
 sphinx-autobuild
 sphinx-book-theme
 sphinx-copybutton
--- a/oauthenticator/cilogon.py
+++ b/oauthenticator/cilogon.py
@@ -123,28 +123,36 @@ class CILogonOAuthenticator(OAuthenticator):
                         "action": "strip_idp_domain",
                         "domain": "utoronto.ca",
                     },
-                },
-                "https://github.com/login/oauth/authorize": {
-                    "username_derivation": {
-                        "username_claim": "username",
-                        "action": "prefix",
-                        "prefix": "github",
-                    },
+                    "allow_all": True,
                 },
                 "http://google.com/accounts/o8/id": {
                     "username_derivation": {
-                        "username_claim": "username",
+                        "username_claim": "email",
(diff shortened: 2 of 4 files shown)


GeorgianaElena (approved):
Thanks for fixing this @consideRatio!

consideRatio:
Thanks for reviewing @GeorgianaElena!!!

PR #685, opened 2023-09-25
[CILogon] Rename allowed_idps to idps (deprecation, not removal)

### LINKED ISSUE #683: [CILogon] Rename `allowed_idps` to `idps` in a non-breaking way?
I think the name `allowed_idps` causes some confusion in oauthenticator v16, and that it should be non-breakingly renamed to just `idps` with a deprecation warning for users still using the `allowed_idps` name.

With a name like `idps`, there is no hint about users from a specific ipd would be allowed or simiarly. Configuring idps is about describing what users we can authenticate/recognize, while `allowed` refers to what users we authorize/allow. Since configuring an idp to be used to authenticate a user doesn't go hand in hand with authorizing a user, I think it should be renamed for clarity.

Together with the proposal in #682, the config for CILogon that is associated with authorizing users then become the following:

- `OAuthenticator.allow_all`
- `OAuthenticator.allowed_users`
- `OAuthenticator.admin_users`
- `OAuthenticator.allow_existing_users`
- `CILogonOAuthenticator.idps[<some idp>].allowed_domains`
- `CILogonOAuthenticator.idps[<some idp>].allow_all`

## Proposal

Rename `allowed_idps` to `idps`, making `allowed_idps` still work as an alias for `idps`, but come with a deprecation warning.

consideRatio:
- Fixes #683

Note that the tests passes for the first commit that runs tests against the old name, but also for the last commit that runs tests against the new name. Like this, I feel quite confident we haven't messed up our deprecation logic, and that makes me confident enough to suggest this name change.

image

### DIFF
--- a/docs/source/tutorials/provider-specific-setup/providers/cilogon.md
+++ b/docs/source/tutorials/provider-specific-setup/providers/cilogon.md
@@ -21,4 +21,4 @@ c.OAuthenticator.client_secret = "[your oauth2 application secret]"
 CILogonOAuthenticator expands OAuthenticator with the following required config,
 read more about it in the configuration reference:
 
-- {attr}`.CILogonOAuthenticator.allowed_idps`
+- {attr}`.CILogonOAuthenticator.idps`
--- a/oauthenticator/cilogon.py
+++ b/oauthenticator/cilogon.py
@@ -17,7 +17,7 @@ from .oauth2 import OAuthenticator, OAuthLoginHandler
 yaml = YAML(typ="safe", pure=True)
 
 
-def _get_select_idp_param(allowed_idps):
+def _get_select_idp_param(idps):
     """
     The "selected_idp" query parameter included when the user is redirected to
     CILogon should be a comma separated string of idps to choose from, where the
@@ -25,11 +25,11 @@ def _get_select_idp_param(allowed_idps):
     remaining idps has no meaning.
     """
     # pick the first idp marked as default, or fallback to the first idp
-    default_keys = [k for k, v in allowed_idps.items() if v.get("default")]
-    default_key = next(iter(default_keys), next(iter(allowed_idps)))
+    default_keys = [k for k, v in idps.items() if v.get("default")]
+    default_key = next(iter(default_keys), next(iter(idps)))
 
     # put the default idp first followed by the other idps
-    other_keys = [k for k, _ in allowed_idps.items() if k != default_key]
+    other_keys = [k for k, _ in idps.items() if k != default_key]
     selected_idp = ",".join([default_key] + other_keys)
 
     return selected_idp
@@ -41,7 +41,7 @@ class CILogonLoginHandler(OAuthLoginHandler):
     def authorize_redirect(self, *args, **kwargs):
         """
         Optionally add "skin" to redirect params, and always add "selected_idp"
-        (aka. "idphint") based on allowed_idps config.
+        (aka. "idphint") based on idps config.
 
         Related documentation at https://www.cilogon.org/oidc#h.p_IWGvXH0okDI_.
         """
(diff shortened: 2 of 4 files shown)


GeorgianaElena (commented):
Personally, I believe the original `allowed_idps` makes more sense than `idps`.

> Configuring idps is about describing what users we can authenticate/recognize, while allowed refers to what users we authorize/allow.

I believe that by configuring the `idps` dict (i.e. `allowed_idps`) we are not describing what users we can authenticate, but what users we authorize.

By default, CILogon allows users to authenticate using any idp from the entire list they support at https://cilogon.org/idplist/. If you go to https://cilogon.org/ you are presented with the entire list of idps you can use to **authenticate**.

But by configuring the `allowed_ips` dict, we are actually choosing to authorize **only** the users that use a specific idp to login. And we do this here:
https://github.com/jupyterhub/oauthenticator/blob/4bad3e819358a217d118109d9ae1d0d50fe98652/oauthenticator/cilogon.py#L32-L34
where we choose to only show the `allowed_idps` as authentication options.

And then at the lines below if any other idp is used
 https://github.com/jupyterhub/oauthenticator/blob/4bad3e819358a217d118109d9ae1d0d50fe98652/oauthenticator/cilogon.py#L302-L305.

consideRatio:
Ah think I understand what you are saying!

I agree that relating to CILogons default, we are saying what subset of idps we can use to authenticate users with - and that is a denial of authornization. I was using a narrower meaning of authorization, thinking about granting or not granting access to an already authenticated user.

---

I agree in how it can make sense to call this config `allowed_idps`, and there is some value to describing that only some idps can be used, but I'd like to champion use of `idps` further as I think the naming `allowed_idps` have too large drawbacks:

- Reading `allowed_idps` next to other config involving `alllow`, it seems ambgious what it does. I think the other `allow` config hints that it could mean the first thing below, but in practice it means the latter:
  - "allow all users authenticated via this IdP"
  - "only these idp among all cilogons supported idps are allowed to authenticate users"
- The `allowed_idps` behavior is different from `allowed_groups` in the sense that its sufficient to be part of a group in `allowed_groups`, but its not sufficient to be a user authenticated with an `allowed_idps` - the latter is just one initial requirement.

---

Maybe we could have a voice chat about this? I've found myself struggling to effectively discuss various topics by writing comments, and as this comment took me ~45-60 minutes to write it could be very helpful for me to transition to a voice chat!

GeorgianaElena:
> Maybe we could have a voice chat about this? I've found myself struggling to effectively discuss various topics by writing comments, and as this comment took me ~45-60 minutes to write it could be very helpful for me to transition to a voice chat!

Sorry for missing this message @consideRatio :( !Let me share my thoughts super quickly below and let's sync to have a chat if we still don't manage to converge.

Thank you for detailing why you are advocating to remove the `allowed` prefix from the name, given the existing context of other `allow` prefixed config. I understand and agree with you that it might cause confusion about what's its assumed meaning given the context vs what it actually does which is to only *permit* or to configure a list of *supported* IDPs for login.

In this spirit, what do you think about renaming it to `permitted_idps` or `supported_idps`, or use other prefix that implies what it does? In this way the *assumed* understanding about what this config does is more close to the truth, without being influenced by the assumed `allow` prefix meaning?

GeorgianaElena (approved):
I came across the name confusion that `allowed_idps` configuration generates, while trying to explain how the `allow_` prefixed config works.

I believe this PR will definitely make things clearer!

Thank you @consideRatio!

consideRatio:
> I came across the name confusion that `allowed_idps` configuration generates, while trying to explain how the `allow_` prefixed config works.

Me too! :D

Thank you for taking the time to think about this @GeorgianaElena, and I'm sorry for not following up on your response earlier in any way - it's always nice to have some kind of confirmation!

I resolved some issues stemming from being an old PR outdated with the main branch etc. and force-pushed a rebase. I really appreciate how thorough the test suite is in this project - it helped me now - thank you for your maintenance effort @GeorgianaElena!!!

PR #694, opened 2023-10-16
openshift: fix fetching of default openshift_auth_api_url

consideRatio:


oauthenticator/openshift.py
```diff
@@ -81,15 +82,25 @@ def _http_request_kwargs_default(self):
...
+        # https://github.com/tornadoweb/tornado/issues/2325#issuecomment-375972739.
+        #
+        def fetch_auth_info():
+            client = HTTPClient()
+            req = HTTPRequest(auth_info_url, **self.http_request_kwargs)
+            resp = client.fetch(req)
+            resp_json = json.loads(resp.body.decode("utf8", "replace"))
+            return resp_json
+
+        with concurrent.futures.ThreadPoolExecutor() as executor:
```
minrk, line 100: When using it for a single thread, specify `1` as the number of threads, as it _could_ create up to n_cpus threads that aren't used (it's currently smarter than that and only allocates threads on demand, but I don't think this has always been the case).

```suggestion
        with concurrent.futures.ThreadPoolExecutor(1) as executor:
```

minrk (approved):
LGTM!

manics (approved):

PR #695, opened 2023-10-18
Add changelog for 16.1.1

consideRatio:
Gets a release out as the OpenShift oauthenticator was broken and is now fixed.

Note that this includes a removal of Okpy OAuthenticator, which would be breaking if someone relied on it still - but we believe nobody does. Due to that, we include that in this patch release. The changelog includes a note about this and guidance for users ok the okpyoauthenticator to migrate if there were any - which we believe there isn't.

image

### DIFF
--- a/docs/source/reference/changelog.md
+++ b/docs/source/reference/changelog.md
@@ -8,6 +8,39 @@ command line for details.
 
 ## 16.1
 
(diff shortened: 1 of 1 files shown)


consideRatio:
Thank you @GeorgianaElena! A release workflow is now triggered

Issue #677, opened 2023-08-23
[Bitbucket] Rework required


consideRatio:
#676 reported an issue, but it untangled to multiple issues that needed to be resolved so I've now opened this issue to summarize the findings. I think a proper rework of BitBucketOAuthenticator is required.

- [ ] `allowed_teams` should be deprecated and renamed `allowed_workspaces`
- [ ] (security) `allowed_teams` works with workspace's `name` fields (display name) that isn't unique, not their `uuid` or `slug` (workspace id) fields
  The functionality of checking workspace membership is currently broken though
- [x] When constructing an Authorization headers for web requests we must send requests with `Bearer` and it mustn't be `bearer` as it is if we reference a previous response's `token_type` value -> https://github.com/jupyterhub/oauthenticator/pull/698
- [ ] When using BitBuchetOAuthenticator, the default scopes isn't sufficient I think.
  I know we make requests to endpoints that at least requires the `account` scope, but perhaps not email.

jyio:
I recently ran into the `Authorization: Bearer` issue while trying to integrate GenericOAuthenticator with Kanidm, and we'd likely hit the same snag with any OAuth2 IdP that doesn't follow Postel's law.

RFC 6749 says [`token_type` is case-insensitive in the access token response](https://www.rfc-editor.org/rfc/rfc6749#section-5.1). RFC 6750 says [`Bearer` should be capitalized in the `Authorization` request header](https://www.rfc-editor.org/rfc/rfc6750#section-2.1). This has caused [many](https://www.ibm.com/support/pages/apar/PI11907) [headaches](https://sgeb.io/posts/fix-go-oauth2-case-sensitive-bearer-auth-headers/) related to inconsistent case usage. Bitbucket and Kanidm are technically RFC-compliant by sending `bearer` and expecting `Bearer`; but I believe it would be more consistent and just as compliant to send `Bearer`. On the other hand, OAuthenticator is non-compliant by receiving `bearer` and sending `bearer`; and the fix is likely simple:

**Option 1:** Hard-code `Bearer`, [as recommended by this user](https://github.com/jupyterhub/oauthenticator/issues/225) [and you](https://github.com/jupyterhub/oauthenticator/issues/676#issuecomment-1690057661). It would work for most cases, but miss the unusual IdP that wants to use another scheme such as Basic, Digest, or HOBA.

[...]

Issue #676, opened 2023-08-22
Bitbucket auth broken


huwf:
### Bug description

I think this might be a bug in Bitbucket rather than OAuthenticator, but I'm not sure, and it may be specific to my setup. The requests to get user information and workspace permissions get a 401 response, handled as a 500 in Jupyterhub when using BitbucketOAuthenticator. 

This appears to be a problem with the way the `Authorization` header is being sent/handled, because including the token in the querystring appears to work. Setting `OAUTH2_USERDATA_REQUEST_TYPE=url` in the environment makes it work for `token_to_user` but `_fetch_user_teams` still breaks

#### Expected behaviour

Setting `Authorization` header should successfully authenticate with bitbucket cloud

#### Actual behaviour

401 status returned

### How to reproduce

1. Setup OAuthentiator as the setup below
2. Try to login with Bitbucket 
### Your personal set up
jupyterhub 4.0.2
oauthenticator 16.0.5

```python
c.JupyterHub.authenticator_class = "bitbucket"
c.OAuthenticator.oauth_callback_url = "http://localhost/hub/oauth_callback"
c.OAuthenticator.client_id = "CLIENT_ID"
c.OAuthenticator.client_secret = "SECRET"
c.OAuthenticator.allowed_users = ["user1", "user2"]
c.BitbucketOAuthenticator.allowed_teams = {"teamname"}
c.SubBitbucketAuthenticator.scope = ["email", "account", "repository", "workspace"]
```

### Workaround

Currently I have setup `OAUTH2_USERDATA_REQUEST_TYPE=url` in the environment to authenticate on the user API, and am subclassing the `_fetch_user_teams` method as below to restrict the workspace.

I feel it might be better solved by putting `userdata_token_method` into `httpfetch` if appropriate and in subclasses setting `self.userdata_token_method = url` in the bitbucket class.

```python
class SubBitbucketOAuthenticator(oauthenticator.BitbucketOAuthenticator):
    async def _fetch_user_teams(self, access_token, token_type):
        """
        Get user's team memberships via bitbucket's API.
        """
        try:
            app_log.debug("Trying usual bitbucket OAuth method...")
            return await super()._fetch_user_teams(access_token, token_type)
        except HTTPClientError as e:
            if e.code != 401:
                raise
        app_log.debug("Trying to authenticate with with access_token in the url")
        headers = self.build_userdata_request_headers(access_token, token_type)
        del headers["Authorization"]

        next_page = url_concat(
[...]

PR #698, opened 2023-11-19
Correcting Bearer Authorization header in OAuth2

yaleman:
Currently the value for the "Authorization" header is taken from the case-insensitive "token_type" field (RFC 6750) sent by the OAuth IDP. This PR corrects for this and matches the RFC 6749 requirement of having title-cased "Bearer".

References: https://github.com/jupyterhub/oauthenticator/issues/677 and https://github.com/jupyterhub/oauthenticator/issues/676

### DIFF
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -716,10 +716,17 @@ class OAuthenticator(Authenticator):
         Builds and returns the headers to be used in the userdata request.
         Called by the :meth:`oauthenticator.OAuthenticator.token_to_user`
         """
+
+        # token_type is case-sensitive, but the headers are
+        if token_type.lower() == "bearer":
+            auth_token_type = "Bearer"
+        else:
+            auth_token_type = token_type
+
         return {
             "Accept": "application/json",
             "User-Agent": "JupyterHub",
-            "Authorization": f"{token_type} {access_token}",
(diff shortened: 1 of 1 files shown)


oauthenticator/oauth2.py
```diff
@@ -716,10 +716,17 @@ def build_userdata_request_headers(self, access_token, token_type):
         Builds and returns the headers to be used in the userdata request.
         Called by the :meth:`oauthenticator.OAuthenticator.token_to_user`
         """
+
+        # token_type is case-sensitive, but the headers are
```
manics, line 720: Should this be _isn't_ case-sensitive?

manics (commented):
Other than the comment this seems reasonable.
@GeorgianaElena what do you think?

oauthenticator/oauth2.py
yaleman: Ah, whoops. 😄 Sorted!

GeorgianaElena (approved):
This looks good to me! Thanks @yaleman and @manics 🚀

consideRatio:
Thank you @yaleman and thank you @jyio for the amazing resaerch in https://github.com/jupyterhub/oauthenticator/issues/677#issuecomment-1817859638 - wow!! :heart: :tada: :sunflower:

PR #701, opened 2023-11-21
[CILogon] allow fnmatch based expressions in `allowed_domains`, such as `*.jupyter.org`

### LINKED ISSUE #692: [CILogon] `allowed_domains` to support wildcards like `*`
From significant usage of CILogonOAuthenticator with 2i2c at https://github.com/2i2c-org/infrastructure, we've found that a quite common need is to declare that "all domains matching a certain pattern" is relevant to match against. For example, all domains ending with `.edu`.

If we want, we could make `allowed_domains` more flexible, and support this in a non-breaking way by code like this in `CILogonOAuthenticator.check_allowed` :

```python
                    for ad in idp_allowed_domains:
                        # fnmatch allow us to use wildcards like * and ?, but
                        # not the full regex. For simple domain matching this is
                        # good enough. If we were to use regexes instead, people
                        # will have to escape all their '.'s, and since that is
                        # actually going to match 'any character' it is a
                        # possible security hole. For details see
                        # https://docs.python.org/3/library/fnmatch.html.
                        if fnmatch(email_domain, ad):
                            return True
```

## Related

- https://github.com/jupyterhub/oauthenticator/issues/547

consideRatio:
- Closes #692 that includes more information

### DIFF
--- a/oauthenticator/cilogon.py
+++ b/oauthenticator/cilogon.py
@@ -2,6 +2,7 @@
 A JupyterHub authenticator class for use with CILogon as an identity provider.
 """
 import os
+from fnmatch import fnmatch
 from urllib.parse import urlparse
 
 import jsonschema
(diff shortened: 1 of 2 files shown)


GeorgianaElena (approved):
Great! Thanks @consideRatio!

minrk (approved):

consideRatio:
Wiee thanks for reviewing @GeorgianaElena!

Issue #706, opened 2023-11-28
GenericOAuthenticator should populate groups if `claim_groups_key` is set


benjimin:
### Proposal

If an administrator chooses to configure `c.GenericOAuthenticator.claim_groups_key`, then the list of groups (from the OAuth token claim) should be made readily available in `auth_state`. 

Setting `c.Authenticator.manage_groups = True` should also cause `spawner.user.groups` to be populated with the same list.

### Use-case

This is so administrators can configure an OAuth provider to manage authentication _and_ group membership, for example to customise the Jupyterhub server profile list based on (externally managed) group memberships.

This idea is explicitly [mentioned](https://z2jh.jupyter.org/en/latest/jupyterhub/customizing/user-environment.html#user-dependent-profile-options) in the Z2JH customisation guide (using OAuth group membership to decide whether to offer GPUs).

### Is this a bug? Or a feature request?

At present `claim_groups_key` seems to have no effect unless `admin_groups` or `allowed_groups` are set. (The `claim_groups_key` [docs](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.generic.html#oauthenticator.generic.LocalGenericOAuthenticator.claim_groups_key) give little indication if its purpose were intended to be so narrow.)

The [docs](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.generic.html#oauthenticator.generic.GenericOAuthenticator.manage_groups) for `manage_groups` lead one to expect that the authenticator should assign a list of group names to the user. 🐜

### Alternative options

I think it is pragmatic to only make changes to `GenericOAuthenticator` at this stage. An alternative would be to make the changes somewhere deeper in the class hierarchy (potentially adjusting the interfaces), because administrators are likely to want other varieties of authenticator to also have similar functionality (of populating groups from externally-managed sources), and it would be advantageous if the expected behaviour and configuration was made consistent. (For example, `GitHubOAuthenticator` and `GlobusOAuthenticator` currently each implement their own peculiar class methods for configuring group management.)

[...]

PR #707, opened 2023-11-28
clarify what claim_groups_key is used for

minrk:
and explicitly state that it's not related to JupyterHub groups, so it's clear and consistent with all other OAuthenticators.

This should alleviate some of the confusion in #706, but doesn't address the feature request of implementing `Authenticator.manage_groups` support, which is more general and should probably be implemented across OAuthenticator.


consideRatio (approved):

consideRatio:
I did `documentation` on this since it was a docstring surfaced to configuration reference - could be `maintenance` also.

minrk:
I think docs is right for docstrings.

PR #718, opened 2024-01-18
Make username_claim callable in CILogon

yuvipanda:
Companion to https://github.com/jupyterhub/oauthenticator/pull/717

Fixes https://github.com/jupyterhub/oauthenticator/issues/712


yuvipanda:
Two issues here:

1. `jsonschema` by default doesn't allow us to really do anything about callables or python objects, so we'll need to find a way to fix that
2. Not sure why the CILogon tests are failing. I need to understand that better, and fix those.

yuvipanda:
I'm not going to be working on this actively right now so going to close this

PR #719, opened 2024-01-20
Add `allowed_scopes` to all authenticators to allow some users based on granted scopes

yuvipanda:
We can already control what scopes we ask for by setting 'scope'. However, OAuth2
[states](https://datatracker.ietf.org/doc/html/rfc6749#section-3.3) that not all the scopes requested may be granted, for whatever reason. This could be because the user choose not to give us the grant, or because the user themselves doesn't have access to grant us this scope (This is how Auth0 uses it for example -
https://auth0.com/docs/get-started/authentication-and-authorization-flow/authorization-code-flow/call-your-api-using-the-authorization-code-flow).

This PR adds an `allowed_scopes` property, which allows granting access to a subset of users based on the presence of a particular scope in the list of scopes granted. This can be used in addition to other authorization mechanisms to allow access to the hub.

### DIFF
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -1048,6 +1048,7 @@ class OAuthenticator(Authenticator):
             granted_scopes = auth_model.get('auth_state', {}).get('scope', [])
             missing_scopes = set(self.required_scopes) - set(granted_scopes)
             if missing_scopes:
+                self.log.info(f"Denying access to user {username} - scopes {missing_scopes} were not granted")
                 return False
             else:
                 return True
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -483,9 +483,6 @@ class OAuthenticator(Authenticator):
         # required scopes must be a subset of requested scopes
         if set(proposal.value) - set(self.scope):
             raise ValueError(f"Required Scopes must be a subset of Requested Scopes. {self.scope} is requested but {proposal.value} is required")
-        if self.allow_all:
-            # Can't set allow_all *and* require scopes
-            raise ValueError("Required Scopes is mutually exclusive with allow_all. Unset one of those configuration properties")
         return proposal.value
 
     extra_authorize_params = Dict(
@@ -1046,11 +1043,8 @@ class OAuthenticator(Authenticator):
         if auth_model is None:
             return True
 
-        if self.allow_all:
-            return True
-
         # If we specific scope grants are required, validate that they have been granted
-        # If not, we explicitly raise an exception here, since they should not be allowed
+        # If not, we explicitly raise an exception here, since the user should not be allowed
         # *regardless* of any other config allowing them access.
         if self.required_scopes:
             granted_scopes = auth_model.get('auth_state', {}).get('scope', [])
@@ -1059,8 +1053,9 @@ class OAuthenticator(Authenticator):
                 message = f"Denying access to user {username}. {self.scope} scopes were requested, {self.required_scopes} are required for authorization, but only {granted_scopes} were granted"
                 self.log.info(message)
                 raise web.HTTPError(403, message)
-            else:
-                return True
+
+        if self.allow_all:
+            return True
 
         # allow users with admin status set to True via admin_users config or
         # update_auth_model override
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -482,7 +482,9 @@ class OAuthenticator(Authenticator):
     def _required_scopes_validation(self, proposal):
         # required scopes must be a subset of requested scopes
         if set(proposal.value) - set(self.scope):
-            raise ValueError(f"Required Scopes must be a subset of Requested Scopes. {self.scope} is requested but {proposal.value} is required")
+            raise ValueError(
+                f"Required Scopes must be a subset of Requested Scopes. {self.scope} is requested but {proposal.value} is required"
+            )
         return proposal.value
 
     extra_authorize_params = Dict(
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -268,6 +268,9 @@ class OAuthenticator(Authenticator):
         help="""
         Allow all authenticated users to login.
 
+        Note that if required_scopes is set, users who aren't granted the scopes
+        listed in that will still not be allowed to login.
+
         .. versionadded:: 16.0
         """,
     )
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -1053,7 +1053,7 @@ class OAuthenticator(Authenticator):
             missing_scopes = set(self.required_scopes) - set(granted_scopes)
             if missing_scopes:
                 self.log.info(
-                    f"Denying access to user {username} - scopes {missing_scopes} were not granted"
+                    f"Denying access to user {username}. {self.scope} scopes were requested, {self.required_scopes} are required for authorization, but only {granted_scopes} were granted"
                 )
                 return False
             else:
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -482,7 +482,7 @@ class OAuthenticator(Authenticator):
         # allowed scopes must be a subset of requested scopes
         if set(proposal.value) - set(self.scope):
             raise ValueError(
-                f"Allowed Scopes must be a subset of Requested Scopes. {self.scope} is requested but {proposal.value} is allowed"
+                f"Allowed scopes must be a subset of requested scopes. {self.scope} is requested but {proposal.value} is allowed"
             )
         return proposal.value
 
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -268,9 +268,6 @@ class OAuthenticator(Authenticator):
         help="""
         Allow all authenticated users to login.
 
-        Note that if required_scopes is set, users who aren't granted the scopes
-        listed in that will still not be allowed to login.
-
         .. versionadded:: 16.0
         """,
     )
@@ -459,19 +456,18 @@ class OAuthenticator(Authenticator):
         """,
     )
 
-    required_scopes = List(
+    allowed_scopes = List(
         Unicode(),
         config=True,
         help="""
-        List of scopes that must be granted to allow login.
+        Allow users who have been granted *all* these scopes to log in.
 
-        All the scopes listed in this config must be present in the OAuth2 grant
-        from the authorizing server to allow the user to login. We request all
-        the scopes listed in the 'scope' config, but only a subset of these may
-        be granted by the authorization server. This may happen if the user does not
-        have permissions to access a requested scope, or has chosen to not give consent
-        for a particular scope. If the scopes listed in this config are not granted,
-        the user will not be allowed to log in.
+        We request all the scopes listed in the 'scope' config, but only a
+        subset of these may be granted by the authorization server. This may
+        happen if the user does not have permissions to access a requested
+        scope, or has chosen to not give consent for a particular scope. If the
+        scopes listed in this config are not granted, the user will not be
+        allowed to log in.
 
         The granted scopes will be part of the access token (fetched from self.token_url).
         See https://datatracker.ietf.org/doc/html/rfc6749#section-3.3 for more
@@ -481,12 +477,12 @@ class OAuthenticator(Authenticator):
         """,
     )
 
-    @validate('required_scopes')
-    def _required_scopes_validation(self, proposal):
-        # required scopes must be a subset of requested scopes
+    @validate('allowed_scopes')
+    def _allowed_scopes_validation(self, proposal):
+        # allowed scopes must be a subset of requested scopes
         if set(proposal.value) - set(self.scope):
             raise ValueError(
-                f"Required Scopes must be a subset of Requested Scopes. {self.scope} is requested but {proposal.value} is required"
+                f"Allowed Scopes must be a subset of Requested Scopes. {self.scope} is requested but {proposal.value} is allowed"
             )
         return proposal.value
 
@@ -1048,16 +1044,14 @@ class OAuthenticator(Authenticator):
         if auth_model is None:
             return True
 
-        # If we specific scope grants are required, validate that they have been granted
-        # If not, we explicitly raise an exception here, since the user should not be allowed
-        # *regardless* of any other config allowing them access.
-        if self.required_scopes:
+        # Allow users who have been granted specific scopes that grant them entry
+        if self.allowed_scopes:
             granted_scopes = auth_model.get('auth_state', {}).get('scope', [])
-            missing_scopes = set(self.required_scopes) - set(granted_scopes)
-            if missing_scopes:
-                message = f"Denying access to user {username}. {self.scope} scopes were requested, {self.required_scopes} are required for authorization, but only {granted_scopes} were granted"
+            missing_scopes = set(self.allowed_scopes) - set(granted_scopes)
+            if not missing_scopes:
+                message = f"Granting access to user {username}, as they had {self.allowed_scopes}"
                 self.log.info(message)
-                raise web.HTTPError(403, message)
+                return True
 
         if self.allow_all:
             return True
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -1048,7 +1048,9 @@ class OAuthenticator(Authenticator):
             granted_scopes = auth_model.get('auth_state', {}).get('scope', [])
             missing_scopes = set(self.required_scopes) - set(granted_scopes)
             if missing_scopes:
-                self.log.info(f"Denying access to user {username} - scopes {missing_scopes} were not granted")
+                self.log.info(
+                    f"Denying access to user {username} - scopes {missing_scopes} were not granted"
+                )
                 return False
             else:
                 return True
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -456,6 +456,24 @@ class OAuthenticator(Authenticator):
         """,
     )
 
+    required_scopes = List(
+        Unicode(),
+        config=True,
+        help="""
+        List of scopes that must be granted to allow login.
+
+        All the scopes listed in this config must be present in the OAuth2 grant
+        from the authorizing server to allow the user to login. We request all
+        the scopes listed in the 'scope' config, but only a subset of these may
+        be granted by the authorization server. This may happen if the user does not
+        have permissions to access a requested scope, or has chosen to not give consent
+        for a particular scope. If the scopes listed in this config are not granted,
+        the user will not be allowed to log in.
+
+        See the OAuth documentation of your OAuth provider for various options.
+        """,
+    )
+
     extra_authorize_params = Dict(
         config=True,
         help="""
@@ -1024,6 +1042,15 @@ class OAuthenticator(Authenticator):
         # automatically with existing users if it was configured truthy
         if username in self.allowed_users:
             return True
+
+        # If we specific scope grants are required, validate that they have been granted
+        if self.required_scopes:
+            granted_scopes = auth_model.get('auth_state', {}).get('scope', [])
+            missing_scopes = set(self.required_scopes) - set(granted_scopes)
+            if missing_scopes:
+                return False
+            else:
+                return True
 
         # users should be explicitly allowed via config, otherwise they aren't
         return False
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -479,7 +479,7 @@ class OAuthenticator(Authenticator):
     )
 
     @validate('required_scopes')
-    def _required_scopes_validation(self, proposal: list[str]):
+    def _required_scopes_validation(self, proposal):
         # required scopes must be a subset of requested scopes
         if set(proposal.value) - set(self.scope):
             raise ValueError(f"Required Scopes must be a subset of Requested Scopes. {self.scope} is requested but {proposal.value} is required")
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -18,7 +18,7 @@ from tornado.auth import OAuth2Mixin
 from tornado.httpclient import AsyncHTTPClient, HTTPClientError, HTTPRequest
 from tornado.httputil import url_concat
 from tornado.log import app_log
-from traitlets import Any, Bool, Dict, List, Unicode, default
+from traitlets import Any, Bool, Dict, List, Unicode, default, validate
 
 
 def guess_callback_uri(protocol, host, hub_server_url):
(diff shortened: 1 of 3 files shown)


yuvipanda:
I've a local version of this deployed in https://github.com/2i2c-org/infrastructure/pull/3618, and it works well for use in auth0.

oauthenticator/oauth2.py
```diff
@@ -456,6 +456,24 @@ def _logout_redirect_url_default(self):
         """,
     )
 
+    required_scopes = List(
```
consideRatio, line 459: I think the terminology is that scopes are what you request, and claims is what you get - so this should be required_claims then I think.

A google provided this:

> Scopes are only present in OAuth Access Tokens. The ID Token never contains scopes but instead contains the resolved claims that belong to the issued scopes.

> Scopes are groups of claims.

So if you request the scope access_levels you can get claims gpu_access=true and tpu_access=false sent back.

consideRatio: Hmmm is it:
- requested scopes,
- issues scopes (groups of claims),
- and claims being key=values?

Then, should this be required_issued_scopes, or required_claims? If its required claims, should it allow to specify the claim key and/or claim key/values?

consideRatio: Ah nevermind, this is not ambiguous, its required (issued) scopes, and oauth2 provides the specific scope key in response - nice!

consideRatio (commented):
This seems reasonable and clear for future maintenance! This looks almost ready to merge to me!

- Could you update the title to include the new config name?
- Could you clarify if "scope" is expected to be returned with the access token request response and/or the userinfo response from the oauth spec, and that we get it from where its specced to be returned?
- Could you provide in code (either config help or inline code comment) reference links to the oauth2 docs of relevance?

yuvipanda:
Thanks for the review, @consideRatio! I've done the three things :)

oauthenticator/oauth2.py
yuvipanda: Thanks for digging through this :) I always forget the terminology if I'm away from oauth2 for like, 3 weeks.

oauthenticator/oauth2.py
```diff
@@ -1025,6 +1047,18 @@ async def check_allowed(self, username, auth_model):
         if username in self.allowed_users:
             return True
 
+        # If we specific scope grants are required, validate that they have been granted
```
consideRatio, line 1050: I've considered if this should be in an earlier stage than check_allowed. Its nice to have it here if seen as authorization logic, but it could be seen as a technical requirement - for example to get relevant info for the username. Then, it would be more suitable that not meeting this criteria is seen ss an error rather than 403.

Hmmmm... Consider if we request email scope and username_claim email and declare it as email scope required but don't get it - should that be another kind of error rather than denied authorization 403?

If we think so, this could be more like an assertion early in update_auth_model instead raising an error maybe? Wdyt?

consideRatio: I'm not sure what i think, but it wouldn't be breaking to relocate this check so I'm open with going onwards either way

yuvipanda: Good question. There are two ways to think about this:

1. It's part of *authentication* logic, as if we aren't granted a specific *claim* from a scope, we can't even log the user in (for username_claim, for example). So if the scope isn't granted, we can't actually validate who this user is. So it gets relocated elsewhere.
2. It is part of *authorization* logic, where scopes are used as a way to 'gate' access *after* we could validate who the user is. So we *know* who the user is, but we don't want to give them access. In this case it stays where it is.

So how do we best determine if this is part of authentication or authorization?

I went to look at possible available scopes for a bunch of authenticators:

- [GitHub scopes list](https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/scopes-for-oauth-apps)
- [Mediawiki rights](https://www.mediawiki.org/wiki/Extension:OAuth/en#User_rights)
- [Standard OpenID connect scopes](https://auth0.com/docs/get-started/apis/scopes/openid-connect-scopes#standard-claims) (as supported by Auth0)

`username_claim` feels clearly part of authentication to me, and right now, if we don't get a `username_claim` back for any reason, it's already a 403. But the 403 is a little bit unclear on why exactly the permission was denied and can be frustrating for the user. However, the scopes that can be related to `username_claim` are few in number. In GitHub for example, it really is just `user` - its unlikely you would want the username to be derived from anything else (for the most part). I think this is generally true for most auth providers, particularly given the prevalance & design decisions of openid connect. This means that when configuring, if the `username_claim` is not present due to a missing `scope`, it *most likely* is because the hub admin did not *ask* for the scope, and it's going to be clear to them in initial testing.

However, you may want to deny *authorization* for some subsets - for example, someone may want to setup a specific ssh key in github for a JupyterHub, and hence reject authorization if they don't grant `write:public_key`, even though it has nothing to do with authentication itself - we know who the user is, just don't want them in here.

I'm a big fan of @GeorgianaElena's work in https://github.com/jupyterhub/oauthenticator/pull/594, and think that separation of authentication and authorization here makes things easier to reason about. I think the most likely use of this feature is *authorization* and not *authentication*. `username_claim` being set to something that isn't present if a specific scope is not granted is IMO a different problem, and we can tackle that differently if you desire - I'd say the primary extra bit there is to find a way to provide useful error information. 

All this to say, I think it's ok for this check to remain here :)

yuvipanda: I opened https://github.com/jupyterhub/oauthenticator/pull/721 to make the missing username_claim situation slightly easier.

minrk (approved):
Makes great sense. Approved by me, pending @consideRatio's more detailed feedback.

oauthenticator/oauth2.py
consideRatio: Thanks for reasoning about this, I think this makes a good case for considering this to be part of authorization (authz) logic rather than authentication (authn) logic!

consideRatio (changes requested):
> If the scopes listed in this config are not granted, the user will not be allowed to log in.

I think this isn't correctly implemented currently.

- If `allow_all` is true, and the `required_scopes` isn't met, this will not deny authorization.
- Further, if a subclass like Generic's `check_allowed` is called, it will simply return True if the base class returns true, but not return false if the base class return false.

In general terms, `check_allowed` as implemented in Authenticator, replaced in OAuthenticator, and augmented in GenericOAuthenticator etc, needs to handle three situations.

1. The authenticated user gets disallowed
2. The authenticated user gets allowed
3. The authenticated user doesn't get disallowed or allowed, which will in the end lead to a default of being disallowed.

This could be implemented if `check_allowed` makes use of True/None/False, or if `check_allowed` makes use of True/Raise error/False. I'm confident we can safely take the raise error path to handle this, but I'm not confident we can adopt a True/None/False approach without breaking interactions with subclasses or JupyterHub itself.

With this considered, I think the path forward is to:
- relocate the `required_scopes` check to the first thing happening in `check_allowed`
- make the `required_scopes` check do nothing or raise an error

yuvipanda:
@consideRatio I spent some more time thinking through this, and agree that as is worded it doesn't actually do what it says in the documentation. So I've made the following changes:

1. If this check fails, we raise a 403 (similar to what happens in other places in the CILogonAuthenticator) with a useful error message (incorporating the suggestions from @minrk)
2. I've moved the check up, but still *after* allow_all - as I think allow_all should basically sidestep everything. However, since this can be a bad footgun, I've also added validation that makes sure that `allow_all` and `required_scopes` are *mutually exclusive* - you can't have both on. I think this makes sense, as by definition if you require some scopes for authorization, you aren't actually 'allowing all'.
3. I added the fact that you can raise a 403 to the method documentation
4. I also added validation to make sure that you aren't requiring scopes you aren't requesting (thanks @minrk).

How does it look to you now?

yuvipanda:
And if we decide throwing a 403 here is ok, we should document it in JupyterHub too - I've done so at https://github.com/jupyterhub/jupyterhub/pull/4682

consideRatio:
To give allow_all special power in allowing, making it more allowing than allowed_users for example, is not intuitive at all in my mind. It would also mess with the use case of allowing all users as long as they show up with the required issued scope - isn't this a key use case?

required_scopes and allow_all makes sense together imo - a user must be allowed and have the required scopes to be authorized - either the user is allowed via allow_all or allowed_users etc.

manics:
The original idea behind `allow_all` is to allow any user without having to define allowed users/groups. For example, admins were inadvertently allowing any GitHub user to login to their installations, which is why we made a breaking change to add this, with a default of False
https://github.com/jupyterhub/oauthenticator/pull/625

If `required_scopes` is mutually exclusive with `allow_all` then you must also also set allowed users/groups, and also means you can't allow all. Therefore I think the most intuitive behaviour is to say `allow_all` is the last check after all other prerequisites.

oauthenticator/oauth2.py
```diff
@@ -1015,6 +1053,19 @@ async def check_allowed(self, username, auth_model):
...
+        # *regardless* of any other config allowing them access.
+        if self.required_scopes:
+            granted_scopes = auth_model.get('auth_state', {}).get('scope', [])
+            missing_scopes = set(self.required_scopes) - set(granted_scopes)
+            if missing_scopes:
+                message = f"Denying access to user {username}. {self.scope} scopes were requested, {self.required_scopes} are required for authorization, but only {granted_scopes} were granted"
+                self.log.info(message)
+                raise web.HTTPError(403, message)
+            else:
+                return True
```
consideRatio, line 1067: I dom't want required scopes to return true ever, now its line allowing users as long as you have the required scopes.

How then could you use this with any other allow logic? Use of this would then bypass all allow logic - its required, but if you have it, you are authorized and can't distinguish a subset of users meeting this requirement.

yuvipanda:
I think there's a bunch of confusion around what `allow_all` really is doing here that's driving this, and I'm going to try to work through that.

The current doc string for 'allow_all' says:

> Allow all authenticated users to login.

So in this conversation (https://github.com/jupyterhub/oauthenticator/pull/719#discussion_r1462531556) we established that required_scopes is part of the authorization check, rather than authentication. So to me that means that if *authentication succeeds*, allow_all should kick in and allow the user *regardless of anything else*. We are basically ignoring

> To give allow_all special power in allowing, making it more allowing than allowed_users for example, is not intuitive at all in my mind.

@consideRatio my reading of the code is that this is the current behavior - if you set `allow_all` to be true, `allowed_users` is completely ignored. There's a `return True` [here](https://github.com/jupyterhub/oauthenticator/blob/5667bf15f8e1dead29c32cc0886146d6ddb835ec/oauthenticator/oauth2.py#L1015) which means the [allowed_users check](https://github.com/jupyterhub/oauthenticator/blob/5667bf15f8e1dead29c32cc0886146d6ddb835ec/oauthenticator/oauth2.py#L1025) is never reached. And I don't see that check being performed anywhere else either. I *do* agree this is unintuitive,  and I'd suggest we clarify this by raising an error if both allow_all and allowed_users is both set. Am happy to open a PR for doing that.

Basically if `allow_all` is set to True, and is meant to 'Allow *all* authenticated users', it means we completely turn off all other *authorization* logic. This would work in cases where we defer *authorization* to the external auth provider as well - for example, if we configure CILogon to accept only a particular university's credentials. 

So my understanding of `allow_all` is that it's really `allow_all_authenticated`, and we completely skip authorization offered as part of base OAuthenticator.

However, I do see that in *subclasses*, we do offer additional checks. So if the following config is present with say GitHubOAuthenticator:

```yaml
allow_all: True
allowed_users: ["user1"]
allowed_organizations: ["org1"]
```

What happens is that `allowed_users` is *silently ignored*, but `allowed_organizations` is enforced! If instead the config is the following:

```yaml

```yaml
allow_all: False
allowed_users: ["user1"]
allowed_organizations: ["org1"]
```

Then both `allowed_users` and `allowed_organizations` is respected.

I think this is very confusing and unintuitive.

So `allow_all` is *not* `allow_all_authenticated`, but as implemented (before this PR), it's really `ignore_allowed_users_and_config_admin_users`. With this PR (as is), it becomes `ignore_allowed_users_and_required_scopes_and_config_admin_users`, which I agree actually makes this worse not better :) So some kind of change is necessary, although I'm not sure what exactly.

Is my understanding of the situation and reasoning correct? What am I missing?

consideRatio:
My understanding is that we currently have ways to allow users, and they can overlap, where allow_all is guaranteed to overlap with any other way.

Since oauthenticator 16, there is no way to be be authorized without being explicitly allowed by one config.

Also in oauthenticator 16, no config able to allow would disallow users. In oauth 15 on the other hand, user not in a github org was disallowed, even if they were allowed in allowed_users or admin_users for example.

So now we have pure allow rules and pure disallow rules or requirements on the users besides being allowed, and the logic is trivial compared to when we had combinations of behavior on allowing/disallowing.

I'm uncertain about your confusion points at the moment, but i think it relates to old oauthenticator 15 behaviors.

yuvipanda:
Since the `allow_all` situation is unrelated to this PR, I've now done the following:

1. Move `required_checks` above `allow_all`, and have it not return anything. So this acts purely as a way to *deny* some people access, rather than *allow* anyone access.
2. Remove the validation that allow_all and required_scopes is mutual. I've kept the validation that required_scopes should be a subset of requested scopes.

yuvipanda:
@consideRatio 

> I'm uncertain about your confusion points at the moment, but i think it relates to old oauthenticator 15 behaviors.

I linked to specific code points in current main branch when discussing my reasons about `allow_all` and `allowed_users`, so I don't think this is true.

yuvipanda:
Regardless, since the `allow_all` situation is unrelated to this PR, I'll open a separate issue about that. I've made changes here as requested.

consideRatio:
> ```yaml
> allow_all: True
> allowed_users: ["user1"]
> allowed_organizations: ["org1"]
> ```
>
> What happens is that allowed_users is silently ignored, but allowed_organizations is enforced!

What do you mean enforced? allowed_organizations is allowing members of those, but doesnt disallow those outside in oauthenticator 16, so you can use allowed_organizations + a few allowed_users + a few admin_users. Use of allow_all just makes it pointless to allow users with allowed_organizations or allowed_users.

But, at the same time. Maybe you want to briefly allow all users, so you toggle it on briefly, and then later toggles it off. So, then its good to allow to be next to other config in a way.

yuvipanda:
@consideRatio I've opened https://github.com/jupyterhub/oauthenticator/issues/723 to carry on this discussion, separated out from the functionality this PR is adding.

oauthenticator/oauth2.py
yuvipanda: This is done.

yuvipanda:
@consideRatio and short version seems to be that my understanding is missing some pieces, so I don't really think there's any action item for anyone else. I may make a PR with a better docstring once I'm done digging through this. I'll update #723 accordingly.

This PR currently implements your suggestions in https://github.com/jupyterhub/oauthenticator/pull/719#pullrequestreview-1838530737.

yuvipanda:
I think I just missed that there's a short circuit here for any `return True`, not just for `allow_all` (https://github.com/jupyterhub/oauthenticator/blob/5667bf15f8e1dead29c32cc0886146d6ddb835ec/oauthenticator/github.py#L173). Sorry for the noise, @consideRatio and @manics.

yuvipanda:
After muddling through a little, I think I've a clearer idea of my issue here.

1. If we say `allow_all` will allow *all* authenticated users,
2. And that `required_scopes` is an *authorization* step, not an *authentication* step
3. The PR as it is now means that `required_scopes` is the *only* authorization step that is not affected by allow_all.

I've made https://github.com/jupyterhub/oauthenticator/pull/719/commits/ea2972c909924f2ed1f99c200571d82fdaac5f83 this commit updating the docstring for allow_all - as `allow_all` will still respect `requested_scopes` even if it ignores every other authorization step.

Does this feel iffy to anyone else?

yuvipanda:
Now that I have a better understanding of what exactly my confusion was, let me rework this PR just a little to make that be better.

yuvipanda:
Ok so the core issue was that this was an authorization check that was kinda being implemented like an authentication check. Since 16, OAuthenticator has many different `allowed_X` properties that gate which users are allowed to access the hub and which are not - these are all *authorization* properties. `required_scopes` was a *bit* like that, but not entirely, and that led down my rabbit hole of confusion.

I've now changed the PR to make this much clearer. The traitlet is now `allowed_scopes`, and if *all* of them are present, it grants you access to the hub automatically. This now works very much like *all* the other `allowed_X` properties, and comes *after* `allow_all` as well. So we don't actually have to modify the docstring of `allow_all` to talk about this exception.

This is different from `allowed_organizations` in that *all* the scopes in this list must be present, while `allowed_organizations` is fine with *any* of the orgs present. I think this could also be changed, and I'll work through that.

But regardless, I think this fits in much nicer than `required_scopes`. I know I went down a blind alley for a while, but I'm very happy with where I've come out of my confusion with. Thanks for the review @consideRatio and @manics

consideRatio:
> The traitlet is now `allowed_scopes`, and if _all_ of them are present, it grants you access to the hub automatically. This now works very much like _all_ the other `allowed_X` properties, **and comes _after_ `allow_all` as well. So we don't actually have to modify the docstring of `allow_all` to talk about this exception.**

I see `allow_all` as one one of several properties allowing users. That other allow properties only allow subsets of the users allowed by `allow_all` isn't weird or a proplem in my mind.

I think it should be allowed to configure `allow_all=True` without erroring even if other allow properties are declared. I also reason that it doesn't matter if you `return True` from point A or B in the `check_allowed` function as implemented in oauthenticator as we never return False before checking all paths towards `return True`.

consideRatio:
@yuvipanda I think the pivot to providing `allowed_scopes` instead of `required_scopes` is reasonable and aligns more with other config. At the same time, its fundamentally different, where there can become a need for `required_scopes` alongside this config, which is fine.

### Naming

I'd like to propose the naming `allowing_scopes`, users issued all scopes in one set of scopes will be allowed access.

I propose another name than `allowed_scopes` as that sounds like its about scopes being allowed (which is hard to make sense of) rather than users issued those scopes are allowed. Overall `allowed_scopes` also sounds like it could be a requirement to meet, but its just one possible way to get allowed access as a user.

### All scopes, or any of the scopes?

If we start with allowing users issued _all_ scopes in a set of scopes via this config, we can then easily either now or later let the config accept a set of "set of scopes", where having any set of scopes fully issued would allow the user.

Example configurations

```python
# allow users for whom both a and b are issued
# this represents the PR in its current form - a single set of scopes is provided
allowing_scope = {"a", "b"}

# allow users for whom either a are issued or b1 and b2 are issued
allowing_scopes = {
  {"a"},
  {"b1", "b2"},
}
```

minrk:
I think `allowed_scopes` is preferable, since it shares a namespace with other `allowed_` config. I think there is a benefit to all such names sharing a prefix and having consistent behavior. It's not clear to me without further explanation what `allowing_scopes` would mean, which is not a natural phrase, to me.

> I think it should be allowed to configure allow_all=True without erroring even if other allow properties are declared.

A warning could provide some experience improvement to make it clearer that some config will have no effect, but I agree an error is too far.

manics:
`allowed_scopes` is going to be inconsistent anyway, since all scopes are required, whereas with the other properties you only have to be a member of one.

If we wanted to be consistent we could go with the suggestion in https://github.com/jupyterhub/oauthenticator/pull/719#issuecomment-1913262663 of makeing `allowed_scopes` a list of lists?

consideRatio:
Considering naming, here is an overview of the config that allows users:

- `admin_users`
- `allowed_users`
- (oauth) `allow_all`
- (oauth) `allow_existing_users`
- (bitbucket) `allowed_teams`
- (cilogon) `allowed_domains`
- (generic, openshift) `allowed_groups`
- (github) `allowed_organizations`
- (gitlab) `allowed_gitlab_groups`
- (globus) `allowed_globus_groups`
- (google) `allowed_google_groups`
- (google) `admin_google_groups`

We have the prefixes `admin`, `allow`, and `allowed` so far, so adding `allowing` would be another.

Looking at the terminology in https://datatracker.ietf.org/doc/html/rfc6749#section-3.3, I think the terminology is "requested scope" and "granted scope" (the access token is "issued" and the scope the token has is "granted").

Do you consider `allow_granted_scope` to be an improvement to `allowed_scope` @yuvipanda @minrk ? I think it would help convey the config function better from its name than `allowed_scope`.

manics:
`allowed_scopes` or similar is going to be inconsistent anyway, since all scopes are required, whereas with the other properties only one is required.

consideRatio:
> `allowed_scopes` or similar is going to be inconsistent anyway, since all scopes are required, whereas with the other properties only one is required.

If we support the sets of sets / lists of lists version, its like allowed_scope_sets, where each set of scopes are granted, making it consistent with being individual parts allowing a user access, but each part is a "scope set" instead of individual requested scope entries.

Function when providing a single set could be made to mean "allow when any of scope entries are granted" (OR) or "allow when all of scope entries are granted" (AND), do you think it should be the OR version if just passing the config a single set/list of scopes @manics?

manics:
I don't know..... given this is a new property we could go for the list-of-lists only? It's more complciated for admins but at least it's unambiguous?

yuvipanda:
Thanks for the comments everyone, I'm very swamped this week but will come back to this next week.

yuvipanda:
Sigh, momentum seems so important to me getting stuff done. There are merge conflicts here already :(

yuvipanda:
ok, so my understanding is that there are two questions at play:

1. Naming
2. Should this be a list of lists from the start?

For naming, I would like to propose we keep this as is, as I agree with @minrk's comments in https://github.com/jupyterhub/oauthenticator/pull/719#issuecomment-1914178078. I also think `allowing` would make sense if *any* of the scopes granted access, but in this case, you need *all* to be granted access.

For list of lists, the question for me is if we can add that in the future in a way that doesn't break backwards compatibility. Can we meaningfully differentiate between a 'list of strings' vs 'list of lists' scenario? The way I approached this is by trying to figure out what documentation for such a property in the future would look like. If we add this feature in the future, here's what the docstring may be like

```
Allow users to login based on the scopes they have been granted.

If this is a list of strings, then users must be granted *all* these scopes
to be able to log in.

If this is a list of list of strings, then the user must be granted all the
scopes in at least one of the lists to be allowed to log in.
```
     
This reads natural enough to me. You can either have a list of list of strings, or for the most common case, a list of strings. It's unambiguous. We don't even know if admins would want to have a list of list of strings. We do know that they would want to have a list of strings. Given that we can evolve this later on in a backwards compatible way, I think it's *safe to try* this one as is.

So to summarize:

1. I think `allowed_scopes` is a good enough name, and consistent with the rest of our config.
2. We can add list of list of strings in the future if that sees an actual use case.

I would *love* to see this be merged! I think it's ready to go as is, now that I've fixed the merge conflicts.

minrk (approved):
I like all of @yuvipanda's explanation and reasoning. 👍 to merge as-is from me.

yuvipanda:
anyone wanna hit merge? :)

yuvipanda:
yay, thank you @minrk

PR #720, opened 2024-01-22
Require jupyterhub>=2.2

manics:
This means we can add support for `manage_groups` without checking the JupyterHub version.

See https://github.com/jupyterhub/oauthenticator/pull/708/ which adds support for `manage_groups`, but requires conditionals in OAuthenticator and the tests.


consideRatio (approved):

yuvipanda:
Yeah, this is good to me :)

PR #721, opened 2024-01-22
Suggest possible config fix when correct scopes aren't requested

yuvipanda:
If the hub config doesn't contain appropriate scopes to request from the provider, the token we retrieve may not have the key we are using as `username_claim`. This is the *most common* reason that `username_claim` is not present in the token. Add a simple debugging hint here to make lives of hub admins easier.

### DIFF
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -770,7 +770,7 @@ class OAuthenticator(Authenticator):
         """
         username = user_info.get(self.username_claim, None)
         if not username:
-            message = (f"No {self.username_claim} found in {user_info}",)
+            message = (f"No {self.username_claim} found in {user_info}. Maybe the hub needs to be configured to request more scopes?",)
             self.log.error(message)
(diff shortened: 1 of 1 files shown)


oauthenticator/oauth2.py
```diff
@@ -770,7 +770,9 @@ def user_info_to_username(self, user_info):
         """
         username = user_info.get(self.username_claim, None)
         if not username:
-            message = (f"No {self.username_claim} found in {user_info}",)
+            message = (
+                f"No {self.username_claim} found in {user_info}. Maybe the hub needs to be configured to request more scopes?",
```
minrk, line 774: Do you want to even suggest the config name `{self.__class__.__name__}.scopes`?

minrk (approved):
Nice! If you think it's useful to reference the config field rather than just the concept, go for it. Otherwise, go ahead and merge.

yuvipanda:
I think it's ok to just mention the concept, rather than the specific config.

PR #722, opened 2024-01-25
[Auth0] Allow redirecting to custom URL after logout

yuvipanda:
This can't be just done by setting logout_redirect_url unfortunately, as client_id is also required to be passed in here (see https://auth0.com/docs/authenticate/login/logout/redirect-users-after-logout).

While that could be done via a callable, this is much cleaner to do.

### DIFF
--- a/oauthenticator/auth0.py
+++ b/oauthenticator/auth0.py
@@ -2,6 +2,7 @@
 A JupyterHub authenticator class for use with Auth0 as an identity provider.
 """
 import os
+from urllib.parse import urlencode
 
 from jupyterhub.auth import LocalAuthenticator
 from traitlets import Unicode, default
@@ -41,6 +42,17 @@ class Auth0OAuthenticator(OAuthenticator):
             "Configuring either auth0_domain or auth0_subdomain is required"
         )
 
+    logout_redirect_to_url = Unicode(
+        config=True,
+        help="""
+        Redirect to this URL after the user is logged out.
+
+        Must be explicitly added to the "Allowed Logout URLs" in the configuration
+        for this Auth0 application. See https://auth0.com/docs/authenticate/login/logout/redirect-users-after-logout
+        for more information.
+        """
+    )
+
     auth0_subdomain = Unicode(
         config=True,
         help="""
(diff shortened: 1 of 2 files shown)


GeorgianaElena (commented):
@yuvipanda, I believe some form of logout and redirect should be possible without the client id too as Auth0 supports logout at [different session "levels"](https://auth0.com/docs/manage-users/sessions/session-layers#session-logout).

- According to [auth0 docs](https://auth0.com/docs/authenticate/login/logout/redirect-users-after-logout), for logout requests that **do not include the client_id parameter**  the returnTo URL must be added to **the Allowed Logout URLs list in the [Advanced tab of the Tenant Settings](https://manage.auth0.com/#/tenant/advanced)** ➡️ which **logouts at Auth0 Session Layer**

- Whereas, for logout requests that **include the client_id parameter**  the returnTo URL must be added to **the Allowed Logout URLs list in the Settings tab of the Auth0 app** that is associated with the specified CLIENT_ID. ➡️ **which logouts out at the Application Session Layer**

Can it be that simply setting the `logout_redirect_url` didn't work because it might not have been added to the appropriate Allowed Logout URLs (i.e. at tenant level)?

Either way, we can support logging out at the app level as well, which is what this PR does. Though the only suggestion would be in this case to name this new config different so that it's less confusing by being less similar to the initial one (logout_redirect_url) and that it's clear that it does an app-level logout?

yuvipanda:
I was experimenting with this today, and here are my results:

| - | `client_id` specified | `client_id` not specified |
| `returnTo` specified | `returnTo` is just ignored, user is redirected to whatever is the first url specified in the logout settings | `returnTo` is just ignored, user is shown a page that just says 'OK' |
| `returnTo` not specified | `returnTo` is just ignored, user is redirected to whatever is the first url specified in the logout settings | `returnTo` is just ignored, user is redirected to whatever is the first url specified in the logout settings |

Ok so that's a bit redundant, but at least what I've discovered so far seems to be that `returnTo` is ignored, and you must specify `client_id` at which point you get redirected to whatever was first in the approved list of logout URLs. This seems contrary to what they have documented

@GeorgianaElena from your links, it's possible that perhaps this logout url was added by the community to the *wrong* location. I'll follow up and debug some more. It would be nice to not need this!

yuvipanda:
From my experimentation in https://github.com/jupyterhub/oauthenticator/pull/722#issuecomment-1922427154, I've noticed that you *must* specify `client_id` for anything to work at all. This completely contradicts Auth0's own documentation.

However, `client_id` is not actually public information - it is available to anyone who clicks the 'login' url. So we don't actually need to protect this. As such, users can get the desired behavior by using the existing `logout_redirect_url` behavior, and specifying `?client_id` directly. This is what I've now documented as we should do with 2i2c (https://github.com/2i2c-org/infrastructure/pull/3883/files#diff-645ca1737a635e7140f9811524d2612c4ff4573f6b0314e5bff33087adb17e12R72).

With that, I think it's ok to close this PR!

Issue #487, opened 2022-02-21
Option to extract additional token in generic OAuth2 authenticator


michaeljfazio:
### Proposed change
I would like to extract an OIDC ID token in auth_state_hook. This seems like it should be straight forward when using
the generic authenticator however it is not OIDC friendly. Specifically, it does not provide an easy way to extract the
id_token. 

### Alternative options
Option 1: Allow users to specify additional tokens to be extracted into auth state (e.g. c.LocalGenericOAuthenticator.extra_tokens['id_token']
Option 2. Automatically detect if the token response contains an "id_token" and add it to the auth_state implicitly.

### Who would use this feature?
In my case this is useful so that I can take the id_token and convert it to temporary AWS credentials server side using something like this auth hook:

```python
def auth_state_hook(spawner, auth_state):
    id_token = auth_state['id_token']
    identity_pool_id = os.environ.get('IDENTITY_POOL_ID')
    cognito_provider = os.environ.get('COGNITO_PROVIDER_URL')[8:]

    cognito = boto3.client('cognito-identity')
    logins = { cognito_provider: id_token }

    # 1. Generate/retrieve pool identity
    identity_id = cognito.get_id(IdentityPoolId=identity_pool_id, Logins=logins)['IdentityId']

    # 2. Get temporary credentials for identity
    credentials = cognito.get_credentials_for_identity(IdentityId=identity_id, Logins=logins)

    # 3. Apply temporary credentials to notebook
    spawner.environment['AWS_ACCESS_KEY_ID'] = credentials['AccessKeyId']
    spawner.environment['AWS_SECRET_ACCESS_KEY'] = auth_state['SecretKey']
    spawner.environment['AWS_SESSION_TOKEN'] = auth_state['SessionToken']
```

manics:
Do you think we could add the entire `token_response` as a field of `auth_state`?
https://github.com/jupyterhub/oauthenticator/blob/3bcb7a7293ba99bae16eddb37e3177082a54234c/oauthenticator/generic.py#L138-L151

mahic:
@manics @michaeljfazio The OAuthenticator hasn't had any releases since 15.1. There has been a major refactoring in the subsystem since the last release, and to me it seems as some of the refactorings now include the id_token as well?

@GeorgianaElena @consideRatio can we expect some releases of this soon? It would help us out a lot :-)

GeorgianaElena:
[...]

PR #725, opened 2024-02-02
Add userdata_from_id_token as alternative to userdata_url

benjimin:
This PR adds support for extracting claims from the _id token_, instead of by exchanging the _access token_ with the _userinfo endpoint_. Addresses #487.

This is needed because some claims may only be available via the id token and not via the userinfo endpoint, for example, _AWS Cognito_ only provides the group membership claim via the id token (i.e., needed to complete support for #706).

For background, OAuth2 was a standard for obtaining an access token ("authorisation not authentication") and OIDC extended this in two ways: by standardising a userinfo endpoint and by accompanying the access token with an id token. (Note this is only the core of OIDC; _complete_ OIDC support also entails discovery mechanisms, etc.) To date, oauthenticator has always discarded the id token that it receives, and relied on the userinfo endpoint for performing authentication.

This PR requires the feature to be explicitly enabled by the admin, without introducing any additional config attribs. It uses a library which is already used [elsewhere](https://github.com/search?q=org%3Ajupyterhub+pyjwt&type=code) by JupyterHub.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -88,8 +88,6 @@ with open('requirements.txt') as f:
 
 
 setup_args['extras_require'] = {
-    # azuread is required for use of AzureADOAuthenticator
-    'azuread': ['pyjwt>=2'],
     # googlegroups is required for use of GoogleOAuthenticator configured with
     # either admin_google_groups and/or allowed_google_groups.
     'googlegroups': [
@@ -106,8 +104,6 @@ setup_args['extras_require'] = {
         'pytest-asyncio>=0.17,<0.23',
         'pytest-cov',
         'requests-mock',
-        # dependencies from azuread:
-        'pyjwt>=2',
         # dependencies from googlegroups:
         'google-api-python-client',
         'google-auth-oauthlib',
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -360,21 +360,36 @@ class OAuthenticator(Authenticator):
     def _token_url_default(self):
         return os.environ.get("OAUTH2_TOKEN_URL", "")
 
+    userdata_from_id_token = Bool(
+        False,
+        config=True,
+        help="""
+        Extract user details from an id token received via a request to
+        :attr:`token_url`, rather than making a follow-up request to the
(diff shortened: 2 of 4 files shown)


minrk:
Thank you! I think this is very sensible. Perhaps config would be clearer if we used an explicit special value, like e.g. `userdata_url = ".id_token"`, rather than encoding this behavior as `None`, or even `userdata_from_id_token = True` that takes precedence over userdata_url.

consideRatio:
I think userdata_from_id_token = true is a good choice here

> Thank you! I think this is very sensible. Perhaps config would be clearer if we used an explicit special value, like e.g. `userdata_url = ".id_token"`, rather than encoding this behavior as `None`, or even `userdata_from_id_token = True` that takes precedence over userdata_url.

I think having a dedicated toggle `userdata_from_id_token = True` would be good, it would also provide us with a clear place to document this behavior.

consideRatio:
Is pyjwr an optional dep for azuread ? Then we can clean it up there.

Also i think we should add a lower bound to pyjwts major version, so we depend on pyjwt>=2 because i recall migrating to v2 from v1 was needed elsewhere

minrk:
> Is pyjwt an optional dep for azuread?

[it is](https://github.com/jupyterhub/oauthenticator/blob/f26b4547af71d344f954a759a4074a31fda117c5/setup.py#L92), we can promote it to an unconditional dependency with the same bounds.

setup.py
```diff
@@ -88,8 +88,6 @@ def run(self):
 
 
 setup_args['extras_require'] = {
-    # azuread is required for use of AzureADOAuthenticator
-    'azuread': ['pyjwt>=2'],
```
consideRatio, line 92: Is removing this entirely going to cause a error for users of this, foe example tljh? Does it require  a major release?

minrk (approved):

manics:
Do you think we could add a test for this to help with future maintenance?

minrk:
@benjimin thanks for the test!

benjimin:
Thanks for the assist and support!

PR #729, opened 2024-02-09
add dedicated doc on details of allowing access

### LINKED ISSUE #727: Provide high level docs on how allow/disallow config works
This is the gist of it how allow/disallow config works is that **each user needs to**:

1. **Successfully authenticate**

   This is the phase where a username is determined, and misc properties of the user such as admin status.
   
   Failure to authenticate can stem from `GoogleOAuthenticator.hosted_domain` or `CILogonOAuthenticator.allowed_idps` config that influences the ability to determine the username.
  
2. **Not be blocked/disallowed**

   Only `Authenticator.blocked_users` is disallowing config
  
3. **Be allowed**

   Examples of such allowing config is: `Authenticator.allow_users`, `Authenticator.admin_users`, `OAuthenticator.allow_all`, `OAuthenticator.allow_existing_users`, `GitHubOAuthenticator.allowed_organizations`, ...)

minrk:
closes #727

### DIFF
--- a/docs/source/topic/allowing.md
+++ b/docs/source/topic/allowing.md
@@ -1,0 +1,143 @@
+(allowing)=
+
+# Allowing access to your JupyterHub
+
+OAuthenticator is about deferring **authentication** to an external source,
+assuming your users all have accounts _somewhere_.
+But many of these sources (e.g. Google, GitHub) have _lots_ of users, and you don't want _all_ of them to be able to use your hub.
+This is where **authorization** comes in.
+
+In OAuthenticator, authorization is represented via configuration options that start with `allow` or `block`.
+
+There are also lots of OAuth providers, and as a result, lots of ways to tell OAuthenticator who should be allowed to access your hub.
+
+## Default behavior: nobody is allowed!
+
+Assuming you have provided no `allow` configuration, the default behavior of OAuthenticator (starting with version 16) is to not allow any users unless explicitly authorized via _some_ `allow` configuration.
+If you want anyone to be able to use your hub, you must specify at least one `allow` configuration.
+
+```{versionchanged} 16
+Prior to OAuthenticator 16, `allow_all` was _implied_ if `allowed_users` was not specified.
+Starting from 16, `allow_all` can only be enabled explicitly.
+```
+
+## Allowing access
+
+There are several `allow_` configuration options, to grant access to users according to different rules.
+
+When you have only one `allow` configuration, the behavior is generally unambiguous: anyone allowed by the rule can login to the Hub, while anyone not explicitly allowed cannot login.
+However, once you start adding additional `allow` configuration, there is some ambiguity in how multiple rules are combined.
+
+```{important}
+Additional allow rules **can only grant access**, meaning they only _expand_ who has access to your hub.
+Adding an `allow` rule cannot prevent access granted by another `allow` rule.
+To block access, use `block` configuration.
+```
+
+That is, if a user is granted access by _any_ `allow` configuration, they are allowed.
+An allow rule cannot _exclude_ access granted by another `allow` rule.
+
+An example:
+
+```python
+c.GitHubOAuthenticator.allowed_users = {"mensah", "art"}
+c.GitHubOAuthenticator.allowed_organizations = {"preservation"}
+```
+
+means that the users `mensah` and `art` are allowed, _and_ any member of the `preservation` organization are allowed.
+Any user that doesn't meet any of the allow rules will not be allowed.
+
+| user  | allowed | reason                                                  |
+| ----- | ------- | ------------------------------------------------------- |
+| art   | True    | in `allowed_users`                                      |
+| amena | True    | member of `preservation`                                |
+| tlacy | False   | not in `allowed_users` and not member of `preservation` |
+
+### `allow_all`
+
+The first and simplest way to allow access is to any user who can successfully authenticate:
+
+```python
+c.OAuthenticator.allow_all = True
+```
+
+This is appropriate when you use an authentication provider (e.g. an institutional single-sign-on provider), where everyone who has an account in the provider should have access to your Hub.
+It may also be appropriate for unadvertised short-lived hubs, e.g. dedicated hubs for workshops that will be shutdown after a day, where you may decide it is acceptable to allow anyone who finds your hub to login.
+
+If `allow_all` is enabled, no other `allow` configuration will have any effect.
+
+```{seealso}
+Configuration documentation for {attr}`.OAuthenticator.allow_all`
+```
+
+### `allowed_users`
+
+This is top-level JupyterHub configuration, shared by all Authenticators.
+This specifies a list of users that are allowed by name.
+This is the simplest authorization mechanism when you have a small group of users whose usernames you know:
+
+```python
+c.OAuthenticator.allowed_users = {"mensah", "ratthi"}
+```
+
+If this is your only configuration, only these users will be allowed, no others.
+
+Note that any additional usernames in the deprecated `admin_users` configuration will also be added to the `allowed_users` set.
+
+```{seealso}
+Configuration documentation for {attr}`.OAuthenticator.allowed_users`
+```
+
+### `allow_existing_users`
+
+JupyterHub can allow you to add and remove users while the Hub is running via the admin page.
+If you add or remove users this way, they will be added to the JupyterHub database, but their ability to login will not be affected unless they are also granted access via an `allow` rule.
+
+To enable managing users via the admin panel, set
+
+```python
+c.OAuthenticator.allow_existing_users = True
+```
+
+```{warning}
+Enabling `allow_existing_users` means that _removing_ users from any explicit allow mechanisms will no longer revoke their access.
+Once the user has been added to the database, the only way to revoke their access to the hub is to remove the user from JupyterHub entirely, via the admin page.
+```
+
+```{seealso}
+Configuration documentation for {attr}`.OAuthenticator.allow_existing_users`
+```
+
+### provider-specific rules
+
+Each OAuthenticator provider may have its own provider-specific rules to allow groups of users access, such as:
+
+- {attr}`.CILogonOAuthenticator.allowed_idps`
+- {attr}`.GitHubOAuthenticator.allowed_organizations`
+- {attr}`.GitLabOAuthenticator.allowed_gitlab_groups`
+- {attr}`.GlobusOAuthenticator.allowed_globus_groups`
+- {attr}`.GoogleOAuthenticator.allowed_google_groups`
+
+## Blocking Access
+
+It's possible that you want to limit who has access to your Hub to less than all of the users granted access by your `allow` configuration.
+`block` configuration always has higher priority than `allow` configuration, so if a user is explicitly allowed _and_ explicitly blocked, they will not be able to login.
+
+The only `block` configuration is the base Authenticators `block_users`,
+a set of usernames that will not be allowed to login.
+
+### Revoking previously-allowed access
+
+Any users who have logged in previously will be present in the JupyterHub database.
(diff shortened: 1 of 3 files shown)


oauthenticator/oauth2.py
```diff
@@ -1086,3 +1088,20 @@ def __init__(self, **kwargs):
...
             )
         super().__init__(**kwargs)
+
+
+# patch allowed_users help string to match our definition
+# base Authenticator class help string gives the wrong impression
+# when combined with other allow options
+OAuthenticator.class_traits()[
+    "allowed_users"
+].help = """
```
minrk, line 1098: The base Authenticator.allowed_users docstring gives the wrong impression of how it relates to other `allow` config, since the base Authenticator doesn't have other `allow` config.

docs/source/topic/allowing.md
```diff
@@ -0,0 +1,143 @@
...
+
+```{seealso}
+Configuration documentation for {attr}`.OAuthenticator.allow_existing_users`
+```
+
+### provider-specific rules
+
+Each OAuthenticator provider may have its own provider-specific rules to allow groups of users access, such as:
+
+- {attr}`.CILogonOAuthenticator.allowed_idps`
```
minrk, line 115: This list is not comprehensive. Should it be? I'm not sure! It will certainly get out-of-date if it's supposed to be.

minrk:
It's unclear to me a bit whether things like allowed_idps/hosted_domain expanations belong here vs the provider-specific doc pages

docs/source/topic/allowing.md
manics: I think it's fine for this to be a non-comprehensive set of examples.
We've got provider specific pages e.g. https://oauthenticator.readthedocs.io/en/latest/tutorials/provider-specific-setup/providers/gitlab.html
In future I think it'd be nice to put more explanations on those pages, so instead of just listing the additional config properties we could add some text to tie it in with the explanation on this page.

docs/source/topic/allowing.md
```diff
@@ -0,0 +1,143 @@
...
+- {attr}`.GitHubOAuthenticator.allowed_organizations`
+- {attr}`.GitLabOAuthenticator.allowed_gitlab_groups`
+- {attr}`.GlobusOAuthenticator.allowed_globus_groups`
+- {attr}`.GoogleOAuthenticator.allowed_google_groups`
+
+## Blocking Access
+
+It's possible that you want to limit who has access to your Hub to less than all of the users granted access by your `allow` configuration.
+`block` configuration always has higher priority than `allow` configuration, so if a user is explicitly allowed _and_ explicitly blocked, they will not be able to login.
```
manics, line 124:
```suggestion
`block` configuration always has higher priority than `allow` configuration, so if a user is both allowed _and_ blocked, they will not be able to login.
```

docs/source/topic/allowing.md
```diff
@@ -0,0 +1,143 @@
...
+a set of usernames that will not be allowed to login.
+
+### Revoking previously-allowed access
+
+Any users who have logged in previously will be present in the JupyterHub database.
+Removing a user's login permissions (e.g. removing them from a GitLab project when using {attr}`.GitLabOAuthenticator.project_ids`) only prevents future logins;
+it does not remove the user from the JupyterHub database.
+This means that:
+
+1. any API tokens, that the user still has access to will continue to be valid, and can continue to be used, and
```
manics, line 136:
```suggestion
1. any API tokens that the user still has access to will continue to be valid, and can continue to be used
```

oauthenticator/oauth2.py
```diff
@@ -291,9 +293,9 @@ class OAuthenticator(Authenticator):
 
         .. warning::
 
-           When this is enabled and you are to remove access for one or more
-           users allowed via other config options, you must make sure that they
-           are not part of the database of users still. This can be tricky to do
+           When this is enabled and you wish to remove access for one or more
+           users previously allowed, you must make sure that they
+           are not removed from the jupyterhub database. This can be tricky to do
```
manics, line 298:
```suggestion
           are removed from the jupyterhub database. This can be tricky to do
```

oauthenticator/oauth2.py
```diff
@@ -1086,3 +1088,20 @@ def __init__(self, **kwargs):
...
+# base Authenticator class help string gives the wrong impression
+# when combined with other allow options
+OAuthenticator.class_traits()[
+    "allowed_users"
+].help = """
+Set of usernames that should be allowed to login.
+
+If unspecified, grants no access.
+
+At least one `allow` configuration must be specified
```
manics, line 1103:
```suggestion
If unspecified, grants no access. You must set at least one other `allow` configuration
```

oauthenticator/oauth2.py
manics: Lower down in this help message there's a note
```
        .. note::
           Allowing existing users is done by adding existing users on startup
           and newly created users to the `allowed_users` set. Due to that, you
           can't rely on this config to independently allow existing users if
           you for example would reset `allowed_users` after startup.

        .. versionadded:: 16.0

        .. versionchanged:: 16.0
```
After reading it several times I still don't understand what the second sentence means. Do you have any thoughts on how to make it clearer?

manics (commented):
A few suggestions, looks good to me!

oauthenticator/oauth2.py
consideRatio: Its probably me trying to convey the hidden coupling of allow_existing_users and allowed_users, which could come as a surprise if looking at allowed_users after startup.

The mechanism for allow_existing_users to allow existing users is through appending users to allowed_users. I wanted this complexity surfaced to be an answer to someone unexpectedly running into confusion about this in some way - maybe via functionality in https://github.com/jupyterhub/jupyterhub/pull/4442

docs/source/topic/allowing.md
consideRatio: Can you use another example than Cilogons allowed_idps? It isn't really an allow config like the others, but a requirement to be from an idp listed.

See https://github.com/jupyterhub/oauthenticator/issues/683 and https://github.com/jupyterhub/oauthenticator/pull/685

consideRatio:
> It's unclear to me a bit whether things like allowed_idps/hosted_domain expanations belong here vs the provider-specific doc pages.

Those two are odd ducks, they shouldn't be classified as allow config. They are requirements to auth currently, and hosted_domain could move to be part of the blocked check if the implementation is refactored.

See https://github.com/jupyterhub/oauthenticator/issues/683

GeorgianaElena (approved):
This is great to have!

docs/source/topic/allowing.md
minrk: removed `allowed_idps`

docs/source/topic/allowing.md
```diff
@@ -0,0 +1,142 @@
...
+In OAuthenticator, authorization is represented via configuration options that start with `allow` or `block`.
+
+There are also lots of OAuth providers, and as a result, lots of ways to tell OAuthenticator who should be allowed to access your hub.
+
+## Default behavior: nobody is allowed!
+
+The default behavior of OAuthenticator (starting with version 16) is to block all users unless explicitly authorized via _some_ `allow` configuration.
+If you want anyone to be able to use your hub, you must specify at least one `allow` configuration.
+
+```{versionchanged} 16
```
consideRatio, line 19: This isn't accurate as it is written now, if for example github configured allowed_orgsnizations without configuring allowed_users, it would mean that allow_all still wasn't implied.

Rather whats accurate (i believe) is that allow_all was implied if no other config starting with allow was specified.

minrk: Right, it should be "no other `allow` config", not just `allowed_users`. Got my JupyterHub and OAuthenticator mixed up.

docs/source/topic/allowing.md
```diff
@@ -0,0 +1,142 @@
...
+This specifies a list of users that are allowed by name.
+This is the simplest authorization mechanism when you have a small group of users whose usernames you know:
+
+```python
+c.OAuthenticator.allowed_users = {"mensah", "ratthi"}
+```
+
+If this is your only configuration, only these users will be allowed, no others.
+
+Note that any additional usernames in the deprecated `admin_users` configuration will also be added to the `allowed_users` set.
```
consideRatio, line 85: Maybe no chance is needed, but they aren't added to the set right? They are simply passing the check_allowed, as admin_users lead to getting admin=true in the authentication dict, and that makes them pass check_allowed in oauthenticstor. So for users of google_admin_groups, they are also granted access like this for example.

docs/source/topic/allowing.md
```diff
@@ -0,0 +1,142 @@
...
+```{warning}
+Enabling `allow_existing_users` means that _removing_ users from any explicit allow mechanisms will no longer revoke their access.
+Once the user has been added to the database, the only way to revoke their access to the hub is to remove the user from JupyterHub entirely, via the admin page.
+```
+
+```{seealso}
+Configuration documentation for {attr}`.OAuthenticator.allow_existing_users`
+```
+
+### provider-specific rules
```
consideRatio, line 111:
```suggestion
### Provider-specific rules
```

docs/source/topic/allowing.md
minrk: I actually thought they were, and they are, [_indirectly_](https://github.com/jupyterhub/jupyterhub/blob/2c932997643c631f3a7874ab39ee51f6e040622d/jupyterhub/app.py#L2132) via `.add_user()`. `allow_existing_users` changes this from the base implementation, and this makes me think we may need an update for https://github.com/jupyterhub/jupyterhub/pull/4701 to actually make this explicit.

docs/source/topic/allowing.md
```diff
@@ -0,0 +1,142 @@
...
+Removing a user's login permissions (e.g. removing them from a GitLab project when using {attr}`.GitLabOAuthenticator.project_ids`) only prevents future logins;
+it does not remove the user from the JupyterHub database.
+This means that:
+
+1. any API tokens that the user still has access to will continue to be valid, and can continue to be used
+2. any still-valid browser sessions will continue to be logged in.
+
+```{important}
+To fully remove a user's access to JupyterHub,
+their login permission must be revoked _and_ their User fully deleted from the Hub,
```
consideRatio, line 140:
```suggestion
their login permission must be revoked _and_ their user fully deleted from the Hub,
```

Not sure about this suggestion, maybe its right to capitalize user ?

minrk: I definitely over-capitalize. I probably used User because of the ORM (i.e. `db.delete(orm.User)` is specifically what's needed), but that's not entirely relevant for docs like this. lowercase is probably better.

oauthenticator/oauth2.py
```diff
@@ -1086,3 +1080,18 @@ def __init__(self, **kwargs):
...
+# when combined with other allow options
+OAuthenticator.class_traits()[
+    "allowed_users"
+].help = """
+Set of usernames that should be allowed to login.
+
+If unspecified, grants no access. You must set at least one other `allow` configuration
+if any users are to have permission to access the Hub.
+
+Any users in `admin_users` will be added to this set.
```
consideRatio, line 1096: I made another comment about this, i don't think this is correct.

docs/source/topic/allowing.md
minrk: > I think i propose a less specific claim is made, where admin status, for example via admin_userd, also leads to being allowed.

👍

oauthenticator/oauth2.py
minrk: Good call, done.

consideRatio:
I'll reflect further on https://github.com/jupyterhub/oauthenticator/pull/729#discussion_r1485885921, but i think there isn't an action point remaining in this PR about it - this LGTM and can be merged.

@minrk - ok to merge?

consideRatio (approved):
Thank you @minrk!!!

minrk:
Thanks for the thoughtful reviews!

PR #730, opened 2024-02-09
add example for deploying with mock-oauth2-server

minrk:
I found this quite useful for testing

We may want to integrate it with our actual tests at some point, but the example is enough to be useful.

This lets JupyterHub be deployed with a full 'real' OAuth provider without any external requirements.

### DIFF
--- a/examples/mock-provider/README.md
+++ b/examples/mock-provider/README.md
@@ -1,6 +1,6 @@
 # Generic OAuth with mock provider
 
-This example uses [mock-oauth2-server][] to launch a standalone local OAuth2 provider and configures GenericOAuthenticator to use it.
+This example uses [mock-oauth2-server] to launch a standalone local OAuth2 provider and configures GenericOAuthenticator to use it.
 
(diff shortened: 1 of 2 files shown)


examples/mock-provider/README.md
```diff
@@ -0,0 +1,21 @@
+# Generic OAuth with mock provider
+
+This example uses [mock-oauth2-server][] to launch a standalone local OAuth2 provider and configures GenericOAuthenticator to use it.
```
consideRatio, line 3:
```suggestion
This example uses [mock-oauth2-server] to launch a standalone local OAuth2 provider and configures GenericOAuthenticator to use it.
```

consideRatio (approved):
Nice!!!

examples/mock-provider/README.md
minrk: I always forget you can do that. Thanks!

consideRatio: I don't know if this works for all markdown types, so lets ignore this suggestion maybe

PR #732, opened 2024-03-17
Add changelog for 16.3.0

consideRatio:
I saw that a release was wanted [in discourse](https://discourse.jupyter.org/t/timeline-for-new-oauthenticator-release/24538/2) by @krassowski and figured we could go for it.

I've opted to propose the version as 16.3.0 instead of 17.0.0 as part of requiring jupyterhub 2.2+ instead of jupyterhub 1.2+. My motivation for not going 17.0.0 is that I'd like to not spend users attention unless its likeley they require to take an action, and I think its unlikeley users are using the latest version of oauthenticator with JupyterHub 1 still. I've however added an "important" admonition about now requring jupyterhub 2.2 just in case.

If someone still use and pin jupyterhub 1 and wants to install this version, I think `pip` will write out that the requested things can't resolve properly - so maybe its acceptable to not go for 17.0.0?

image

### DIFF
--- a/docs/source/reference/changelog.md
+++ b/docs/source/reference/changelog.md
@@ -6,6 +6,48 @@ command line for details.
 
 ## [Unreleased]
 
+### 16.3.0 - 2024-03-20
+
+```{important}
+This release includes a security patch for {attr}`.GoogleOAuthenticator.hosted_domain`,
+see [GHSA-55m3-44xf-hg4h] for details.
+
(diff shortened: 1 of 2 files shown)


consideRatio:
I've added the release date for tuesday next week, hoping that this PR is reviewed by then.

krassowski (approved):
Releasing this as a minor version looks good to me - if someone is pinning to older `jupyterhub`, they will just not get this new version.

PR #739, opened 2024-04-05
gh: Allow setting JupyterHub groups from GitHub teams

yuvipanda:
Looking at https://github.com/jupyterhub/oauthenticator/pull/735, I realize it will not actually allow us to use GitHub teams or orgs as JupyterHub groups, since that's an extra API call. This PR keeps it simple, and adds config to pick this up from teams list directly.


yuvipanda:
I don't actually want to do this, see https://github.com/jupyterhub/oauthenticator/pull/735#issuecomment-2040799664 for more information.

yuvipanda:
We decided against going this way https://github.com/jupyterhub/oauthenticator/pull/735#issuecomment-2040799664

PR #741, opened 2024-04-25
Move allowed_groups and admin_groups to base authenticator

yuvipanda:
A simplification of https://github.com/jupyterhub/oauthenticator/pull/735, moving 2 of the 3 traitlets. This is a straight up move, without any functional breaking changes.

- `admin_groups` allows setting members of some groups as admins.
- `allowed_groups` allows setting what groups should be allowed to login.

Both of these are more useful with claim_groups_key, as that allows an *external* party to drive group memberships. Without that, I guess primarily this depends on membership within the JupyterHub admin UI.

Splitting this up helps us get this moving faster, as figuring out how to move `claim_groups_key` is going to be slightly more involved.

### DIFF
--- a/oauthenticator/generic.py
+++ b/oauthenticator/generic.py
@@ -8,7 +8,7 @@ from functools import reduce
 from jupyterhub.auth import LocalAuthenticator
 from jupyterhub.traitlets import Callable
 from tornado.httpclient import AsyncHTTPClient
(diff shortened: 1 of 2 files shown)


yuvipanda:
I was wrong, looks like it's not actually easier to move just these two - I thought `get_user_groups` was common to all authenticators, but not the case.

PR #744, opened 2024-06-11
Add changelog for v16.3.1

yuvipanda:
While trying to get a fix out for https://github.com/jupyterhub/oauthenticator/security/advisories/GHSA-gprj-3p75-f996, instead of making a backported release, I released from main. This was a mistake, but not a costly one. As I start doing more releases, I'll get better.

I could not get github-activity to work, so I manually constructed the changelog to the best of my ability. I reported the issue upstream in https://github.com/executablebooks/github-activity/issues/96.


yuvipanda:
Given that I made the release already, I'll say 'mea culpa' and merge this myself, since I want people to be able to see a changelog when they look. I'll get better at making releases, I swear.

minrk:
Sorry for the trouble, and thanks for making the release!

PR #748, opened 2024-07-26
[Google] Add switch to strip domain from username

### LINKED ISSUE #733: [Google] Decouple `hosted_domain` from domain stripping usernames
Currently, if `GoogleOAuthenticator.hosted_domain` is configured to a list with a single domain, the username is stripped of that domain. If another hosted_domain is added, no domains are stripped from the usernames etc.

This is a fragile configuration, where a change to `hosted_domain` for authorization purposes leads to changes in usernames. We should make sure those changes are decoupled from each other I think.

This isn't an urgent issue or something we need to fix, but I wanted it represented as a GitHub issue for now.

0mar:
Taking a stab at #733. 

I added a switch to the Google OAuthenticator to explicitly indicate whether domains should be stripped, which by default is set to False. If set to True, domain names will be stripped, and if in addition there are multiple domains specified, a warning will be produced. Perhaps this should be an error instead?

Closes #733

### DIFF
--- a/oauthenticator/google.py
+++ b/oauthenticator/google.py
@@ -118,7 +118,7 @@ class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
 
     @default('strip_domain')
     def _strip_if_single_domain(self):
-        return len(self.hosted_domain) > 1
+        return len(self.hosted_domain) <= 1
 
     @validate('strip_domain')
     def _check_multiple_hosted_domain(self, strip_domain):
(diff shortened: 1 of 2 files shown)


manics:
Thanks for working on this. At the moment it's going to silently break most existing users who only have a single domain. Even if we highlight it in a changelog it's likely to be missed.

I can think of two options:
- never strip the domain if there are multiple domains, so this new parameter only affects single domains and defaults to the current behaviour.
- use a dynamic default (you can use a traitlets default method to set this) so if it's not set by the admin the behaviour is unchanged from the current.

0mar:
Thanks for the feedback. The second option sounds like it covers the most use cases, so I've gone with that one

GeorgianaElena (approved):
This looks good to me @0mar! Thank you for fixing it.

PR #751, opened 2024-08-23
add OAuthenticator.modify_auth_state_hook, allow get_user_groups / auth_state_groups_key to be async

### LINKED ISSUE #750: Managing group information properly when tokens don't include groups claim key?
We are upgrading a custom authenticator based on `GenericOAuthenticator` from 15.x to 16.x (or more likely, the currently in-development 17.0).  We need to adapt our usage of `manage_groups` and `allowed_groups`.  At this time our tokens do not include a groups claim key, so we've been getting groups information from a separate API call made from an overridden `authenticate()`.  (Indeed, we see the warnings in the comments about that being the wrong way to do it, and we are working towards getting away from doing that.) While we see that a callable can be used to set the groups information in `get_user_groups`, it isn't async, nor is `get_user_groups`, etc.  We're exploring options to get the groups info we needed added upstream, but in the meantime we're wondering if you can recommend a way for us to customize groups management?  Or is it feasible to make the call to `get_user_groups` non-blocking?

minrk:
closes #750 

it feels a little weird to have `auth_state_groups_key` be a callable that actually fetches external info purely based on the config name, but it's what we have. Is there a better arrangement for adding more fields to the user model? e.g. something after `token_to_user`? I noted in #750 that _extending_ token_to_user would work. Should we have a _hook_ after token_to_user to allow additional modifications to the user model before things like groups get added, or is making this method/hook async enough?

### DIFF
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -368,6 +368,28 @@ class OAuthenticator(Authenticator):
         """,
     )
 
+    modify_auth_state_hook = Callable(
+        config=True,
+        default_value=None,
+        allow_none=True,
+        help="""
+        Callable to modify `auth_state`.
+
+        Will be called with the Authenticator instance and the existing auth_state dictionary
+        and must return the new auth_state dictionary:
+
+        ```
+        auth_state = [await] modify_auth_state_hook(authenticator, auth_state)
+        ```
+
+        This hook is called _before_ populating group membership,
+        so can be used to make additional requests to populate additional fields
+        which may then be consumed by `auth_state_groups_key` to populate groups.
+
+        This hook may be async.
+        """,
+    )
+
     @observe("allowed_groups", "admin_groups", "auth_state_groups_key")
     def _requires_manage_groups(self, change):
         """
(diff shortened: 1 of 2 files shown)


yuvipanda:
ya, how about an `augment_auth_state_hook` that gets called after authenticate, and can put random stuff in `auth_state`? Then we can leave the existing group management stuff as is, and replace all the 'override authenticator, override authenticate just to put stuff in `auth_state`' with that one hook.

yuvipanda:
It can take the authenticator instance and existing auth_state as params, and return the augmented auth_state

minrk:
Added `modify_auth_state_hook` config, run after `build_auth_state` but before `update_auth_model`, so it can modify the inputs to `get_user_groups`.

I left the async allowance in `get_user_groups`, since I think that makes sense generally, but now 'the way to do this' would be to use the hook (user) / build_auth_state (subclass).

yuvipanda (commented):
Generally looks good, one specific question about exception handling mechanism.

minrk:
I don't see your question, did I miss something?

oauthenticator/oauth2.py
```diff
@@ -1136,6 +1180,18 @@ async def update_auth_model(self, auth_model):
...
+    async def _call_modify_auth_state_hook(self, auth_state):
+        """Call the modify_auth_state_hook"""
+        try:
+            auth_state = self.modify_auth_state_hook(self, auth_state)
+        except Exception as e:
+            # let hook errors raise, nothing in auth should suppress errors
+            self.log.error(f"Error in modify_auth_state_hook: {e}")
+            raise
+        if isawaitable(auth_state):
+            auth_state = await auth_state
```
yuvipanda, line 1192: Wouldn't the exception be thrown here on the `await` than earlier? Wouldn't that require us to be inside the `try`? I could be wrong...

yuvipanda:
oops, something ate it! I've asked again.

oauthenticator/oauth2.py
minrk: You're 100% right, yes! Fixed.

yuvipanda:
Thanks, @minrk

PR #757, opened 2024-09-02
update docstrings on relationship between `allowed_groups` and `allow_[provider_group_equivalents]`

### LINKED ISSUE #756: New group config in OAuthenticator overlaps with `allowed_google_groups` for example and may not work - what do we do?
#735 moved group related config in Generic to OAuthenticator, and did a breaking change related to that group config previosly found only in GenericOAuthenticator.

<details><summary>Documentation about breaking change from #735</summary>

> ## Backwards compatibility
> * The existing `claim_groups_key` behavior is preserved, by being passed on to `auth_state_groups_key` in the base. It has been marked as deprecated. This is not a backwards compatibility break.
> * All group related functionality now **requires** `manage_groups` to be `True`, which was not the case earlier. Before this, if `manage_groups` is false but any of the group related authorization functionality (`allowed_groups` and `admin_groups`) is used, they control group related behavior **but don't show up as JupyterHub groups**. This causes confusion, as the 'groups' field in the admin panel will be empty, and possible other group related behavior (such as future profile list filtering, for example) would _not_ respect these groups. We basically would end up with _two_ group concepts - First class JupyterHub groups (which will show up in admin panel, API, can be _edited_ by admins, etc) as well as second class 'Authenticator' groups (which are only used for authorization and 'disappear' after that). I think this is unnecessary complication, and this is a good time to remove this distinction. Now, any kind of group related authorization functionality _requires_ `manage_groups` to be `True`, and we are back to having only one notion of 'group'. We also remove the confusing part where you may have `allowed_groups` set to something, manually modify the groups the user is a part of in JupyterHub admin, and it silently has no effect. This is a _breaking change_ for people who used groups functionality but set `manage_groups` to be `False`. However, I think that usage is fairly minor, because of the confusing behavior it causes. I have added the 'breaking' label here regardless.
> 
> ## Breaking change
> * All group related functionality (`allowed_groups`, `admin_groups`, `claims_group_key` and `auth_state_groups_key`) now also requires `manage_groups` to be set to `True`

</details>

We have two situations to address in my mind:

- Some classes have two sets of group config without guidance on whats supported and recommended to use.

  For example, these configs are available side by side with `allowed_groups` now.

  `GitHubOAuthenticator.allowed_organizations`
  `GitLabOAuthenticator.allowed_gitlab_groups`
  `GlobusOAuthenticator.allowed_globus_groups`
  `GoogleOAuthenticator.allowed_google_groups`
- New set of group config may not work consistently across authenticator classes

  Searching for `def update_auth_model` in the code base, it seems we typically don't call the base class implementation, making logic required for the new set of group config.
  
  Tests were updated in Generic and OpenShift, but not in other classes such as GitHub, GitLab, Globus, Google.

minrk:
should clarify that they do not conflict, and are used in different circumstances.

Adds more links between config and tries to clarify that JupyterHub groups != provider groups/teams/organizations/etc.

closes #756

### DIFF
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -334,9 +334,17 @@ class OAuthenticator(Authenticator):
         Unicode(),
         config=True,
         help="""
-        Allow members of selected groups to log in.
+        Allow members of selected JupyterHub groups to log in.
 
-        Requires `manage_groups` to also be `True`.
+        Requires :attr:`manage_groups` to also be `True`.
+        Typically also requires :attr:`auth_state_groups_key` to be configured to populate the JupyterHub groups.
+        
+        This option is *independent* of other configuration such as :attr:`.GitLabOAuthenticator.allowed_gitlab_groups`,
+        which do not populate the *JupyterHub* groups,
+        and do not require :attr:`manage_groups` to be True.
+        
+        .. versionadded:: 17
+            Previously available only on :class:`.GenericOAuthenticator`
         """,
     )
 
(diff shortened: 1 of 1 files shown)


consideRatio:
Thank you @minrk!!!

PR #763, opened 2024-09-22
[Google] Add `include_nested_groups` config

### LINKED ISSUE #762: Allow authentication via nested groups for Google OAuthenticator
### Proposed change
Allow authentication via nested groups for Google OAuthenticator. Google Workspace now allows nested Google Groups, but the Google OAuthenticator has no option for allowing auth this way.

This would be helpful for those who manage the IAM for many hubs.

### Suggest a solution
I have a fork that implements this feature. The particular issues that need to be discussed are:

- backwards compatibility: ensuring those who upgrade don't all of the sudden have users who can login due to unanticipated group structures
- security: see backwards compatibility
- login experience: currently `_fetch_user_groups` has a `FIXME` related to it being non-asynchronous. Adding support for nested groups may exacerbate this a bit as it takes longer to determine membership given the Google API's current limitations

Most of this can be solved by required explicit opt-in with:
```python
c.GoogleOAuthenticator.allow_nested_groups = True
```

If there's interest I can spruce up what I have and put in a PR for further discussion.

jrdnbradford:
First go at this, closes #762.

`_fetch_user_groups` is now `_fetch_member_groups` since it now checks user _and_ group memberships in groups. I've updated some of the variable names from `user_*` to `member_*` to reflect this. Whether nested groups are taken into account is set by a new `allow_nested_groups` config option.

The Google Python APIs for this are quite limited and this causes some slowness in the login experience for those who are members of a lot of groups. 

I couldn't find where to update the [docs for this authenticator](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.google.html#module-oauthenticator.google). When I try to suggest an edit on that page it goes to `reference/api/gen/oauthenticator.google.rst` which doesn't exist, I assume it's built elsewhere. If there's anywhere I can update this, let me know.

I am also not entirely sure how to go about adding a mock test for this case.

I'd love feedback on how to make this better. Thanks! 🚀

### DIFF
--- a/oauthenticator/google.py
+++ b/oauthenticator/google.py
@@ -391,7 +391,7 @@ class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
 
         resp = self.service.groups().list(userKey=user_email).execute()
         user_groups = {
-            g['email'].split('@')[0] for g in resp.get('groups', [{'email': None}])
+            g['email'].split('@')[0] for g in resp.get('groups', []) if g.get('email')
         }
 
         # Recursively check for nested groups if allowed
--- a/oauthenticator/google.py
+++ b/oauthenticator/google.py
@@ -363,6 +363,7 @@ class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
             cache_discovery=False,
             http=http,
         )
+
     def _setup_service(self, user_email_domain, http=None):
         """
         Set up the service client for Google API.
@@ -379,7 +380,14 @@ class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
         )
         return service
 
-    def _fetch_user_groups(self, user_email, user_email_domain, http=None, checked_groups=None, processed_groups=None):
+    def _fetch_user_groups(
+        self,
+        user_email,
+        user_email_domain,
+        http=None,
+        checked_groups=None,
+        processed_groups=None,
+    ):
         """
         Return a set with the google groups a given user is a member of, including nested groups if allowed.
         """
@@ -398,22 +406,24 @@ class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
 
         self.log.debug(f"Fetched groups for {user_email}: {user_groups}")
 
-        # Add the current user's groups to the checked groups
         checked_groups.update(user_groups)
-
         self.log.debug(f"Checked groups after update: {checked_groups}")
 
-        # Recursively check for nested groups if allowed
         if self.allow_nested_groups:
             for group in user_groups:
                 if group not in processed_groups:
-                    nested_groups = self._fetch_user_groups(group + "@" + user_email_domain, user_email_domain, http, checked_groups, processed_groups)
+                    nested_groups = self._fetch_user_groups(
+                        f"{group}@{user_email_domain}",
+                        user_email_domain,
+                        http,
+                        checked_groups,
+                        processed_groups,
+                    )
                     checked_groups.update(nested_groups)
                     processed_groups.add(group)
 
         self.log.debug(f"user_email {user_email} is a member of {checked_groups}")
         return checked_groups
-
 
 
 class LocalGoogleOAuthenticator(LocalAuthenticator, GoogleOAuthenticator):
--- a/oauthenticator/google.py
+++ b/oauthenticator/google.py
@@ -106,7 +106,6 @@ class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
     )
 
     allow_nested_groups = Bool(
-        Set(Unicode()),
         config=True,
         help="""
         Allow members of nested Google groups to sign in.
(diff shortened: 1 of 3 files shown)


oauthenticator/google.py
```diff
@@ -105,6 +105,21 @@ def _userdata_url_default(self):
...
+    allow_nested_groups = Bool(
+        config=True,
+        help="""
+        Allow members of nested Google groups to sign in and/or administer
+        TLJH. If `True` the authenticator will recursively check a user's 
+        group memberships and allow sign in and/or grant admin rights to
+        a user that is a member of a group that has nested membership in
+        any group in `allowed_google_groups` or `admin_google_groups`.
+
+        Use of this requires configuration of `gsuite_administrator`, and
```
consideRatio, line 117: Lets trim away some docs here, letting the groups config describe these requirements about service account keys etc, and settle by saying that this is a config that makes the config options admin/allowed groups accept the listed groups' nested groups as well.

Or, is this an additional requirement that isnt a requirement on admin groups / allowed groups?

oauthenticator/google.py
```diff
@@ -105,6 +105,21 @@ def _userdata_url_default(self):
         """,
     )
 
+    allow_nested_groups = Bool(
+        config=True,
+        help="""
+        Allow members of nested Google groups to sign in and/or administer
+        TLJH. If `True` the authenticator will recursively check a user's
```
consideRatio, line 112: Tljh mentioned but shouldnt be, as it can be used for any jupyterhub deployment, tljh/z2jh/no-dostribution etc

jrdnbradford: Found it's way in via osmosis so to speak, as I used TLJH as my main testing. Will remove!

jrdnbradford:
@consideRatio I have added some mock tests for the nested functionality.

oauthenticator/google.py
```diff
@@ -373,13 +376,51 @@ def _fetch_user_groups(self, user_email, user_email_domain, http=None):
...
+                if group not in processed_groups:
+                    processed_groups.add(group)
+                    nested_groups = self._fetch_member_groups(
+                        f"{group}@{user_email_domain}",
+                        user_email_domain,
+                        http,
+                        checked_groups,
+                        processed_groups,
+                    )
+                    checked_groups.update(nested_groups)
```
consideRatio, line 420: I've come to appreciate de-nesting code like this, but its just an opinionated refactor we don't need to do.

```suggestion
                if group in processed_groups:
                    continue
                processed_groups.add(group)
                nested_groups = self._fetch_member_groups(
                    f"{group}@{user_email_domain}",
                    user_email_domain,
                    http,
                    checked_groups,
                    processed_groups,
                )
                checked_groups.update(nested_groups)
```

oauthenticator/google.py
```diff
@@ -373,13 +376,51 @@ def _fetch_user_groups(self, user_email, user_email_domain, http=None):
...
         )
+        return service
+
+    def _fetch_member_groups(
+        self,
+        member_email,
+        user_email_domain,
+        http=None,
+        checked_groups=set(),
+        processed_groups=set(),
```
consideRatio, line 387: I think you may be required to have this be `..._groups=None`, because I think otherwise its a risk that the default value is re-used when it shouldn't be.

Googling made me find [this response](https://www.reddit.com/r/learnpython/comments/118ucmz/why_do_python_functions_reuse_pointer_to_mutable/#:~:text=Functions%20are%20first%20class%20objects,uses%20that%20same%20default%20argument.:):

>Functions are first class objects in python, and the default argument is set during the function definition, not when its called.
>
>So the empty list is getting assigned as the default argument on function definition.

So I think indeed, this has to be assigned `None`, and if/when needed, initialized within the function to an empty set.

oauthenticator/google.py
```diff
@@ -105,6 +105,14 @@ def _userdata_url_default(self):
         """,
     )
 
+    allow_nested_groups = Bool(
```
consideRatio, line 108: Can we make this `include_nested_groups` after all? I've gone back and forth in my mind about this, but I recall you also said you thought it could make sense.

I think my concluding motivation for making this be `include` is that it affects more than the `allowing` part, such as being declared an admin if found in a `admin_google_groups` and if the group is to be listed in auth state, which in turn currently or in the future (i don't remember current status) may be used to influence jupyterhub groups if `manage_groups` is set true.

jrdnbradford:
Awesome, thanks @consideRatio! Let me make these changes and test. I'll request another review when ready. Also, don't feel pressure to get this merged. No rush.

jrdnbradford:
Changes made. Thanks for your time and help on this, @consideRatio.

consideRatio (approved):
The string splitting at the @ character and keeping track of a user domain etc is a bit confusing, but it was confusing before this PR was made as well though. Lets go for a merge though, i think this should be an improvement.

consideRatio:
Thank you for working this thoroughly @jrdnbradford!!

PR #764, opened 2024-09-26
[Google] Make looking up google groups far less blocking

### LINKED ISSUE #607: The google authenticator's calls to check groups is done sync, can it be made async?
Quickly written, I can fill this in in more detail later.

jrdnbradford:
Closes #607 and should only be merged after #763 as it is stacked. There might be some conflicts that I will resolve.

The main change was removing the creation of the `service` and instead retrieving an oauth token and using an asynchronous `aiohttp` session to get the groups. If Google ever releases an asynchronous client that uses the `service` then we should switch to that because it's more readable.

### DIFF
--- a/oauthenticator/google.py
+++ b/oauthenticator/google.py
@@ -366,17 +366,17 @@ class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
         http=None,
         checked_groups=None,
         processed_groups=None,
+        credentials=None,
     ):
         """
         Return a set with the google groups a given user/group is a member of, including nested groups if allowed.
         """
-        if not hasattr(self, 'credentials'):
-            self.credentials = await self._setup_credentials(user_email_domain)
 
+        credentials = credentials or await self._setup_credentials(user_email_domain)
         checked_groups = checked_groups or set()
         processed_groups = processed_groups or set()
 
-        headers = {'Authorization': f'Bearer {self.credentials.token}'}
+        headers = {'Authorization': f'Bearer {credentials.token}'}
         url = f'https://www.googleapis.com/admin/directory/v1/groups?userKey={member_email}'
         group_data = await self.httpfetch(
             url, headers=headers, label="fetching google groups"
--- a/oauthenticator/google.py
+++ b/oauthenticator/google.py
@@ -14,7 +14,7 @@ from .oauth2 import OAuthenticator
 
 class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
     user_auth_state_key = "google_user"
-    _credentials = None
+    _service_credentials = None
 
     @default("login_service")
     def _login_service_default(self):
@@ -315,27 +315,29 @@ class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
         # users should be explicitly allowed via config, otherwise they aren't
         return False
 
-    def _get_credentials(self, user_email_domain):
+    def _get_service_credentials(self, user_email_domain):
         """
         Returns the stored credentials or fetches and stores new ones.
 
         Checks if the credentials are valid before returning them. Refreshes
         if necessary and stores the refreshed credentials.
         """
-        if not self._credentials or not self._is_token_valid():
-            self._credentials = self._setup_credentials(user_email_domain)
+        if not self._service_credentials or not self._is_token_valid():
+            self._service_credentials = self._setup_service_credentials(
+                user_email_domain
+            )
 
-        return self._credentials
+        return self._service_credentials
 
     def _is_token_valid(self):
         """
         Checks if the stored token is valid.
         """
-        if not self._credentials:
+        if not self._service_credentials:
             return False
-        if not self._credentials.token:
+        if not self._service_credentials.token:
             return False
-        if self._credentials.expired:
+        if self._service_credentials.expired:
             return False
 
         return True
@@ -364,7 +366,7 @@ class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
 
         return credentials
 
-    def _setup_credentials(self, user_email_domain):
+    def _setup_service_credentials(self, user_email_domain):
         """
         Set up the oauth credentials for Google API.
         """
@@ -399,7 +401,7 @@ class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
         Return a set with the google groups a given user/group is a member of, including nested groups if allowed.
         """
 
-        credentials = credentials or self._get_credentials(user_email_domain)
+        credentials = credentials or self._get_service_credentials(user_email_domain)
         checked_groups = checked_groups or set()
         processed_groups = processed_groups or set()
 
--- a/oauthenticator/google.py
+++ b/oauthenticator/google.py
@@ -314,7 +314,7 @@ class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
         # users should be explicitly allowed via config, otherwise they aren't
         return False
 
-    async def _service_client_credentials(self, scopes, user_email_domain):
+    def _service_client_credentials(self, scopes, user_email_domain):
         """
         Return a configured service client credentials for the API.
         """
@@ -338,11 +338,11 @@ class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
 
         return credentials
 
-    async def _setup_credentials(self, user_email_domain):
+    def _setup_credentials(self, user_email_domain):
         """
         Set up the oauth credentials for Google API.
         """
-        credentials = await self._service_client_credentials(
+        credentials = self._service_client_credentials(
             scopes=[f"{self.google_api_url}/auth/admin.directory.group.readonly"],
             user_email_domain=user_email_domain,
         )
@@ -372,7 +372,7 @@ class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
         Return a set with the google groups a given user/group is a member of, including nested groups if allowed.
         """
 
-        credentials = credentials or await self._setup_credentials(user_email_domain)
+        credentials = credentials or self._setup_credentials(user_email_domain)
         checked_groups = checked_groups or set()
         processed_groups = processed_groups or set()
 
--- a/oauthenticator/google.py
+++ b/oauthenticator/google.py
@@ -4,6 +4,7 @@ A JupyterHub authenticator class for use with Google as an identity provider.
 
 import os
 
+import aiohttp
 from jupyterhub.auth import LocalAuthenticator
 from tornado.auth import GoogleOAuth2Mixin
 from tornado.web import HTTPError
@@ -243,7 +244,7 @@ class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
 
         user_groups = set()
         if self.allowed_google_groups or self.admin_google_groups:
-            user_groups = self._fetch_member_groups(user_email, user_domain)
+            user_groups = await self._fetch_member_groups(user_email, user_domain)
         # sets are not JSONable, cast to list for auth_state
         user_info["google_groups"] = list(user_groups)
 
@@ -314,7 +315,7 @@ class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
         # users should be explicitly allowed via config, otherwise they aren't
         return False
 
-    def _service_client_credentials(self, scopes, user_email_domain):
+    async def _service_client_credentials(self, scopes, user_email_domain):
         """
         Return a configured service client credentials for the API.
         """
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -1,3 +1,4 @@
+aiohttp
```
consideRatio, line 1: Remove this from here, and instead add it as an optional dependency for the `googlegroups` extras listed in setup.py.

We'll need to allow import of this to fail as well when doing so, so similarly to how its done here:

```python
        try:
            from google.auth.transport import requests
        except:
            raise ImportError(
                "Could not import google.auth.transport's requests,"
                "you may need to run 'pip install oauthenticator[googlegroups]' or not declare google groups"
            )
```

consideRatio (commented):
We could use `aiohttp`, but there is also the OAuthenticator base class function below that makes use of `tornado.HTTPRequest` which can be done async. If using this could work, that would be preferred as we avoid introducing another way of making HTTP requests.

```python
    async def httpfetch(
        self, url, label="fetching", parse_json=True, raise_error=True, **kwargs
    ):
        """Wrapper for creating and fetching http requests

        Includes http_request_kwargs in request kwargs
        logs error responses, parses successful JSON responses

        Args:
            url (str): url to fetch
            label (str): label describing what is happening,
                used in log message when the request fails.
            parse_json (bool): whether to parse the response as JSON
            raise_error (bool): whether to raise an exception on HTTP errors
            **kwargs: remaining keyword args
                passed to underlying `tornado.HTTPRequest`, overrides
                `http_request_kwargs`
        Returns:
            parsed JSON response if `parse_json=True`, else `tornado.HTTPResponse`
        """
        request_kwargs = self.http_request_kwargs.copy()
        request_kwargs.update(kwargs)
        req = HTTPRequest(url, **request_kwargs)
        return await self.fetch(
            req, label=label, parse_json=parse_json, raise_error=raise_error
        )
```

jrdnbradford:
Thanks for your guidance here on using `httpfetch`, @consideRatio.

consideRatio:
Oh beautiful!!! Have you been able to test that this works as well? I think our test coverage won't be able to tell so well if it does

jrdnbradford:
I've tested this on `the-littlest-jupyterhub` `1.0.0` on the following, including with the new `include_nested_groups` config:
```bash
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 24.04.1 LTS
Release:        24.04
Codename:       noble
```

If you like, I can probably test with several logins at the same time sometime soon. 🚀

oauthenticator/google.py
```diff
@@ -389,20 +370,22 @@ def _fetch_member_groups(
...
         """
-        # FIXME: When this function is used and waiting for web request
-        #        responses, JupyterHub gets blocked from doing other things.
-        #        Ideally the web requests should be made using an async client
-        #        that can be awaited while JupyterHub handles other things.
-        #
-        if not hasattr(self, 'service'):
-            self.service = self._setup_service(user_email_domain, http)
+        if not hasattr(self, 'credentials'):
+            self.credentials = await self._setup_credentials(user_email_domain)
```
consideRatio, line 374: I think this could be trouble, but I'm not sure.

`self` is an authenticator object, I'm not sure how long lived they are and if they are tied to a specific user. I think they are tied to a specific user, but I think they may be somewhat long lived, due to that, it could be that credentials is requested and saved but that they expire after a while, and then things go bad thereafter.

@minrk could you help us here, whats the lifecycle of an authenticator instance - and are they instanciated per user?

jrdnbradford:
> @jrdnbradford I pushed a commit removing a dependency I think wasn't used more, do you think that was correct? I really appreciate those complexity reductions, thanks for working this!!!

I'm pretty sure you're right, now that we don't build the service that isn't needed. I'll test to make sure. I've also scheduled a time tomorrow to get several semi-simultaneous logins on a dev TLJH with the oauthenticator changes. I'll report back after.

oauthenticator/google.py
jrdnbradford: Sounds like this alleviates @consideRatio's worry, but seems like I may need to add some logic to refresh the oauth token depending on how long-lived this object is?

consideRatio: Ah no if the object is long lived, as demonstrated by @manics with count being incremented over time, it means that setting credentials will be done once and between users, and the'll expire etc.

I think the thing to do is to just ensure that you get credentials once per recursive-call-chain to the fetch member groups function, like is done for checked groups parameter - its initialized once, but wont be reused between users etc

consideRatio: If the credentials used to fetch groups has been tied to a user, then we absolutely must re-acquire credentials and not re-use those.

oauthenticator/google.py
```diff
@@ -314,7 +314,7 @@ async def check_allowed(self, username, auth_model):
         # users should be explicitly allowed via config, otherwise they aren't
         return False
 
-    def _service_client_credentials(self, scopes, user_email_domain):
+    async def _service_client_credentials(self, scopes, user_email_domain):
```
consideRatio, line 317: This function doesnt seem to be relevant to have async, because it doesnt await anything.

oauthenticator/google.py
jrdnbradford: I believe my latest commit [7884369](https://github.com/jupyterhub/oauthenticator/pull/764/commits/7884369e7f16330ca1492d725c634f6d8b7b6116) adds what you're getting at here, @consideRatio. No longer an attrib of the object, it is refreshed for each login. 

According to [Google's oauth docs](https://developers.google.com/identity/protocols/oauth2#expiration) there is a limit to the number of refresh tokens before the oldest ones begin to be invalidated, but this limit does **not** apply to service accounts so I believe we've got everything resolved here.

I will test this with a group of users today.

oauthenticator/google.py
```diff
@@ -338,71 +338,54 @@ def _service_client_credentials(self, scopes, user_email_domain):
...
-        )
-        service = self._service_client(
-            service_name='admin',
-            service_version='directory_v1',
-            credentials=credentials,
-            http=http,
-        )
-        return service
+        request = requests.Request()
+        credentials.refresh(request)
```
consideRatio, line 359: I'm not sure what goes on here, but i think this may involve API communication etc, making python blocked. Not great, but we can live with it i figure.

oauthenticator/google.py
```diff
@@ -338,71 +338,54 @@ def _service_client_credentials(self, scopes, user_email_domain):
...
-            self.service = self._setup_service(user_email_domain, http)
 
+        credentials = credentials or await self._setup_credentials(user_email_domain)
         checked_groups = checked_groups or set()
         processed_groups = processed_groups or set()
 
-        resp = self.service.groups().list(userKey=member_email).execute()
+        headers = {'Authorization': f'Bearer {credentials.token}'}
+        url = f'https://www.googleapis.com/admin/directory/v1/groups?userKey={member_email}'
+        group_data = await self.httpfetch(
```
consideRatio, line 381: This is really relevant to await, because we'll do it many times now.

oauthenticator/google.py
consideRatio: This PR is an incremental improvement no matter what.

Im not sure what goes on here though, is it making API requests etc?

oauthenticator/google.py
jrdnbradford: Fixed!

oauthenticator/google.py
jrdnbradford: It is indeed making an API request, and I've been trying to find a way to make it async will no luck. Here's are my initial thoughts re what we should do after reading through more docs and learning more about how the authenticator works:

1. Go back to having the creds be set in the object instead of setting them up and refreshing the token for each login
2. Add methods that check whether the creds exist and the token is valid/non-expired and refresh them using the synchronous call if not. See https://google-auth.readthedocs.io/en/latest/reference/google.auth.credentials.html#google.auth.credentials.Credentials.

This will be the flow:

1. When the first user logs in, creds are setup and a token is retrieved (synchronously)
2. When user 1 + N user logs in, the authenticator checks that the token is still valid. If so, no need to make another _synchronous_ call to get it, use it so fetch groups _asynchronously_. If not, send the _synchronous_ request to refresh the token.

This way we reduce the number of synchronous calls to nearly 0 while also accounting for [potential token expiration](https://developers.google.com/identity/protocols/oauth2#expiration).

jrdnbradford: Latest commit implements the above, @consideRatio.

Also, the test with multiple repeated logins several users went well, seems like the general async fetching of groups works well.

oauthenticator/google.py
```diff
@@ -14,6 +14,7 @@
 
 class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
     user_auth_state_key = "google_user"
+    _credentials = None
```
manics, line 17: Can you rename this, and the associated methods, to something like `service_credentials` to make it clear these aren't related to user credentials?

oauthenticator/google.py
```diff
@@ -314,6 +315,31 @@ async def check_allowed(self, username, auth_model):
...
 
+    def _get_credentials(self, user_email_domain):
+        """
+        Returns the stored credentials or fetches and stores new ones.
+
+        Checks if the credentials are valid before returning them. Refreshes
+        if necessary and stores the refreshed credentials.
+        """
+        if not self._credentials or not self._is_token_valid():
+            self._credentials = self._setup_credentials(user_email_domain)
```
consideRatio, line 326: Credentials seems to be received for a `user_email_domain`, doesn't that makes them different between users? If so, isn't it causing issues to re-use them for any user?

I figure we would at least need to have a dictionary of credentials, with one for each user_email_domain, right?

oauthenticator/google.py
consideRatio: Its very elegant to do it like this, nice! I wonder if a tweak may be needed though, as discussed in https://github.com/jupyterhub/oauthenticator/pull/764/commits/389d4b2a723fe48ddce215f8c20210124d0fc363#r1788133777.

I don't get what credentials we have, but since we reference `user_email_domain`, it makes me think there may be an issue when we have mutliple users with different domain in their email. I really don't know though - but from a shallow code inspection and little knowledge on the involved kinds of credentials I suspect it could be an issue.

oauthenticator/google.py
manics: If multiple people login at the same time `_setup_credentials` could be called multiple times in parallel. Is this likely to cause any problems?

jrdnbradford: Ha, this totally flew past me. All our hubs are single domain, so it didn't occur to me. Will fix!

oauthenticator/google.py
jrdnbradford: This will make things much clearer, thanks @manics.

oauthenticator/google.py
jrdnbradford: Ha, @consideRatio this totally flew past me. All our hubs are single domain, so it didn't occur to me. Latest commit stores creds in a dictionary of `{domain: credentials}`. These creds, or in this case the token, can be used multiple times for authenticating users of the particular domain.

jrdnbradford:
I have added support so that creds/tokens are saved on a domain-basis in a dictionary and are refreshed only if necessary. My testing with a single domain works, but I'll need some more time to test multiple Google workspace domains in a JupyterHub installation to ensure it works.

consideRatio (approved):
This is brilliant, really pleased about the refresh logic detail and the removal of now unused python dependency.

If this passes your end to end testing I figure we go for a merge!

@manics what do you think?

oauthenticator/google.py
```diff
@@ -338,71 +369,56 @@ def _service_client_credentials(self, scopes, user_email_domain):
...
         """
         Return a set with the google groups a given user/group is a member of, including nested groups if allowed.
         """
-        # FIXME: When this function is used and waiting for web request
-        #        responses, JupyterHub gets blocked from doing other things.
-        #        Ideally the web requests should be made using an async client
-        #        that can be awaited while JupyterHub handles other things.
-        #
-        if not hasattr(self, 'service'):
-            self.service = self._setup_service(user_email_domain, http)
```
manics, line 406:
```suggestion
# WARNING: There's a race condition here if multiple users login at the same time.
# This is currently ignored.
```

manics (commented):
I've suggested adding a comment to help with future maintenance, e.g. if a change somewhere results in the race condition causing a failure. Otherwise I agree this is good to merge.

jrdnbradford:
Awesome, will report back after I setup another Workspace domain for testing this,

oauthenticator/google.py
```diff
@@ -338,71 +369,57 @@ def _service_client_credentials(self, scopes, user_email_domain):
...
-
+        # WARNING: There's a race condition here if multiple users login at the same time.
+        # This is currently ignored.
+        credentials = credentials or self._get_service_credentials(user_email_domain)
+        token = credentials[user_email_domain].token
```
consideRatio, line 409: @manics I've tried to understand this race condition, but I've failed to see it. Are you sure there is one?

My understanding made explicit: whenever an `await` is showing up, then and only then is Python possibly switching to execute something else.

consideRatio:
> Turns out I don't have the access to another workspace domain to do this testing. @consideRatio and @manics, any contacts that use Google with multiple domains that could install the dev version for a test?

I'm not able to get this done, but I'm willing to go for a merge anyhow I think.

As @manics wrote:

> Otherwise I agree this is good to merge.

I'll go for it!

Thank you a lot for thorough work on this @jrdnbradford !

jrdnbradford:
If there are any issues please @ me and I'd be happy to fix!

Issue #468, opened 2021-11-05
[Generic] Generic Oauthenticator Not support PKCE


iamjxb:
For OIDC login using code flow with PKCE, the generic oauthenticator currently cannot generate code challenge and send code challenge method or code verifier. 
It is possible to add them through the extra_params and extra_authorize_params but that would make it "hard coded" and less secure. 
Is there a way to generate code challenge and verifier during runtime? If not, can we make a feature request for this?
Thank you.

minrk:
Supporting PKCE sounds like a totally valid feature request, and seems pretty generic. If I [understand it correctly][pkce], what we would need is a boolean option for `use_pkce`, where if True, we add:

[pkce]: https://www.oauth.com/oauth2-servers/pkce/

in `OAuthLoginHandler`:

1. we generate a random string (`code_verifier`)
2. hash it with sha256
3. add params to `authorize_redirect`: `code_challenge=$hashed_verifier&code_challenge_method=S256`

in `OAuthCallbackHandler`:

1. retrieve `code_verifier` (persistence TBD; its own cookie, I guess?)
2. send original `code_verifier` in `code_verifier` param for the token request

minrk:
Does it make a lot of sense to use PKCE with authorization_code flow? Reading about it, it seems to mostly be a replacement for mobile apps and single-page apps where the client secret cannot be reasonably protected. But since the code exchange request comes from the server side, this doesn't seem to be an issue in our case.

humanSupafly:
> For OIDC login using code flow with PKCE, the generic oauthenticator currently cannot generate code challenge and send code challenge method or code verifier. It is possible to add them through the extra_params and extra_authorize_params but that would make it "hard coded" and less secure. Is there a way to generate code challenge and verifier during runtime? If not, can we make a feature request for this? Thank you.

Does generic oauthenticator support pkce?

How did you make it work using the extra_params and "hardcode" approch?

PR #765, opened 2024-09-28
Add `enable_pkce` config, True by default

renan-r-santos:
Fixes https://github.com/jupyterhub/oauthenticator/issues/468

### DIFF
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -684,10 +684,10 @@ class OAuthenticator(Authenticator):
         False,
         config=True,
         help="""
-        Require Proof Key for Code Exchange (PKCE) for OAuth2 authorization code flow
+        Require Proof Key for Code Exchange (PKCE) for OAuth2 authorization code flow.
         
         Only the S256 code challenge method is supported.
-        `RFC 7636 <https://datatracker.ietf.org/doc/html/rfc7636>`.
+        `RFC 7636 <https://datatracker.ietf.org/doc/html/rfc7636>`_.
         """,
     )
 
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -119,7 +119,7 @@ class OAuthLoginHandler(OAuth2Mixin, BaseHandler):
 
         state = {"state_id": state_id, "next_url": next_url}
 
-        if self.authenticator.pkce:
+        if self.authenticator.require_pkce:
             # https://datatracker.ietf.org/doc/html/rfc7636#section-4
             code_verifier = secrets.token_urlsafe(43)
             code_challenge = hashlib.sha256(code_verifier.encode("utf-8")).digest()
@@ -680,13 +680,13 @@ class OAuthenticator(Authenticator):
         """,
     )
 
-    pkce = Bool(
+    require_pkce = Bool(
         False,
         config=True,
         help="""
-        Whether to use PKCE (Proof Key for Code Exchange) for the OAuth2 flow.
-
-        Only the S256 method is supported.
+        Require Proof Key for Code Exchange (PKCE) for OAuth2 authorization code flow
+        
+        Only the S256 code challenge method is supported.
         `RFC 7636 <https://datatracker.ietf.org/doc/html/rfc7636>`.
         """,
     )
@@ -1008,7 +1008,7 @@ class OAuthenticator(Authenticator):
             "data": data,
         }
 
-        if self.pkce:
+        if self.require_pkce:
             # https://datatracker.ietf.org/doc/html/rfc7636#section-4.5
             cookie_state = handler.get_state_cookie()
             if not cookie_state:
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -12,7 +12,6 @@ import secrets
 import uuid
 from functools import reduce
 from inspect import isawaitable
-from typing import cast
 from urllib.parse import quote, urlencode, urlparse, urlunparse
 
 import jwt
@@ -682,13 +681,11 @@ class OAuthenticator(Authenticator):
     )
 
     pkce = Bool(
-        os.environ.get("OAUTH2_PKCE", "False").lower() in {"true", "1"},
+        False,
         config=True,
         help="""
         Whether to use PKCE (Proof Key for Code Exchange) for the OAuth2 flow.
 
-        When using PKCE, the client secret is not required to be sent to the
-        token endpoint.
         Only the S256 method is supported.
         `RFC 7636 <https://datatracker.ietf.org/doc/html/rfc7636>`.
         """,
@@ -1013,9 +1010,7 @@ class OAuthenticator(Authenticator):
 
         if self.pkce:
             # https://datatracker.ietf.org/doc/html/rfc7636#section-4.5
-            callback_handler = cast(OAuthCallbackHandler, handler)
-
-            cookie_state = callback_handler.get_state_cookie()
+            cookie_state = handler.get_state_cookie()
             if not cookie_state:
                 raise web.HTTPError(400, "OAuth state missing from cookies")
 
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -93,6 +93,18 @@ class OAuthLoginHandler(OAuth2Mixin, BaseHandler):
             STATE_COOKIE_NAME, state_cookie_value, expires_days=1, httponly=True
         )
 
+    def _generate_pkce_params(self):
+        # https://datatracker.ietf.org/doc/html/rfc7636#section-4
+        # It is recommended that the output of the random number generator creates
+        # a 32-octet sequence which is base64url-encoded to produce a 43-octet URL
+        # safe string to use as the code verifier.
+        code_verifier = secrets.token_urlsafe(32)
+        code_challenge = hashlib.sha256(code_verifier.encode("utf-8")).digest()
+        code_challenge_base64 = (
+            base64.urlsafe_b64encode(code_challenge).decode("utf-8").rstrip("=")
+        )
+        return code_verifier, code_challenge_base64
+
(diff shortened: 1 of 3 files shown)


oauthenticator/oauth2.py
```diff
@@ -980,6 +1011,20 @@ def build_access_tokens_request_params(self, handler, data=None):
             "data": data,
         }
 
+        if self.pkce:
+            # https://datatracker.ietf.org/doc/html/rfc7636#section-4.5
+            callback_handler = cast(OAuthCallbackHandler, handler)
```
manics, line 1016: Personally I have no objection to introducing typing, but I think we should discuss it in a separate issue first.

oauthenticator/oauth2.py
```diff
@@ -663,6 +681,19 @@ def _allowed_scopes_validation(self, proposal):
...
     )
 
+    pkce = Bool(
+        os.environ.get("OAUTH2_PKCE", "False").lower() in {"true", "1"},
+        config=True,
+        help="""
+        Whether to use PKCE (Proof Key for Code Exchange) for the OAuth2 flow.
+
+        When using PKCE, the client secret is not required to be sent to the
+        token endpoint.
```
manics, line 691: https://www.oauth.com/oauth2-servers/pkce/authorization-request/ says

> The authorization server can require that public clients must use the PKCE extension. This is really the only way to allow native apps to have a secure authorization flow without using the client secret, especially without the redirect URI security that’s available with web-based clients.

Does this imply it should still be sent unless the server requires PKCE?

oauthenticator/oauth2.py
```diff
@@ -663,6 +681,19 @@ def _allowed_scopes_validation(self, proposal):
         """,
     )
 
+    pkce = Bool(
+        os.environ.get("OAUTH2_PKCE", "False").lower() in {"true", "1"},
```
manics, line 685: I don't think we need this environment variable, it's present in some of the other config options for backwards compatibility.

oauthenticator/tests/test_oauth2.py
```diff
@@ -206,3 +207,57 @@ async def test_add_user_override(
...
+    )
+    assert (
+        login_handler.authorize_redirect.call_args.kwargs["extra_params"][
+            "code_challenge_method"
+        ]
+        == "S256"
+    )
+
+
+async def test_callback_handler_pkce():
```
manics, line 235: Can you also add a test to verify an error is returned if PKCE is requested but the server doesn't support it?

renan-r-santos: Done. I guess the way an Oauth server tells it doesn't support PKCE would be by returning a 403 when the client tries to exchange the code for a token.

oauthenticator/oauth2.py
renan-r-santos: Removed.

oauthenticator/oauth2.py
renan-r-santos: You're right. In fact, the code itself was contradicting this comment. I removed the part about `client_secret`.

oauthenticator/oauth2.py
renan-r-santos: Oh, sorry about that. I agree that this subject should be a separate discussion.

renan-r-santos:
By the way, I think this PR also helps a bit in getting OAuthenticator compatible with [OAuth 2.1](https://oauth.net/2.1/) (still in draft) as:
| PKCE is required for all OAuth clients using the authorization code flow

oauthenticator/tests/test_oauth2.py
minrk: Won't it work but just not be checked for validity if the provider doesn't support it (ignored extra parameters)? If that's true, should it be on by default?

manics: If the parameters are definitely ignored then we could always send them, and change the property name to `require_pkce` to enforce it on the client side.

manics: @renan-r-santos Sorry, I realise now I was unclear in my request. I was thinking of testing that `raise web.HTTPError(400, "Missing code_verifier")` is raised when the server silently ignores the PKCE request.

Regarding whether or not to always send the PKCE request, how about if we rename the parameter `require_pkce` instead of `pkce`, but keep the current implementation (only send the PKCE field when `require_pkce = True`? That lets us switch to always sending PKCE in future if we want, without having to change or add any parameters.

renan-r-santos: > Regarding whether or not to always send the PKCE request, how about if we rename the parameter require_pkce instead of pkce, but keep the current implementation (only send the PKCE field when require_pkce = True?

That sounds like a good plan to me. I've updated the PR to reflect that.

> I was thinking of testing that raise web.HTTPError(400, "Missing code_verifier") is raised when the server silently ignores the PKCE request.

`raise web.HTTPError(400, "Missing code_verifier")` won't get raised if the server silently ignores the PKCE request. `code_verifier` is data that we store in a cookie together with `state_id` and `next_url` in the login handler, but `code_verifier` isn't sent to or returned from the OAuth provider during login. It is only during code exchange that the client grabs the `code_verifier` previously stored in a cookie and sends it to the server so it can hash it and compare it with the `code_challenge`.

So, the error you mentioned can only happen if the cookie got somehow deleted or corrupted between login and callback handlers. If you still think it is worth adding a test for that, let me know and I'll update the PR. I could be wrong, but I don't think there's a way for a client to know if a server ignores PKCE parameters. On the other hand, a server can enforce that clients use PKCE.

minrk: Yeah, I think PKCE is only for the provider to check; clients only provide information. So the only reason to disable it that I can see is if some provider raises on unrecognized arguments, which is [officially wrong](https://www.rfc-editor.org/rfc/rfc6749#section-3.1):

> The authorization server MUST ignore unrecognized request parameters.

So if we're only talking about valid OAuth providers, we don't even need to make it optional.

From the [PKCE spec](https://datatracker.ietf.org/doc/html/rfc7636#section-5):

> As the OAuth 2.0 [[RFC6749](https://datatracker.ietf.org/doc/html/rfc6749)] server responses are unchanged by this
   specification, client implementations of this specification do not
   need to know if the server has implemented this specification or not
   and SHOULD send the additional parameters as defined in [Section 4](https://datatracker.ietf.org/doc/html/rfc7636#section-4) to
   all servers.

i.e. it's always right to send PKCE, and it's entirely up to the provider to decide whether to validate or not.

oauthenticator/oauth2.py
```diff
@@ -115,7 +117,22 @@ def get(self):
         state_id = self._generate_state_id()
         next_url = self._get_next_url()
 
-        cookie_state = _serialize_state({"state_id": state_id, "next_url": next_url})
+        state = {"state_id": state_id, "next_url": next_url}
+
+        if self.authenticator.require_pkce:
+            # https://datatracker.ietf.org/doc/html/rfc7636#section-4
+            code_verifier = secrets.token_urlsafe(43)
```
minrk, line 124: FWIW, the spec [recommends](https://datatracker.ietf.org/doc/html/rfc7636#section-4.1) a 32-byte token. The input to `secrets.token_*` is the number of random bytes not the encoded length, so with

```suggestion
            code_verifier = secrets.token_urlsafe(32)
```

`len(code_verifier)` will be 43.

oauthenticator/oauth2.py
```diff
@@ -663,6 +680,17 @@ def _allowed_scopes_validation(self, proposal):
         """,
     )
 
+    require_pkce = Bool(
+        False,
+        config=True,
+        help="""
+        Require Proof Key for Code Exchange (PKCE) for OAuth2 authorization code flow.
```
minrk, line 687: Sorry for giving you the runaround a bit on names, but let's use:

```suggestion
    enable_pkce = Bool(
        True,
        config=True,
        help="""
        Enable Proof Key for Code Exchange (PKCE) for OAuth2 authorization code flow.
```

I think `require_pkce` gives the wrong impression about what oauthenticator is in a position to enforce.

oauthenticator/oauth2.py
```diff
@@ -680,13 +680,13 @@ def _allowed_scopes_validation(self, proposal):
...
 
-    pkce = Bool(
+    require_pkce = Bool(
         False,
         config=True,
         help="""
-        Whether to use PKCE (Proof Key for Code Exchange) for the OAuth2 flow.
-
-        Only the S256 method is supported.
+        Require Proof Key for Code Exchange (PKCE) for OAuth2 authorization code flow
```
consideRatio: Is this to say that the oauth server requires pkce, so that the client is configured to say the server requires it? Or, is it that the client is configured to require it independent of the server, or for servers that supports it optionally and/or requires it?

I'm not clear about the situation, but i hope that these confusions can be clarified in the docstring.

consideRatio: Thank you @renan-r-santos for working this!!!

manics: I think this is what confused me at first- based on the discussions in this PR only the server can require PKCE, but a client can't.

oauthenticator/tests/test_oauth2.py
renan-r-santos: Thank you very much for the references, especially

> The authorization server MUST ignore unrecognized request parameters.

which I didn't know and makes a lot of difference.

renan-r-santos:
Guys, I think this is ready for another round of reviews. I made a few improvements and incorporated important points that we discussed here in the `enable_pkce` docs. Thank you for your patience!

consideRatio (approved):
I read the docstring and considered the config naming and think its great, i'll leave the implementation review to another person as i think its already been considered and i've not onboarded myself enough to review it myself yet.

Thank you @renan-r-santos for working this so thoroughly!! ♥️🎉🥳

PR #768, opened 2024-10-08
[Google] Handle pagination in group API calls

### LINKED ISSUE #767: Lookup of Google Groups needs to account for pagination
### Proposed change
Found while working on #763 and #764. This issue was present in the previous code as well as the updated call in #764.

Currently the API call to fetch Google Groups doesn't handle pagination. See the [API docs](https://developers.google.com/admin-sdk/directory/reference/rest/v1/groups/list) for reference.

This has most likely not been reported yet because the default max number of groups retrieved is high enough that it hasn't caused a problem for users of Google Groups, although I cannot find what this number is (it is missing in the [REST API docs](https://developers.google.com/admin-sdk/directory/reference/rest/v1/groups/list) and the [Python docs](https://developers.google.com/resources/api-libraries/documentation/admin/directory_v1/python/latest/admin_directory_v1.groups.html)).

### Alternative options
We could ignore this problem since it apparently hasn't caused any issues, or we could add `maxResults=200` to expand the number of potential groups retrieved. But if someone attempts to use the Google Oauthenticator and their Workspace users are members of hundreds of groups, there is potential for failed logins for both these alternative options.

Those who use Google Groups for authentication.

Add logic to handle pagination and retrieve all groups.

jrdnbradford:
Closes #767. 

The `http` param was leftover from #764 and has been removed.

The main change is checking for a paging token and continuing to retrieve groups as long as it exists.

### DIFF
--- a/oauthenticator/google.py
+++ b/oauthenticator/google.py
@@ -403,7 +403,6 @@ class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
         self,
         member_email,
         user_email_domain,
-        http=None,
         checked_groups=None,
         processed_groups=None,
         credentials=None,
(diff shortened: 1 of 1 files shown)


minrk:
Great, thank you!

PR #350, opened 2020-05-01
Adding refresh_token to GoogleOAuthenticator

missingcharacter:
I've been working on giving my users a unified experience using G Suite, we ran into a situation where we wanted to have access to files in Google Drive and most applications that interact with Google Drive (Example: https://github.com/astrada/google-drive-ocamlfuse) require a `refresh_token` in order to work.

Since users already authenticate to Google when they login to jupyterhub I thought it was best to do all at once. 

These are the changes and I made and I thought the rest of the community would benefit from them too.

PR #772, opened 2024-10-17
Simplify retrieval of encrypted auth state

minrk:
there's a public method for this, no need to call decryption methods directly

### DIFF
--- a/oauthenticator/oauth2.py
+++ b/oauthenticator/oauth2.py
@@ -16,7 +16,6 @@ from urllib.parse import quote, urlencode, urlparse, urlunparse
 
 import jwt
 from jupyterhub.auth import Authenticator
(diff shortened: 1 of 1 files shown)


minrk:
Actually, on re-reading the only discussion of this that I can find [here](https://github.com/jupyterhub/oauthenticator/pull/350#discussion_r420138763), I _think_ this only came up during the development of the refresh_token auth when the previous launch had the right scopes to request a refresh token, while the latest config does not. I think the right behavior in this situation is to not have a refresh token, since that reflects current config.

minrk:
I'm _slightly_ less than 100% sure, though.

manics (approved):

manics:
Oops, I loaded this page before your most recent comments so didn't see them before merging.

PR #778, opened 2024-12-02
Add how-to doc on refresh tokens

minrk:
docs for #579 

it would be nice to do a 17.2 with this, I think

### DIFF
--- a/docs/source/how-to/refresh.md
+++ b/docs/source/how-to/refresh.md
@@ -44,7 +44,7 @@ because it ensures the token is valid when the server starts.
 This example requires granting users read access to their own `auth_state`.
 If you plan to provide users with access tokens,
 `auth_state` does not typically include information your users won't have access to with the token itself,
-but it is worth making sure that your Authenticator configuration places anything in `auth_state`
+but it is worth making sure that your Authenticator configuration doesn't put anything in `auth_state`
 that you do not want users to be able to see.
 ```
 
--- a/docs/source/how-to/refresh.md
+++ b/docs/source/how-to/refresh.md
@@ -1,6 +1,6 @@
 # Refreshing user authentication
 
-JupyterHub has a mechanism called [`refresh_user`](inv:jupyterhub:py:method#jupyterhub.auth.Authenticator.refresh_user) that is meant to 'refresh' information from the Authentication provider periodically.
+JupyterHub has a mechanism called [`refresh_user`](inv:jupyterhub:py:method#jupyterhub.auth.Authenticator.refresh_user) that is meant to _refresh_ information from the Authentication provider periodically.
 This allows you to make sure things like group membership or other authorization info is up-to-date.
 In OAuth, this can also mean making sure the access token has not expired.
 This is particularly useful in deployments where an access token from the oauth provider is passed to the Server environment,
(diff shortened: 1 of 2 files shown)


docs/source/how-to/refresh.md
```diff
@@ -0,0 +1,124 @@
...
+At this point:
+
+1. When a user logs in, the OAuth user info and access token are encrypted and persisted in the Hub database.
+2. When the server token requests the user model at `/hub/api/user`, an `auth_state` field will be present, containing the current auth state.
+3. Further, when accessing `/hub/api/user` the `refresh_user` logic is triggered if `auth_refresh_age` has elapsed since the last refresh.
+
+This means that you can access `/hub/api/user` with `$JUPYTERHUB_API_TOKEN` and it will **always return a valid access token**,
+even if the currently stored token has expired when the request is made.
+
+To retrieve the access token, make a request to `${JUPYTERHUB_API_URL}/hub/user` with `${JUPYTERHUB_API_TOKEN}, e.g. from Python:
```
manics, line 90:
```suggestion
To retrieve the access token, make a request to `${JUPYTERHUB_API_URL}/hub/user` with `${JUPYTERHUB_API_TOKEN}`, e.g. from Python:
```

docs/source/how-to/refresh.md
```diff
@@ -0,0 +1,124 @@
+# Refreshing user authentication
+
+JupyterHub has a mechanism called [`refresh_user`](inv:jupyterhub:py:method#jupyterhub.auth.Authenticator.refresh_user) that is meant to 'refresh' information from the Authentication provider periodically.
```
manics, line 3:
```suggestion
JupyterHub has a mechanism called [`refresh_user`](inv:jupyterhub:py:method#jupyterhub.auth.Authenticator.refresh_user) that is meant to _refresh_ information from the Authentication provider periodically.
```
I think italics should look better when rendered:
image

docs/source/how-to/refresh.md
```diff
@@ -0,0 +1,124 @@
...
+## Refreshing tokens from user sessions
+
+If your user sessions use access tokens from your oauth provider and those tokens may expire during user sessions,
+you can rely on this mechanism to get fresh access tokens from JupyterHub.
+
+The first step is to grant the _server_ token access to read auth state for its owner.
+Users do not have permission to read their own auth state by default,
+but `auth_state` is where the `access_token` is stored.
+We need to grant the `admin:auth_state!user` scope to both the `user` and `server` roles,
+so that requests with `$JUPYTERHUB_API_TOKEN` will have permission to read the access token:
```
manics, line 50: Is there likely to be anything in `auth_state` that an admin wouldn't want a user to see?

minrk: It's worth mentioning this, because it is possible in theory and there are not currently more fine-grained controls. I don't believe so with the current OAuthenticators, at least, which have the userinfo and token responses. Essentially, all the information OAuthenticator puts in auth state is retrievable with the access token we are after, so if you are doing this to pass the access token, access to auth state isn't giving access to any more info than you are already meaning to give by relaying the access token itself.

minrk: Added a caveat to the top of this section

docs/source/how-to/refresh.md
```diff
@@ -0,0 +1,132 @@
...
+This is most useful when the server is being passed an access token
+because it ensures the token is valid when the server starts.
+
+## Refreshing tokens from user sessions
+
+```{warning}
+This example requires granting users read access to their own `auth_state`.
+If you plan to provide users with access tokens,
+`auth_state` does not typically include information your users won't have access to with the token itself,
+but it is worth making sure that your Authenticator configuration places anything in `auth_state`
```
manics, line 47:
```suggestion
but it is worth making sure that your Authenticator configuration doesn't put anything in `auth_state`
```
Go ahead and merge after this fix (or equivalent)

minrk: Thanks!

minrk:
Thanks!

PR #780, opened 2024-12-03
add refresh_user_hook

minrk:
allows customizing/overriding refresh_user behavior when the default isn't desirable.

testing `refresh_user` on 2i2c revealed a use case I had not considered, which is reliance on 'fake' users that don't actually exist in the oauth provider, and `refresh_user` means those users cannot actually do things, because their auth state is never valid. I added what should be the solution for 2i2c's health check accounts as an example in the docs. This makes enabling refresh_user by default a more disruptive change than I realized, if we want to switch the default to a more explicit opt-in. I still think it is the right thing to do in most cases, so in that way makes sense as a default (eventually, if not immediately).

I also hadn't realized that `refresh_pre_spawn` is overridden to True by default in OAuthenticator, which was surprising since it's been that way for _years_, despite `refresh_user` never having been implemented before now.

Maybe I should have done a beta of 17.2. We can yank 17.2, if folks feel that would take some pressure off, and then take a more careful beta process pushing this out again as 17.3.

### DIFF
--- a/docs/source/how-to/refresh.md
+++ b/docs/source/how-to/refresh.md
@@ -132,3 +132,38 @@ c.Authenticator.auth_refresh_age = 0
 in which case the new `refresh_user` method will not be called.
 This is equivalent to the behavior of OAuthenticator 17.1 and earlier,
 where the default `refresh_user` was called, but did nothing.
+
+## Customizing refresh behavior
+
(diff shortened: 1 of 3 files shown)


GeorgianaElena (approved):
Thank you for fixing this so quickly @minrk, this looks good to be merged! Also, I'm +1 on getting this out again as 17.3

PR #781, opened 2024-12-11
release 17.3.0

minrk:
only has #780 which proved necessary for refresh_user to not cause problems in our jupyter-health 2i2c deployment where we were testing this. With this, it's been working great.

As I mentioned there, I'm okay with yanking 17.2 and doing a beta here to give refresh_user more testing time, or just make this release and keep moving. Both work for me. I suspect the problem caused by refresh_user for 2i2c is rare (automatic creation of users not recognized by the Authenticator), but it may not be the only one.

### DIFF
--- a/docs/source/reference/changelog.md
+++ b/docs/source/reference/changelog.md
@@ -6,6 +6,27 @@ command line for details.
 
 ## [Unreleased]
 
+## 17.3
+
+### [17.3.0] - 2024-12-11
+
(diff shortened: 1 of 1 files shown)


consideRatio (approved):

consideRatio:
Going foe 17.3.0 seems fine to me!

minrk:
Thanks! Published 17.3.0

Issue #790, opened 2025-03-31
Buffer for renewing both id_tokens and access_tokens when they are about to expire, but not actually expired.


jabbera:
### Proposed change

If the id_token or access_token are going to expire within the next N minutes, I'd like to be able to have the refresh process renew them. 

We use the tokens as part of the oauth OBO flow in bot hooks and activly on servers. Knowing the access token isn't expired exactly when the refresh process runs isn't helpful. First the id_token expiration isn't checked at all it seems, second the access/id_token could not be expired a T0 when refresh_runs but at T+1 when we need it in the hook it could be expired.

### Alternative options

We have implemented the following hook: 

It's ugly because it depends on the internal implementation of an HTTPClientError with a 400 error code is what causes the refresh process:

````
    async def modify_auth_state_hook(self, _, auth_state: dict[str, Any]) -> dict[str, Any]:
        if not (self._token_almost_expired(auth_state.get("access_token", None)) or self._token_almost_expired(auth_state.get("id_token", None))):
            self.log.error(f"Tokens are not expired for: {auth_state['user']['name']}")
            return auth_state

        raise HTTPClientError(
            401,
            "Tokens are expired. Please re-authenticate.",
        )


    def _token_almost_expired(self, str_token: str | None):
        """Returns True if the token is expired."""
        if not str_token:
            return True

        decoded_token = jwt.decode(str_token, options={"verify_signature": False})
        token_expiration = datetime.fromtimestamp(decoded_token["exp"], tz=timezone.utc)
        deadline = datetime.now(timezone.utc) + self._token_expiration_buffer
        return token_expiration < deadline
```

### Who would use this feature?
Anyone who needs to use the access_token or id_token during hooks or server lifetime


### (Optional): Suggest a solution
<!-- Describe what you think needs to be done. Doing that is an excellent first step to get the feature implemented. -->

minrk:
Absolutely! I tried to figure this out in a generic way, but couldn't since OAuth doesn't standardize this, though OIDC does.

It might be that the [refresh_user_hook](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.oauth2.html#oauthenticator.oauth2.OAuthenticator.refresh_user_hook) is what you want, which is where we check if access tokens are still valid.

We could provide a default OIDC implementation of this with a max age config so you can import one instead of reimplementing it.

jabbera:
[...]

PR #795, opened 2025-06-06
Fix handling of expired id tokens

consideRatio:
This fix is only be relevant when `userdata_from_id_token` is True.


jabbera:
If you are looking at this I would encourage you to combine this: https://github.com/jupyterhub/oauthenticator/issues/790 into the feature. Instead of looking at if the token is expired, look at if it is or is about to expire within some buffer. id tokens can be used outside the context of the refresh process for lots of useful things. Knowing they are not expired exactly when the refresh is called is not as helpful IMO.

PR #818, opened 2026-05-22
update release workflow for tighter permissions

minrk:
only download-artifact and publish to pypi have permission to upload, not the whole build process

simplified version of jupyterhub's current release workflow (without the added tests and nodejs setup)

needs additional action:

- [x] create `release` environment restricted to all tags


consideRatio (approved):

minrk:
@consideRatio yes, looks right! I was investigating if you could do it with the API (so I could script it across all repos), but there's no API that I can find for creating environment branch/tag rules

PR #823, opened 2026-06-19
[CILogon] support callable username_claim

### LINKED ISSUE #712: Can not use CILogonOAuthenticator with ORCID

minrk:
restores compatibility with base class by allowing a callable, both on the class level OAuthenticator.username_claim, and on the per-idp level.

idp-level username_claim has priority, so class-level username_claim is only used when idp.username_derivation.username_claim is _not_ defined.

closes #712 with by allowing callable `username_claim` but if we wanted a simple _declarative_ solution to orcid (e.g. strip url prefix analogous to strip idp domain), we could also do that separately if it's likely to be common enough.

Example orcid config:

```python
def orcid_username(user_info):
    assert user_info["idp"] == 'http://orcid.org/oauth/authorize'
    # Only modify usernames if orcid is used
    # oidc is of the form https://orcid.org/<orcid-id>
    return user_info['oidc'].split('/')[-1]

c.CILogonOAuthenticator.username_claim = orcid_username
c.CILogonOAuthenticator.idps = {
    'http://orcid.org/oauth/authorize': {
        'username_derivation': {
            # could set username_claim: callable here, too
            'action': 'prefix',
            'prefix': 'orcid',
    },
    'other': {
        'username_derivation': {
            'username_claim': 'sub', # username_claim callable not used
    },
}
```

### DIFF
--- a/oauthenticator/cilogon.py
+++ b/oauthenticator/cilogon.py
@@ -216,6 +216,13 @@ class CILogonOAuthenticator(OAuthenticator):
         .. versionchanged:: 15.0
 
            Changed format from a list to a dictionary.
+
+        .. versionchanged:: 17.5
+
+           `username_claim` can be a callable,
+           and top-level `CILogonOAuthenticator.username_claim` can be a callable,
+           restoring feature parity with base OAuthenticator.
+           The top-level `username_claim` callable will only be used when `username_derivation.username_claim` is not defined for a given identity provider.
         """,
     )
 
@@ -272,7 +279,6 @@ class CILogonOAuthenticator(OAuthenticator):
         "strip_idp_domain": ("idps", "15.0.0", False),
         "shown_idps": ("idps", "16.0.0", False),
         "additional_username_claims": ("idps", "16.0.0", False),
-        "username_claim": ("idps", "16.0.0", False),
         "allowed_idps": ("idps", "17.4.0", True),
         **OAuthenticator._deprecated_oauth_aliases,
     }
@@ -309,14 +315,6 @@ class CILogonOAuthenticator(OAuthenticator):
         """,
     )
     additional_username_claims = List(
-        config=True,
-        help="""
-        .. versionremoved:: 16.0
-
-           Use :attr:`idps`.
-        """,
-    )
-    username_claim = Unicode(
         config=True,
         help="""
         .. versionremoved:: 16.0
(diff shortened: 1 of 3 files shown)


oauthenticator/cilogon.py
```diff
@@ -369,12 +367,26 @@ def _user_info_to_unprocessed_username(self, user_info):
...
+                raise web.HTTPError(500, f"Configuration error in idp: {user_idp}")
+
+        if callable(username_claim):
+            username = username_claim(user_info)
+        else:
+            username = user_info.get(username_claim, None)
         if not username:
             message = f"Configured username_claim {username_claim} for {user_idp} was not found in the response {user_info.keys()}"
-            self.log.error(message)
             raise web.HTTPError(500, message)
```
consideRatio, line 390: Should we do `self.log.error` alongside raising a web.HTTPError or not?

consideRatio (approved):
a logging question, but otherwise lgtm!

oauthenticator/cilogon.py
minrk: Every httperror we raise gets logged on the request (`500 GET /path  message...`), so we only need these when the message we want to log differs from what we want to report to the user, such as more detail about context, etc. 

This code would log the same message twice.

consideRatio: Ah okay!

PR #828, opened 2026-07-13
Added allowed_hosted_domains to GoogleOAuthenticator

sean-morris:
Adds `allowed_hosted_domains` config to allow a list of Google-hosted domains to authenticate to the hub while also honoring `allowed_users`, `admin_users` and `allowed_google_groups`. Note: The domains in `allowed_users` and `admin_users` are not required to be in the `allowed_hosted_domains` list, but of course, must be valid Google accounts.

### DIFF
--- a/oauthenticator/google.py
+++ b/oauthenticator/google.py
@@ -161,7 +161,7 @@ class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
            stripped to exclude the `@domain` part.
 
         Users not restricted by this configuration must still be explicitly
-        allowed by a configuration intended to allow users, like `allow_all`
+        allowed by a configuration intended to allow users, like `allow_all`,
         `allowed_users`, `allowed_hosted_domains`, or `allowed_google_groups`.
 
         .. warning::
--- a/oauthenticator/google.py
+++ b/oauthenticator/google.py
@@ -245,7 +245,7 @@ class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
                 "as allow_all already grants access to everyone."
             )
         return proposal.value
-  
+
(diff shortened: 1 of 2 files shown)


oauthenticator/google.py
```diff
@@ -195,10 +198,54 @@ def _cast_hosted_domain(self, proposal):
...
+    @validate('allowed_hosted_domains')
+    def _check_allowed_hosted_domains(self, proposal):
+        if proposal.value and self.allow_all:
+            raise TraitError(
+                "GoogleOAuthenticator.allowed_hosted_domains can't be used together with "
+                "GoogleOAuthenticator.allow_all, as allow_all already grants access to everyone."
+            )
         return [hd.lower() for hd in proposal.value]
 
+    @validate('allow_all')
```
minrk: Summarizing from out-of-band review, I don't think we need these `allow_all` checks. `allowed_hosted_domains` isn't special in terms of being overridden by `allow_all`.

And the _restrictive_ `hosted_domains` still makes sense with `allow_all=True`

oauthenticator/google.py
```diff
@@ -195,10 +198,54 @@ def _cast_hosted_domain(self, proposal):
...
+        Grant access to any user whose Google `hd` claim matches the one of the
+        domains listed.
+
+        Unlike `hosted_domain`, this does not restrict or block anyone on its
+        own. It's purely an additional way to grant access, checked
+        alongside `allowed_users`, `admin_users`, and `allowed_google_groups`.
+
+        Cannot be combined with `allow_all`, since `allow_all` already grants
+        access to everyone, making this configuration meaningless.
+        """,
```
minrk, line 214:
```suggestion

        .. versionadded:: 17.5
        """,
```

sean-morris: got it thanks

PR #830, opened 2026-07-15
deprecate hosted_domain in favor of restrict_hosted_domains, always apply strip_domain if True, clarify docstrings

minrk:
related to #828 and the confusion that led to it

In discussing this, I also considered deprecating `hosted_domain` in favor of `restrict_hosted_domains` to be even clearer and disambiguate from the new `allow_hosted_domains`, but didn't take that step here, this is just the docstring.

### DIFF
--- a/oauthenticator/google.py
+++ b/oauthenticator/google.py
@@ -125,10 +125,9 @@ class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
     strip_domain = Bool(
         config=True,
         help="""
-        Strip the username to exclude the `@domain` part. 
-        This happens by default when there is only one hosted domain specified.
+        Strip the username to exclude the `@domain` part.
 
-        Should usually combined with `restrict_hosted_domains` set to a single domain.
+        Should usually combined with `restrict_hosted_domains` containing a single domain.
 
         .. deprecated:: 17.5
             strip_domain is enabled by default when the deprecated `hosted_domain` is
@@ -140,7 +139,6 @@ class GoogleOAuthenticator(OAuthenticator, GoogleOAuth2Mixin):
             strip_domain is always applied if True, not just for accounts in `hosted_domain`.
 
         .. warning::
-
             If domains are stripped from usernames and multiple domains are allowed,
             multiple Google accounts may access the same JupyterHub account.
         """,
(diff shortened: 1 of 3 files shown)


manics (approved):
I think this is fine, and deprecating is also fine if you want since the configuration is pretty complex.

Is it intentional that `strip_domain` is ignored for everything other than `hosted_domain`
https://github.com/jupyterhub/oauthenticator/blob/5ffbbe90615cb0cfcfffc8b1f52c28851662ba21/oauthenticator/google.py#L250-L251

i.e. domains are never stripped if you're using `allowed_domains`?

oauthenticator/google.py
```diff
@@ -158,22 +158,35 @@ def _check_multiple_hosted_domain(self, strip_domain):
...
+           You can opt-out of this behavior by setting `GoogleOAuthenticator.strip_domain = False`.
 
-        Users not restricted by this configuration must still be explicitly
-        allowed by a configuration intended to allow users, like `allow_all`,
-        `allowed_users`, `allowed_hosted_domains`, or `allowed_google_groups`.
+        This config only **restricts** access, it does not **grant** any users access.
+        Users in these domains must still be explicitly
+        allowed by additional configuration intended to allow users,
+        such as
+        `allow_all`,`allowed_users`, `allowed_hosted_domains`, or `allowed_google_groups`, etc.
```
yuvipanda:
```suggestion
        `allow_all`, `allowed_users`, `allowed_hosted_domains`, or `allowed_google_groups`, etc.
```

oauthenticator/google.py
```diff
@@ -125,55 +125,95 @@ def _userdata_url_default(self):
     strip_domain = Bool(
         config=True,
         help="""
-        Strip the username to exclude the `@domain` part. 
-        This happens by default when there is only one hosted domain specified
+        Strip the username to exclude the `@domain` part.
 
-        .. warning::
+        Should usually combined with `restrict_hosted_domains` containing a single domain.
```
manics:
```suggestion
        Should be combined with `restrict_hosted_domains` containing a single domain.
```

oauthenticator/google.py
```diff
@@ -125,55 +125,95 @@ def _userdata_url_default(self):
...
+        **Only users in these domains** are considered for authentication with JupyterHub.
 
         .. warning::
 
-           Changing this config either to or from having a single entry is a
-           disruptive change as the same Google user will get a new username,
-           either without or with a domain name included.
+           Changing this config either to or from having a single entry
+           will change the default value of `strip_domain`,
+           (True when `hosted_domain` has exactly one domain, False, otherwise).
```
manics:
```suggestion
           (True when `hosted_domain` has exactly one domain, False otherwise).
```

oauthenticator/google.py
```diff
@@ -199,6 +239,36 @@ def _cast_hosted_domain(self, proposal):
...
+        .. seealso::
+
+            - :attr:`~.GoogleOAuthenticator.strip_domain`
+            - :attr:`~.GoogleOAuthenticator.allowed_hosted_domains` for granting access
+              to members of a domain without *excluding* accounts from other sources.
+        """,
+    )
+
+    @validate('restrict_hosted_domains')
+    def _cast_restric_hosted_domains(self, proposal):
```
manics:
```suggestion
    def _cast_restrict_hosted_domains(self, proposal):
```

